PR #3, opened 2022-12-15
feat(mock): update state mock

pavlzk:


InfamousVague:
Keeping around but closing for now

PR #9, opened 2022-12-21
chore(Build): Move over build files and windows configs

InfamousVague:


InfamousVague:
This is largely just a copy of https://github.com/Satellite-im/Uplink-UI/pull/7 but fixes a few conflicts and retains build.rs locations, The config may not be needed since we generate a default one in config.rs anyways so realistically it is probably just a runtime file; however it'd be nice to retain it so that we can more clearly specify default values.

PR #15, opened 2022-12-22
feat(toast notification):  toast notifications now auto-dismiss

sdwoodbury:
The toast notifications are not displayed in the UI. However, I verified that toast notifications are removed from the state when their time expires. 

state.ui.toast_notifications is now a HashMap - will want to pass the Uuid of a toast to the UI so that the correct toast notification can be updated when the cursor hovers it. 

state modifications are performed at the root of the application - called from main.rs. don't want to hide state changes in various files. 

`state.add_hook` didn't quite work because of ownership issues so I uses a `use_future` and passed it `state.inner()`. this `inner` doesn't include the reference to the context, so the ownership problems were avoided. another hook (a dioxus hook) was used to force a re-render when a toast was removed from the state.


sdwoodbury:
the changes to `ui/src/utils/notifications.rs` may not be necessary. do we want to use notify-rust at all? if not then perhaps that file should be removed.

PR #23, opened 2022-12-26
bootstrapped themes

InfamousVague:


InfamousVague:
This is almost ready to go, but there is a bug I can't seem to resolve where the files inside of the extra themes folder are not being properly copied over to the users cache directory properly. I tried to get this to work in `main.rs#52` specifically in `copy_assets`; it creates the directory properly if it doesn't exist, but it doesn't actually copy the contents of the directory.  Also need to fix `.gitignore` to properly ignore .DS_Store files.

lgmarchi:
> This is almost ready to go, but there is a bug I can't seem to resolve where the files inside of the extra themes folder are not being properly copied over to the users cache directory properly. I tried to get this to work in `main.rs#52` specifically in `copy_assets`; it creates the directory properly if it doesn't exist, but it doesn't actually copy the contents of the directory. Also need to fix `.gitignore` to properly ignore .DS_Store files.

Problem fixed changing themes relative path

PR #24, opened 2022-12-26
feat(warp): add outline for warp interaction

sdwoodbury:
This is one way we might add warp to Uplink-UI. This approach does the following
- execute Warp in a thread separate from the Dioxus UI
- communicates with the UI via channels
- the UI polls the channel, matches on the event, and updates state accordingly
- an additional channel is used for the UI to command warp, such as sending messages, adding/removing friends, ect. 

The purpose of this PR is to get feedback from the team before spending the time to fully implement this.


ui/src/warp_runner.rs
```diff
@@ -0,0 +1,198 @@
...
+        Self {
+            notify: Arc::new(Notify::new()),
+            ran_once: false,
+        }
+    }
+
+    // spawns a thread which will terminate when WarpRunner is dropped
+    pub fn run(&mut self, tx: WarpEventTx, rx: WarpCmdRx) {
+        if self.ran_once {
+            panic!("WarpRunner called run() multiple times");
```
dariusc93: Could use an assert here instead. Eg `assert!(self.run_once, "WarpRunner called run() multiple times");`

ui/src/warp_runner.rs
```diff
@@ -0,0 +1,198 @@
...
+            Err(_) => {
+                //doesnt exist so its set
+                let tess = Tesseract::default();
+                tess.set_file(DEFAULT_PATH.join(".keystore"));
+                tess.set_autosave();
+                tess
+            }
+        };
+
+        let (account, messaging, storage) = match warp::async_block_in_place_uncheck(
```
dariusc93: This could probably get moved inside of the tokio task instead and strip off the `async_block_in_place_uncheck`

dariusc93 (approved):
Looks like a good start. I am curious on how this would turn out with merging the streams together into the select while also receiving commands over the channel for different things (though it should work just fine in practice but tokio::select picks a random branch to poll for fairness rather than polling in order but this shouldnt be noticeable) 

I did leave a couple of small comments, but couple things to note though :

If this will be done this way, there is no need for the struct wrappers around the box object. That was mostly done to satisfy dioxus need for `PartialEq` for the property. 

 Might want to check to see how dioxus handles the order for the task when a component rerenders (eg does it drop the task before loading a new one or does it load a new one then drop previous). Even though the lock should drop and not cause a dead lock, i can imagine if there is consistent rerendering of the root component (if it ever does), that the lock would delay things while the task is awaiting to be dropped. 
 
It could get a bit messy with large branches of select so may want to have different functions to handle different commands/events and pass along what would be needed for those functions, etc.

dariusc93:
Another thing to add since i was looking at it a bit more: I dont know how this will be on the frontend if the UI would be awaiting an event or a response or if it would continue as normal until there is a response back and use that accordingly. If awaiting (eg either waiting for a response from the channel or if a oneshot is sent over while awaiting for the response), then it would require a use_future.

InfamousVague (approved):

InfamousVague:
I generally like this approach, I won't pretend to understand the intricacies but the abstraction from state and being deeply wired into the UI being avoided is a huge plus.

InfamousVague:
One small request, as this grows it might distract from the point of main.rs, might be worth having the function to create these bindings in another file which is included in main.

sdwoodbury:
> Even though the lock should drop and not cause a dead lock, i can imagine if there is consistent rerendering of the root component (if it ever does), that the lock would delay things while the task is awaiting to be dropped

I think the `use_future` in `main.rs` won't get re-created on re-render because the dependencies won't change.

ui/src/warp_runner.rs
sdwoodbury: thanks! i feared assert! would be compiled out but turns out that isn't true.

ui/src/warp_runner/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+                };
+
+            // this was the only way to get a mutable static variable. but this channel should only be read here.
+            let mut rx = rx.lock().await;
+
+            loop {
+                tokio::select! {
+                    // RayGun events
+                    // MultiPass events
+                    // ect
```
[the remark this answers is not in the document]
sdwoodbury: > One small request, as this grows it might distract from the point of main.rs, might be worth having the function to create these bindings in another file which is included in main.

done

PR #27, opened 2022-12-27
Use latest dioxus from master branch

calogero93:


ui/src/utils/mod.rs
```diff
@@ -6,7 +6,7 @@ use crate::state::Theme;
 
 pub mod format_timestamp;
 pub mod language;
-pub mod notifications;
+//pub mod notifications;
```
sdwoodbury, line 9: @InfamousVague  will we be using our custom notifications from now on? if so then the code relying on `rust-notify` can be removed.

kit/Cargo.toml
```diff
@@ -6,8 +6,12 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-dioxus = { version = "0.2.4", features = ["desktop"] }
-dioxus-heroicons = { git="https://github.com/Satellite-im/dioxus-heroicons" }
+dioxus = { git="https://github.com/DioxusLabs/dioxus", branch = "master" }
```
sdwoodbury: I feel that it would be safer to specify a `rev` here rather than rely on the `master` branch, which constantly changes. same for the `Cargo.toml` file in `ui/`.

sdwoodbury (changes requested):

kit/Cargo.toml
dariusc93: Yea it is safer and would also make sure everybody who builds are using the same commit than one person still using a older while another may be using a newer. On top of that, it would prevent possible bugs from being merged in in newert commits.

ui/src/utils/mod.rs
InfamousVague: We will be using external notifications, the internal ones are for errors and things like "copied to clipboard successfully"

InfamousVague:
If this tests out fine then it looks great to me! I agree though we need a rev

InfamousVague:
image
I also seem to have a bug where a bunch of text is missing

sdwoodbury (approved):

dariusc93 (approved):

PR #37, opened 2022-12-30
fix: Clone and show text instead of a div

dariusc93:
### What this PR does 📖

- Show text instead of a div

- N/A


InfamousVague:
tested locally this fixes translations.

PR #56, opened 2023-01-04
feat(MediaPlayer): improve handling of calls and MediaPlayer

### LINKED ISSUE #44: Convert popout player to new window.
Now that we have multi-window support in dioxus, we should just spawn the popout player in a new window and let the OS handle snapping and other window features. This will also allow us to remove a good chunk of javascript

sdwoodbury:
### What this PR does 📖

- the controls for the current call (mute, silence, ect) were moved from State::UI to State::UI::Call
- for a video call, UI::Call stores the WebView (currently this only happens when the video is popped out). the popout_player persists when one switches chats. 
- ToggleMedia and EndAll were changed because
    1) ToggleMedia allowed for setting multiple chats to having active media but only one media should be active at a time. 
    2) EndAll only makes sense if every chat can have active media (see previous point about this) 

- Resolve Satellite-im/Uplink#44 

### Additional comments 🎤
for some reason if overlays are toggled some number of times, ending with them being enabled, and the app is closed, the app crashes. But the WebViews close themselves when the UI struct is dropped. I suspect the test code in main.rs has something to do with this, and have left the problem for now in hopes that when overlays are actually used, this bug will go away.

### DIFF
--- a/ui/src/components/chat/compose.rs
+++ b/ui/src/components/chat/compose.rs
@@ -14,6 +14,7 @@ use super::sidebar::build_participants_names;
 pub fn Compose(cx: Scope) -> Element {
     let state = use_shared_state::<State>(cx)?;
     let active_chat = state.read().get_active_chat().unwrap_or_default();
+    let active_chat_id = active_chat.id;
     let message_groups = state.read().get_sort_messages(&active_chat);
 
     let without_me = state.read().get_without_me(active_chat.participants.clone());
@@ -34,8 +35,6 @@ pub fn Compose(cx: Scope) -> Element {
     let participants_name = build_participants_names(&without_me);
 
     let active_media = Some(active_chat.id) == state.read().chats.active_media;
-    // used to make this chat the active media chat
-    let active_media_chat = active_chat.clone();
  
     // TODO: Pending new message divider implementation
     // let _new_message_text = LOCALES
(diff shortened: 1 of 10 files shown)


phillsatellite (approved):

phillsatellite:
I wanted to mention a little thing I found, it seems like if you click the close button multiple times, it will just force the pop out into fullscreen and wont actually exit pop out player. I showed @sdwoodbury and going to create ticket for it 🔨 

https://user-images.githubusercontent.com/93608357/210622939-1d786c57-a32f-4d76-89eb-d269c091c267.mov

InfamousVague:
Two things I found:

- When the call is ended, the video continues to play.
- When the video is popped out, it should display the "media detatched" component like it used to.

sdwoodbury:
@InfamousVague 

>     * When the call is ended, the video continues to play.
  that issue is for ticket Satellite-im/Uplink-Backlog#2. and about that - it seems to be a problem with the video tag. i'm not sure it makes sense to spend time debugging that because as far as I know, we won't be using the video tag for video streaming. Rather we will be using some new thing that the Dioxus team is working on. 

>     * When the video is popped out, it should display the "media detatched" component like it used to.
i'll fix this.

sdwoodbury:
using `write()` instead of `write_silent()` to create the popout causes the video to not display in the popout window. makes no sense. if `write_silent()` is used and the user then clicks on one of the chats, the UI updates, the "media detached" message is displayed in the chat, and the video continues to work in the popout.

sdwoodbury:
Will revisit this later

PR #60, opened 2023-01-05
chore(Dioxus): bump to 3.0 🎉

InfamousVague:


ui/Cargo.toml
```diff
@@ -39,7 +39,7 @@ soloud = "1.0.2"
 open = "3.2.0"
 wry = { version = "0.23.4" }
 
-notify-rust = "=4.5.10"
+notify-rust = "4.6.0"
```
dariusc93: I had it at the previous version due to build errors with 4.6 and zbus that i have yet to investigate into. If there isnt any specific feature needed in 4.6, it would be best to retain it at 4.5 for the time being or use the following (assuming it builds on other platforms as well).

```suggestion
notify-rust = { version = "4.6.0", default-features = false, features = ["d"] }
```

PR #65, opened 2023-01-05
feat(cmdline): add command line args and miscellaneous clean up

sdwoodbury:
### What this PR does 📖
- this was a precursor to adding handling for warp events but I realized the login pages need to be added first, and decided to request this get merged now rather than later, to prevent future merge conflicts. 
- this PR lets the user specify whether to use mock data and specify a different location for the .warp and .uplink folders
- it also cleans up some things in State and reduced the size of the rsx! macro for the app element within main.rs 

- Resolve #


phillsatellite (approved):

ui/src/main.rs
```diff
@@ -252,8 +258,81 @@ fn app(cx: Scope) -> Element {
     let user_lang_saved = state.read().settings.language.clone();
     utils::language::change_language(user_lang_saved);
 
+    let pre_release_text = get_local_text("uplink.pre-release");
+
+    let theme = match &state.read().ui.theme {
```
dariusc93: This could probably be simplified by doing

```suggestion
    let theme = state.read().ui.theme.as_ref().map(|theme| theme.styles.clone()).unwrap_or_default();
```

ui/src/warp_runner/mod.rs
```diff
@@ -24,7 +29,8 @@ type Storage = Box<dyn Constellation>;
 type Messaging = Box<dyn RayGun>;
 
 pub enum WarpEvent {
-    None,
+    RayGun(Box<RayGunEventKind>),
```
dariusc93: Im guessing the events are boxed due to clippy? If so, those can be ignored in this case, though nothing is wrong with having it there if you dont want to ignore the clippy error by using `allow`

ui/src/warp_runner/mod.rs
```diff
@@ -70,20 +76,49 @@ impl WarpRunner {
...
+            let (mut account, mut messaging, _storage) =
+                match warp_initialization(WARP_PATH.clone(), tesseract.clone(), false).await {
                     Ok((i, c, s)) => (i, c, s),
                     Err(_e) => todo!(),
                 };
 
             // this was the only way to get a mutable static variable. but this channel should only be read here.
             let mut rx = rx.lock().await;
 
+            let mut raygun_stream = get_raygun_stream(&mut messaging).await;
```
dariusc93: Though there isnt a order restriction, it may be best to move this below the multipass stream loop

ui/src/warp_runner/mod.rs
```diff
@@ -70,20 +76,49 @@ impl WarpRunner {
...
+
+            let mut multipass_stream = loop {
```
dariusc93: This could be moved into an async function as well to keep it clean.

dariusc93: To simplify it, it could probably also be used with the `join!` macro.

dariusc93 (approved):
Did leave some comments and suggestions (all non-blocking) but overall looks good to me.

ui/src/warp_runner/mod.rs
sdwoodbury: yes, clippy complained that one variant was 17 bytes while the other was 240 bytes. but having to box everything is annoying.

ui/src/warp_runner/mod.rs
sdwoodbury: i like the idea of making separate functions for the futures. but `join!` won't terminate in response to the `Notified` being triggered.

PR #75, opened 2023-01-06
chore(Icons): Move icons into workspace, move gen to build script.

InfamousVague:


sdwoodbury (approved):
I changed the Paths to not use slashes to avoid a potential problem when building for windows. i also fixed some clippy warnings. looks good.

PR #77, opened 2023-01-06
feat(login): add unlock and auth layouts

sdwoodbury:
### What this PR does 📖

- Uplink now requires a password before logging in. if no account is created, the user may create one. 
- for now, until we decide what the flow should be, creating a new account will always fail if there's already an account at  --path. 
- translations aren't used because there's an outstanding PR related to translations which would just conflict with this. 
- the UI isn't in the final state. but as far as I know, it hasn't been designed either. 
- the `Input` element was changed to include a valid flag in the callback. that way 'onchange' and 'onenter' will know if the input should be used. 

- Resolve #


kit/src/elements/input/mod.rs
```diff
@@ -249,7 +248,7 @@ pub fn Input<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
                         class: "clear-btn",
                         onclick: move |_| {
                             *val.write() = "".into();
-                            emit(&cx, val.read().to_string());
+                            emit(&cx, val.read().to_string(), false);
```
sdwoodbury, line 253: this could trigger a false error message

ui/src/layouts/auth.rs
```diff
@@ -0,0 +1,129 @@
...
+use dioxus_router::use_router;
+use futures::StreamExt;
+use kit::{
+    elements::{
+        button::Button,
+        input::{Input, Options, Validation},
+    },
+    icons::Icon,
+};
+use tokio::sync::oneshot;
```
dariusc93: I would usually use oneshot from futures crate over whats in tokio

ui/src/layouts/auth.rs
```diff
@@ -0,0 +1,129 @@
...
+                warp_cmd_tx
+                    .send(WarpCmd::Tesseract(TesseractCmd::CreateIdentity {
+                        username: username.into(),
+                        passphrase: pin.into(),
+                        rsp: tx,
+                    }))
+                    .expect("UnlockLayout failed to send warp command");
+
+                let res = rx
+                    .blocking_recv()
```
dariusc93: Any reason to use `blocking_recv`? This would be best to use in a sync context as this would block the thread until either something is received or the channel is closed. Performing this in an async context would block all running task in that thread. If tokio doesnt have an async recv (been awhile since ive used oneshot from tokio so dont know for sure), you could use the futures oneshot instead.

ui/src/layouts/auth.rs
```diff
@@ -0,0 +1,129 @@
...
+    let ch = use_coroutine(cx, |mut rx: UnboundedReceiver<(String, String)>| {
+        to_owned![warp_cmd_tx, router];
+        async move {
+            while let Some((username, pin)) = rx.next().await {
+                //println!("auth got input");
+                let (tx, rx) = oneshot::channel::<Result<(), warp::error::Error>>();
+
+                warp_cmd_tx
+                    .send(WarpCmd::Tesseract(TesseractCmd::CreateIdentity {
+                        username: username.into(),
```
dariusc93: Due to username being a `String` there is no need to convert it to `String`
```suggestion
                        username,
```

ui/src/layouts/auth.rs
```diff
@@ -0,0 +1,129 @@
...
+                        passphrase: pin.into(),
```
dariusc93: Due to pin being a `String` there is no need to convert it to `String`
```suggestion
                        passphrase: pin,
```

ui/src/warp_runner/commands.rs
```diff
@@ -0,0 +1,54 @@
...
+pub enum TesseractCmd {
+    KeyExists {
+        key: String,
+        rsp: oneshot::Sender<bool>,
+    },
+    Unlock {
+        passphrase: String,
+        rsp: oneshot::Sender<Result<(), warp::error::Error>>,
+    },
+    CreateIdentity {
```
dariusc93, line 20: Im not sure if having this under `TesseractCmd` would be a clear since `Tesseract` itself doesnt create or handle anything identity related but more of an encrypted key-value store.

ui/src/warp_runner/mod.rs
```diff
@@ -51,27 +58,34 @@ impl std::ops::Drop for WarpRunner {
...
+        };
         Self {
             notify: Arc::new(Notify::new()),
             ran_once: false,
+            tesseract,
         }
     }
 
+    pub fn tesseract_initialized(&self) -> bool {
+        self.tesseract.exist("keypair")
```
dariusc93: Although this is fine to do, but it may be best to check multipass itself to determine if it been initialized.

ui/src/warp_runner/mod.rs
```diff
@@ -51,27 +58,34 @@ impl std::ops::Drop for WarpRunner {
...
-            Ok(tess) => tess,
-            Err(_) => {
-                //doesnt exist so its set
-                let tess = Tesseract::default();
-                tess.set_file(WARP_PATH.join(".keystore"));
-                tess.set_autosave();
-                tess
-            }
-        };
+        let mut tesseract = self.tesseract.clone();
```
dariusc93, line 89: Due to tesseract relying more on interior mutability, it doesnt need to be mutable directly. It shouldnt give any errors though but eventually a warming may come up about it it being unnecessary

dariusc93 (commented):
Looks good so far but I did leave some suggestions, with one mostly focused on being the blocking_recv.

ui/src/layouts/auth.rs
dariusc93: I looked into it and you should be able to do `rx.await` instead of `rx.blocking_recv()`

sdwoodbury: thanks for catching that. that was a big mistake on my part

phillsatellite:
Hey Stu! I noticed while testing that if the User puts an Emoji as they're Username the input field will appear green like its a accepted Username and when they try to proceed app will crash. If you'd rather have this made into a separate ticket so we don't hold up this pr just lemme know and I'll write up! 👍 

https://user-images.githubusercontent.com/93608357/211066772-98ef0b57-922e-4cca-8cf0-af17290d371d.mov

Screenshot 2023-01-06 at 12 35 53 PM

sdwoodbury:
> Hey Stu! I noticed while testing that if the User puts an Emoji as they're Username the input field will appear green like its a accepted Username and when they try to proceed app will crash. If you'd rather have this made into a separate ticket so we don't hold up this pr just lemme know and I'll write up! +1
> Screen.Recording.2023-01-06.at.12.35.34.PM.mov
> 
> Screenshot 2023-01-06 at 12 35 53 PM

this is probably an issue with the "Input" element. i set `alpha_numeric_only` to `true` so I'd think that emojis would be invalid input. this could probably be a separate issue

phillsatellite:
> > Hey Stu! I noticed while testing that if the User puts an Emoji as they're Username the input field will appear green like its a accepted Username and when they try to proceed app will crash. If you'd rather have this made into a separate ticket so we don't hold up this pr just lemme know and I'll write up! +1
> > Screen.Recording.2023-01-06.at.12.35.34.PM.mov
> > Screenshot 2023-01-06 at 12 35 53 PM
> 
> this is probably an issue with the "Input" element. i set `alpha_numeric_only` to `true` so I'd think that emojis would be invalid input. this could probably be a separate issue

Sounds good! I'll write up ticket and approve this pr 💥

phillsatellite (approved):

dariusc93:
> Hey Stu! I noticed while testing that if the User puts an Emoji as they're Username the input field will appear green like its a accepted Username and when they try to proceed app will crash. If you'd rather have this made into a separate ticket so we don't hold up this pr just lemme know and I'll write up! +1
> Screen.Recording.2023-01-06.at.12.35.34.PM.mov
> 
> Screenshot 2023-01-06 at 12 35 53 PM

As stated in #80 the main crash is due to the length of the username being less than 4 characters long (as well as other parts still being implemented) and not specifically due to it being an emoji. That shouldnt block this PR though since it is a separate issue

dariusc93 (approved):
LGTM. The only thing I will say is the `Cargo.lock` can remain since this mostly pertains to a UI rather than a library, but nothing is wrong with removing it either.

ui/src/warp_runner/mod.rs
sdwoodbury: maybe I should get rid of that for now. there's a button to let the user create a new account. no need to disable the login field automatically

sdwoodbury:
> LGTM. The only thing I will say is the `Cargo.lock` can remain since this mostly pertains to a UI rather than a library, but nothing is wrong with removing it either.

It definitely seems to be the practice to check in the Cargo.lock for an application, though for now maybe we can leave it out.

PR #91, opened 2023-01-07
chore(Icons): Adds additional icons via feather icons

InfamousVague:
### What this PR does 📖
Adds a bunch of icons


InfamousVague:
Worth noting, once we've decided which icons we need for the basic deployment of the app, we should either make our own or use entirely open source icons (which there are plenty) and support allowing users to use other icon sets (like vscode does)

PR #101, opened 2023-01-09
chore(icons): Change output of icon codegen

### LINKED ISSUE #96: Change icon crate to output to OUT_DIR
Currently in the icon crate, build.rs outputs directly to the `src` directory of the crate, however since this is being done at compile time, it should output to the path in the `OUT_DIR` env variable instead and use the `include` macro to include the files at compile time by grabbing the file from the `OUT_DIR`. This would reduce the need in having the files update in `src`, be included in the commit, prevent possible future conflicts, and cut back on the need of rustfmt during compile time (which is needless for any compile time codegen).

dariusc93:
### What this PR does 📖

- Change output from src directory to `OUT_DIR`
- Uses macro to include files from `OUT_DIR` directory.

- Resolve #96


sdwoodbury (approved):
looks like a nice way to not have to keep checking in outline.rs and solid.rs whenever new icons get added.

phillsatellite (approved):

PR #102, opened 2023-01-09
update(ui): add aria labels for ui elements

luisecm:
### What this PR does 📖
- Adding aria labels for UI elements in Kit and UI folders for Files/Folders, Settings, Chat, Friends and Auth Screen
- Having these aria labels will allow to keep adding screen object files in the tests repository

- Resolve #

### DIFF
--- a/kit/src/elements/file/mod.rs
+++ b/kit/src/elements/file/mod.rs
@@ -12,6 +12,8 @@ pub struct Props<'a> {
     #[props(optional)]
     disabled: Option<bool>,
     #[props(optional)]
+    aria_label: Option<String>,
+    #[props(optional)]
     with_rename: Option<bool>,
     #[props(optional)]
     onrename: Option<EventHandler<'a, String>>,
@@ -25,6 +27,13 @@ pub fn get_text(cx: &Scope<Props>) -> String {
     match &cx.props.text {
         Some(val) => val.to_owned(),
         None => String::from(""),
+    }
+}
+
+pub fn get_aria_label(cx: &Scope<Props>) -> String {
+    match &cx.props.aria_label {
+        Some(val)   => val.to_owned(),
+        None        => String::from(""),
     }
 }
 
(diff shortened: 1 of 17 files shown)


ui/src/layouts/files.rs
```diff
@@ -52,6 +54,7 @@ pub fn FilesLayout(cx: Scope<Props>) -> Element {
                                 Button {
                                     icon: Icon::FolderPlus,
                                     appearance: Appearance::Secondary,
+                                    aria_label: "add-folder".into(),
```
InfamousVague, line 57: Have you made sure this displays as expected? Only reason I ask is because Button is a custom element and IDK if it has the aria_label prop (unless you added that previously.)

luisecm: Yes, I added the aria_label as a prop in the UI kit before. Here is a screenshot on how it would be displayed:

image

kit/src/elements/file/mod.rs
```diff
@@ -28,6 +30,13 @@ pub fn get_text(cx: &Scope<Props>) -> String {
     }
 }
 
+pub fn get_aria_label(cx: &Scope<Props>) -> String {
+    match &cx.props.aria_label {
```
dariusc93: As a suggestion (no need to accept it here)
```suggestion
    cx.props.aria_label.clone().unwrap_or_default()
```
That way you dont have to worry about match conditions (which can apply to similar functions like this)

dariusc93 (approved):
Looks good to me from a glance over. Did leave a comment/suggestion but its non-blocking.

phillsatellite (approved):

phillsatellite:
Asked Luis if he wanted to make the changes suggested before we merge and he did so put changes request so everyone knows not to merge yet 😄

stavares843 (commented):
very cool 🎉

kit/src/elements/file/mod.rs
luisecm: Thanks! @dariusc93 I went ahead and changed the get_aria_label functions that I have added before to work like this

luisecm:
> Asked Luis if he wanted to make the changes suggested before we merge and he did so put changes request so everyone knows not to merge yet 😄

Changes applied!

PR #106, opened 2023-01-10
feat(misc): misc improvements

sdwoodbury:
### What this PR does 📖

- The way command line options are parsed and stored has been refactored to only parse the arguments once and store everything in a static variable 
- The Warp event and command channels were refactored to use a struct instead of a tuple

- Resolve #

### Additional comments 🎤
I made these changes while trying to fix issue 93. I got stuck but suspect that changing the login flow, which we need to do anyway, will fix the issue. I'm attempting to commit these improvements now so that the next PR doesn't get cluttered.


phillsatellite:
Hey @sdwoodbury ! Not sure if I was missing a step or not but I get this error when trying to compile. I tried cargo run --release and then cargo run -- --mock. I also made sure to update and clean cargo

Screenshot 2023-01-10 at 10 20 50 AM

sdwoodbury:
> Hey @sdwoodbury ! Not sure if I was missing a step or not but I get this error when trying to compile. I tried cargo run --release and then cargo run -- --mock. I also made sure to update and clean cargo
> 
> Screenshot 2023-01-10 at 10 20 50 AM

i can fix this. i used 'cargo update' but it looks like `ui` wasn't updated, only `kit`

sdwoodbury:
weird. my local repository says that Warp commit `ed3b5de968fb623a7d76170a5d7dd3f3c79d3e1b` has a `multipass::identity::IdentityStatus::Unknown` but I don't see that in the Warp repository.

sdwoodbury:
i had to uninstall and re-install rustup to fix it locally. looks like it builds now.

sdwoodbury:
i fixed my local build. had to re-install rustup and got rid of sccache just for good measure.

kit/Cargo.toml
```diff
@@ -15,6 +15,7 @@ icons = { path="../icons" }
 timeago = "0.4.0"
 humansize = "2.0.0"
 shared = { path = "../shared" }
+warp = { git = "https://github.com/Satellite-im/Warp", rev = "ed3b5de968fb623a7d76170a5d7dd3f3c79d3e1b"}
```
dariusc93: Im not against this (since it would make it easier to convert from/to other fields than what I temporarily did for mapping identity status and platform), though i think the kit crate wasnt suppose to include warp.

ui/Cargo.toml
```diff
@@ -39,6 +39,7 @@ opener = "0.5.0"
 soloud = "1.0.2"
 open = "3.2.0"
 wry = { version = "0.23.4" }
+lazy_static = "1.4.0"
```
dariusc93, line 42: Its usually best to use once_cell over lazy_static since `Lazy` doesnt have to be in a macro and is const.

ui/src/components/chat/compose.rs
```diff
@@ -115,7 +110,7 @@ pub fn Compose(cx: Scope) -> Element {
                                 UserImage {
                                     loading: **loading,
                                     platform: platform,
-                                    status: status,
+                                    status: active_participant.identity_status().into(),
```
dariusc93: :+1:

ui/src/warp_runner/commands.rs
```diff
@@ -51,10 +51,12 @@ pub async fn handle_multipass_cmd(
             passphrase,
             rsp,
         } => {
+            println!("create_identity: unlock tesseract");
```
dariusc93: Might want to use logging here

ui/src/warp_runner/commands.rs
```diff
@@ -51,10 +51,12 @@ pub async fn handle_multipass_cmd(
...
             if let Err(e) = tesseract.unlock(passphrase.as_bytes()) {
                 let _ = rsp.send(Err(e));
                 return;
             }
+            println!("create_identity: account.create_identity");
```
dariusc93: ditto

phillsatellite (approved):

dariusc93 (approved):
Looks good overall.

phillsatellite:
Worked perfect for me after the new changes, thanks Stu!

kit/Cargo.toml
InfamousVague: The UI Kit should be purely UI and not include any deps to warp.

PR #115, opened 2023-01-10
chore(misc): Cleanup of codebase

dariusc93:
### What this PR does 📖

- Cleans up code base and logic
- Remove unnecessary references
- Remove string conversion for enum where Display is implemented directly (Display implements this directly)

- N/A


.gitignore
```diff
@@ -1,6 +1,7 @@
 .DS_Store
 
 target
+kit/src/compiled_styles.css
```
dariusc93, line 4: Honestly, its really tempting to have this be outputted to OUT_DIR, similar to whats done #96 and include it from that path instead. Would keep the codebase clean, but that would be done in a separate PR.

kit/src/elements/folder/mod.rs
```diff
@@ -84,7 +79,7 @@ pub fn Folder<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
                 },
                 with_rename.then(|| rsx! (
                     Input {
-                        disabled: *disabled,
+                        disabled: disabled,
```
sdwoodbury, line 82: i'm surprised that `cargo fmt` didn't try to collapse this

sdwoodbury (approved):
very nice!

kit/src/elements/folder/mod.rs
dariusc93: Usually in macros, rustfmt doesnt have much of an effect on restructuring (or if it does, i havent been seeing it). I also know due to `Input` being a prop instead of a struct, it may not collapse like it would with your usual structs fields (while it is expanding at compile time)

phillsatellite (approved):

PR #122, opened 2023-01-12
update(global): responsive changes

InfamousVague:
This PR focuses primarily on the sidebar and main route views when in a mobile view. It also adds a new meta value which will be updated in a future PR to include a binding to the Wry window to detect resize events. Lastly we have added a getter into the UI which acts as a basic "is this mobile?" check. There will be a future is_mobile() getter but as this version is based purely on screen size, I chose to use the `is_minimal_view` until we have things that we literally only want to do on mobile.

**Adds:**

- Responsive-ish sidebar functionality between the minimal and maximum views.
- Moves friends sub nav to top of page so that it can be rendered more efficiently with the sidebar / back button.
- Adds button icons for sidebar to the topbar
- Adds meta values in state to render content conditionally based on available screen real-estate.
- Various small fixes around responsiveness.

**Needs:**

- Ability for sidebar to "snap" to full screen more efficiently somewhere in between mobile and tablet view.
- Resize listener for state.
- Various responsive fixes.

Will address these in a future PR.


phillsatellite (approved):

phillsatellite:
I found a few minor things worth mentioned while testing, create a ticket here with screenshots and recordings https://github.com/Satellite-im/Uplink/issues/125 

 -some of the pages you can scroll horizontally while in mobile view
-when you do chat in mobile it will only show sidebar
-and then I saw this too but I wasnt sure if this was intended or not but you are in chat, go to mobile view, go to friends, sidebar stays collapsed

PR #138, opened 2023-01-13
chore(tests): add GH workflow for CI UI tests execution

### LINKED ISSUE #137: Add CI execution for UI tests on MacOS

luisecm:
### What this PR does 📖

- Add github workflow file to execute on CI the UI tests for Uplink under MacOS

- Resolve #137

### DIFF
--- a/.github/workflows/ui-test-execution.yml
+++ b/.github/workflows/ui-test-execution.yml
@@ -1,4 +1,4 @@
-name: UI Automated Tests on MacOS
+name: UI Automated Tests on MacOS 🧪
 
 on:
   pull_request:
@@ -9,7 +9,7 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
-  build-app 🖥️:
+  build-app:
     runs-on: macos-latest
 
     steps:
@@ -64,7 +64,7 @@ jobs:
             Uplink-Mac-Universal.zip
             Uplink-Mac-Universal.zip.sha256.txt
 
-  test 🧪:
+  test:
     needs: build-app
     runs-on: macos-latest
 
--- a/.github/workflows/ui-test-execution.yml
+++ b/.github/workflows/ui-test-execution.yml
@@ -1,0 +1,142 @@
+name: UI Automated Tests on MacOS
+
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


phillsatellite (approved):

.github/workflows/ui-test-execution.yml
```diff
@@ -0,0 +1,142 @@
...
+          restore-keys: ${{ runner.os }}-cargo-
+
+      - name: Install Rust 💿
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
+
+      - name: Install Protobuf 💿
```
dariusc93, line 39: Not sure how it will work with a mac target but you could use `arduino/setup-protoc@v1` action for installing protobuf. 

See https://github.com/Satellite-im/Warp/blob/main/.github/workflows/warp_dart_update.yml#L16-L17 

Not a blocker though, especially if the gh action script doesnt work with a mac target

dariusc93 (approved):
Looks good to me from a glance through. Left a suggestion, but not blocking

Also for caching, while it would help a whole lot, do keep in mind that it can build up in storage overtime possibly exceeding the cache limit and causing it not to cache everything. Not sure if its possible to increase the cache limit though, but just something to keep in mind. Just to give an example. Building warp in debug profile alone have the `target` directory a little over 12GB while in the release profile, it would be about 3GB. This would likely increase as dependencies update, etc. This doesnt include the registry index and cache either.

.github/workflows/ui-test-execution.yml
luisecm: Thanks @dariusc93 ! Trying out in a new commit the `arduino/setup-protoc@v1` action to install protobuf. Could you let me know how can I improve the cargo cache step or if it is better to remove it? I think the action should work and build the app even without this step

InfamousVague (approved):

.github/workflows/ui-test-execution.yml
dariusc93: The caching part can stay.

luisecm: Thanks!

PR #141, opened 2023-01-13
feat(friends): mostly wire up friends to warp

sdwoodbury:
### What this PR does 📖

- uses warp to add a friend and accept the friend request
- uses warp to load conversations
- modifies the chats page and chats sidebar to use a "loading" state if there is not an active chat. this can happen for 2 reasons: the active chat isn't yet loaded from warp, or no messages have been sent,
- creates a new method on State which handles events from warp_runner and updates state accordingly 
- uses the "adapter pattern" to convert from warp events into something which makes sense to the `State` struct.  
- on the Friends, chats, sidebar, and main pages, express the enormous `rsx!` macro with multiple function calls. 
- not sure why the `auth` page is deleted again. i thought it was deleted in a previous PR, in favor of the "unlock" and "create_account" layouts
- don't use `Conversation::Default` because state expects to have the id of every conversation in `chats.all` and using `Conversation::Default` causes runtime errors. 
- allow `UserImage` to have zero participants, when it's loading 
- make the `Button` element use an `IconButton` instead of a `Button` because on the top bar for the add friends page (and for some reason only there), the Icon was getting the `onclick` event and not the `Button` 
- don't store messages in State but rather fetch them from Warp. same for conversations. however, store a list of Uuids for conversations which should be in the sidebar. 

There is still more to do, but this PR  already touches 22 files. 

- Resolve #

### Additional comments 🎤

known bugs 
- when logging in, the "create account" button briefly becomes active, allowing a user to try to re-create their account again. this will fail. 
- when copying your DID, the input box doesn't accept it unless the copy button for that window was first pressed. 
- when a friend request is accepted, the sender doesn't get updated 
- can't send messages (not really a bug, that one hasn't been implemented)
- adding, removing, and blocking friends updates the UI but doesn't trigger the required warp commands


sdwoodbury:
note that this requires the `--no-mock` flag. the warp features are disable when using mock data.

ui/src/components/friends/add.rs
```diff
@@ -8,32 +12,99 @@ use kit::{
...
+                    .send(WarpCmd::MultiPass(MultiPassCmd::RequestFriend {
+                        did,
+                        rsp: tx,
+                    }))
+                    .expect("AddFriendLayout failed to send warp command");
+
+                let res = rx.await.expect("failed to get response from warp_runner");
+                match res {
+                    Ok(_) => request_sent.set(true),
+                    Err(_) => todo!("failed to send friend request"),
```
sdwoodbury: this crashes if you try to add yourself as a friend. should fix

dariusc93: warp does send an error if one tries to add themself as a friend

sdwoodbury:
> note that this requires the `--no-mock` flag. the warp features are disable when using mock data.

turns out that this isn't entirely true but would certainly allow for some errors that shouldn't normally happen

ui/src/warp_runner/commands.rs
```diff
@@ -26,6 +34,28 @@ pub enum MultiPassCmd {
...
     },
+    RequestFriend {
+        did: DID,
+        rsp: oneshot::Sender<Result<(), warp::error::Error>>,
+    },
+    InitializeFriends {
+        rsp: oneshot::Sender<Result<friends::Friends, warp::error::Error>>,
+    },
+    GetOwnIdentity {
+        rsp: oneshot::Sender<Result<DID, warp::error::Error>>,
```
dariusc93, line 46: The oneshot should return `Identity` since `MultiPass::get_own_identity` returns one own identity, or maybe rename it to something like `GetOwnDid`. Not exactly a blocker though

ui/src/components/friends/add.rs
```diff
@@ -8,32 +12,108 @@ use kit::{
...
+            let warp_cmd_tx = WARP_CMD_CH.tx.clone();
+            while rx.next().await.is_some() {
+                let (tx, rx) = oneshot::channel::<Result<DID, warp::error::Error>>();
+                warp_cmd_tx
+                    .send(WarpCmd::MultiPass(MultiPassCmd::GetOwnIdentity { rsp: tx }))
+                    .expect("AddFriendLayout failed to send warp command");
+
+                let res = rx.await.expect("failed to get response from warp_runner");
+                match res {
+                    Ok(did) => my_id.set(Some(did.to_string())),
```
dariusc93, line 111: Any reason to convert it to a string instead of sending `DID` itself?

ui/src/components/friends/friends_list/mod.rs
```diff
@@ -15,16 +16,63 @@ use warp::multipass::identity::Relationship;
...
+                                rsp: tx,
+                            }))
+                            .expect("failed to send cmd");
+
+                        let rsp = rx.await.expect("command cancelled");
+
+                        match rsp {
+                            Ok(c) => c,
+                            Err(e) => {
+                                println!("failed to create conversation: {}", e);
```
dariusc93, line 66: Keep in mind, one of the errors is `Error::ConversationExist` which yields the `Conversation` of an existing conversation that been created (should only happen with a direct conversations and not group conversations) . If it havent been handled before receiving the oneshot, it should be handled here otherwise to handle other errors (when the time comes for wiring that part of it).

ui/src/warp_runner/commands.rs
```diff
@@ -79,5 +139,76 @@ pub async fn handle_multipass_cmd(
...
         }
+        MultiPassCmd::RequestFriend { did, rsp } => {
+            let r = match account.send_request(&did).await {
+                Ok(_) => Ok(()),
+                Err(e) => Err(e),
+            };
+            let _ = rsp.send(r);
+        }
+        MultiPassCmd::GetOwnIdentity { rsp } => {
+            let r = match account.get_own_identity().await {
```
dariusc93: Could do 

```rust
let r = account.get_own_identity().await.map(|id| id.did_key());
let _ = rsp.send(r);
```

ui/src/warp_runner/commands.rs
```diff
@@ -45,6 +75,36 @@ pub async fn handle_tesseract_cmd(cmd: TesseractCmd, tesseract: &mut Tesseract)
...
+                }
+                let _ = rsp.send(Ok(all_chats));
+            }
+            Err(_e) => {
+                // do nothing. will cancel the channel
+            }
+        },
+        RayGunCmd::CreateConversation { recipient, rsp } => {
+            match messaging.create_conversation(&recipient).await {
+                Ok(conv) => {
```
dariusc93:
```suggestion
                Ok(conv) | Err(Error::ConversationExist {
        conversation: conv,
    }) => {
```

Would suggest trying it in IDE but you should try this error too as a precaution since it would yield a conversation type.

ui/src/warp_runner/ui_adapter/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+    FriendRequestReceived(state::Identity),
+    FriendRequestSent(state::Identity),
+    FriendAdded(state::Identity),
+    FriendRemoved(state::Identity),
+    FriendRequestCancelled(state::Identity),
+    FriendOnline(state::Identity),
+    FriendOffline(state::Identity),
+}
+
+pub async fn did_to_identity(did: DID, account: &mut super::Account) -> state::Identity {
```
dariusc93: It might be best to return a `Result` since there can be many cases where it may field but shouldnt kill the application

ui/src/warp_runner/ui_adapter/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+    let warp_identity = account
+        .get_identity(did.into())
+        .await
+        .expect("could not get warp identity");
```
dariusc93: Using `expect` here is fine since the only time this *should* fail is if multipass extension isnt initialized.

ui/src/warp_runner/ui_adapter/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+    state::Identity::from(
+        warp_identity
+            .first()
+            .cloned()
+            .expect("could not get warp identity"),
```
dariusc93: Should use `ok_or(Error::IdentityDoesntExist)` if `Result` is used on the function.

ui/src/warp_runner/ui_adapter/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+        }
+        MultiPassEventKind::FriendAdded { did } => {
+            let identity = did_to_identity(did, account).await;
+            MultiPassEvent::FriendAdded(identity)
+        }
+        MultiPassEventKind::FriendRemoved { did } => {
+            let identity = did_to_identity(did, account).await;
+            MultiPassEvent::FriendRemoved(identity)
+        }
+        MultiPassEventKind::IdentityOnline { did } => {
```
dariusc93, line 127: Fine to use, but I wouldnt rely on this or `MultiPassEventKind::IdentityOffline` event all that much right now. It works, but sensitive (though I did a bit of refactoring that improves on it, but still subjected to things like timeouts, etc, which may trigger offline and online repeatedly - will improve on it though very soon).

ui/src/warp_runner/ui_adapter/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+    state::Identity::from(
+        warp_identity
+            .first()
+            .cloned()
+            .expect("could not get warp identity"),
+    )
+}
+
+pub async fn dids_to_identity(
+    dids: Vec<DID>,
```
dariusc93, line 47: Could probably have this be an `Iterator` instead to reduce allocations to a `Vec`. Not a blocker though

dariusc93 (commented):
Looks good through my first glance through it. I did leave some comments and suggestions but one suggestion would be to simply the match conditions when sending results over oneshot. Eg

```rust
            let r = match account.send_request(&did).await {
                Ok(_) => Ok(()),
                Err(e) => Err(e),
            };
            let _ = rsp.send(r);
```

could be simplied to 

```rust
            let r = account.send_request(&did).await;
            let _ = rsp.send(r);
```

Since all youre doing is sending the result over. Others, you could use `map`, etc. Just a matter of simplifying the logic a bit,

EDIT: Ill do another look through in my IDE soon.

phillsatellite:
So going through I found a few things, since this is just the start of adding warp if we want to merge this in and have me create tickets for that, its cool with me just lemme know 👍 

- There were a good amount of times that I had a crash when trying to send a request 

https://user-images.githubusercontent.com/93608357/212429500-8923ae7c-568a-4466-94fb-caad913873ed.mov

- I also had a few times it would crash when favoriting a chat

https://user-images.githubusercontent.com/93608357/212429582-ce9ce8e2-b3bb-4a6c-b383-5510f232d196.mov

- There were a couple times I would accept the request and go to the chat and it would crash (I'm assuming this was because its not updating on the other users side maybe)

https://user-images.githubusercontent.com/93608357/212429919-5ab85fa9-eeac-4451-b5ee-8f16de5d1b58.mov

- And the last thing it looks like a lot of the icons were pushed up a pixel or two and the language dropdown would be blank until changed

Screenshot 2023-01-13 at 4 08 19 PM

Screenshot 2023-01-13 at 4 07 23 PM

Screenshot 2023-01-13 at 4 07 16 PM

Screenshot 2023-01-13 at 4 04 04 PM

dariusc93:
> So going through I found a few things, since this is just the start of adding warp if we want to merge this in and have me create tickets for that, its cool with me just lemme know +1
> 
>     * There were a good amount of times that I had a crash when trying to send a request
> 
> 
> crash.when.sending.friend.request.mov
> 

This is due to that section not being implemented to handle any outside errors (though maybe it be best to print, log or panic with that error under the todo macro?

>     * I also had a few times it would crash when favoriting a chat
> 
> 
> Fav.error.mov
> 

This is due to https://github.com/Satellite-im/Uplink/blob/feat/warp-friends2/kit/src/components/user_image_group/mod.rs#L50

One should use `get` instead of getting the element from the index if the slice itself. That way you can check to see if its `Some` and use the value there than to assume it will always have an element in the index, which could panic if its out of bounds. 

>     * There were a couple times I would accept the request and go to the chat and it would crash (I'm assuming this was because its not updating on the other users side maybe)
> 
> 
> Going.to.chat.mov
> 
>     * And the last thing it looks like a lot of the icons were pushed up a pixel or two and the language dropdown would be blank until changed
> 
> 
> Screenshot 2023-01-13 at 4 08 19 PM
> 
> Screenshot 2023-01-13 at 4 07 23 PM
> 
> Screenshot 2023-01-13 at 4 07 16 PM
> 
> Screenshot 2023-01-13 at 4 04 04 PM

The other errors would require a bit more deeper check to determine if it should be fixed in this PR or be put into a separate ticket.

ui/src/components/friends/add.rs
sdwoodbury: this goes to the clipboard

from line 55
```
  if let Some(id) = my_id.get().clone() {
        let mut clipboard = Clipboard::new().unwrap();
        clipboard.set_text(id).unwrap();
        state
            .write()
            .mutate(Action::AddToastNotification(ToastNotification::init(
                "".into(),
                "Copied ID to clipboard!".into(),
                None,
                5,
            )));
        my_id.set(None);
    }
```

ui/src/components/friends/friends_list/mod.rs
sdwoodbury: thank you for pointing this out. and for adding it to the code :grin:

ui/src/warp_runner/ui_adapter/mod.rs
sdwoodbury: looking at this suggestion and the next two, i think i should probably use proper error handling in `warp_runner::commands`.

sdwoodbury:
https://user-images.githubusercontent.com/8636602/212446939-830e4a01-15c9-4a90-929d-49475ce2507f.mp4

sdwoodbury:
still panics due to favorites

dariusc93:
> still panics due to favorites

On https://github.com/Satellite-im/Uplink/blob/feat/warp-friends2/kit/src/components/user_image_group/mod.rs#L34 maybe change it to 

(see code in EDIT)

Since its panicking because `participants` is empty and its using the index of it. Long term, it would be best to use `get` and check to see if its `Option::Some(_)` as a precaution to prevent it from panicking.  

EDIT: Actually, I am thinking about a different logic. Doing that would still make it loading false since it would be empty

Instead maybe do

```rust
let loading = cx.props.loading.unwrap_or_default() || cx.props.participants.is_empty();
```

It would only be true if one or the other is true. If both are false, loading would be false.

InfamousVague:
> Instead maybe do
> let loading = cx.props.loading.unwrap_or_default() || cx.props.participants.is_empty();
> It would only be true if one or the other is true. If both are false, loading would be false.

I was thinking we could refactor state and just use the assumption that if data is None then it's loading, since if it's none we would otherwise not even render the element it feels like a safe assumption. This would mean making things in state Optional like a chat's participants being and option if we expect to have to wait for that data at some point. Effectively any data we might need to wait for in state should either be an option, or be empty (in the case of collections of data)

sdwoodbury:
> > Instead maybe do
> > let loading = cx.props.loading.unwrap_or_default() || cx.props.participants.is_empty();
> > It would only be true if one or the other is true. If both are false, loading would be false.
> 
> I was thinking we could refactor state and just use the assumption that if data is None then it's loading, since if it's none we would otherwise not even render the element it feels like a safe assumption. This would mean making things in state Optional like a chat's participants being and option if we expect to have to wait for that data at some point. Effectively any data we might need to wait for in state should either be an option, or be empty (in the case of collections of data)

the error is happening because I changed some fields from `#[serde(default)]` to `#[serde(skip)]`. It seems like a `Chat` shouldn't even exist without `participants`. For this reason, making `participants` an `Option` might be confusing. 

The only reason that `participants` might be empty is if the chat gets loaded from somewhere besides the `state.json` file. In this case I loaded `participants` and `messages` from `Warp`. 

 maybe `Chat.participants` shouldn't be `skip` but should be persisted in `state.json`.  One could also add a `bool` to the various structs in `State`. `Chats.initialized` could indicate that every `Chat` is valid.

InfamousVague:
> > > Instead maybe do
> > > let loading = cx.props.loading.unwrap_or_default() || cx.props.participants.is_empty();
> > > It would only be true if one or the other is true. If both are false, loading would be false.
> > 
> > 
> > I was thinking we could refactor state and just use the assumption that if data is None then it's loading, since if it's none we would otherwise not even render the element it feels like a safe assumption. This would mean making things in state Optional like a chat's participants being and option if we expect to have to wait for that data at some point. Effectively any data we might need to wait for in state should either be an option, or be empty (in the case of collections of data)
> 
> the error is happening because I changed some fields from `#[serde(default)]` to `#[serde(skip)]`. It seems like a `Chat` shouldn't even exist without `participants`. For this reason, making `participants` an `Option` might be confusing.
> 
> The only reason that `participants` might be empty is if the chat gets loaded from somewhere besides the `state.json` file. In this case I loaded `participants` and `messages` from `Warp`.
> 
> maybe `Chat.participants` shouldn't be `skip` but should be persisted in `state.json`. One could also add a `bool` to the various structs in `State`. `Chats.initialized` could indicate that every `Chat` is valid.

Agreed, I think I'd suggest it more for top level things like the chats themselves. as well as higher level things like the sidebar chats, messages  etc.

ui/src/components/friends/add.rs
```diff
@@ -8,32 +12,109 @@ use kit::{
...
+    let state = use_shared_state::<State>(cx)?;
+    let friend_input = use_state(cx, String::new);
+    let friend_input_valid = use_state(cx, || false);
+    let request_sent = use_state(cx, || false);
+    // used when copying the user's id to the clipboard
+    let my_id: &UseState<Option<String>> = use_state(cx, || None);
     // Set up validation options for the input field
-    let validation_options = Validation {
+    let friend_validation = Validation {
         // The input should have a maximum length of 32
```
lgmarchi: It is good to delete these comments about length, can be confused for other developers.

ui/src/components/friends/add.rs
```diff
@@ -8,32 +12,109 @@ use kit::{
...
-        // Use the default options for the remaining fields
-        ..Options::default()
-    };
+    // todo: add translations for toasts
+    if *request_sent.get() {
+        state
+            .write()
+            .mutate(Action::AddToastNotification(ToastNotification::init(
+                "".into(),
+                "Friend Request Sent!".into(),
```
lgmarchi: If not add on translates, let a TODO to add soon these strings. Not a block.

ui/src/components/friends/add.rs
```diff
@@ -8,32 +12,109 @@ use kit::{
...
+    }
+
+    if let Some(id) = my_id.get().clone() {
+        let mut clipboard = Clipboard::new().unwrap();
+        clipboard.set_text(id).unwrap();
+        state
+            .write()
+            .mutate(Action::AddToastNotification(ToastNotification::init(
+                "".into(),
+                "Copied ID to clipboard!".into(),
```
lgmarchi: If not add on translates, let a TODO to add soon these strings. Not a block.

ui/src/components/friends/friends_list/mod.rs
```diff
@@ -42,14 +90,16 @@ pub fn Friends(cx: Scope) -> Element {
                         },
                         sorted_friends.into_iter().map(|friend| {
                             let did = friend.did_key();
+                            let chat = state.read().get_chat_with_friend(&friend);
```
lgmarchi, line 94: Question. Is not possible to do just a variable, and clone it in places where we need instead of cloning it in other variables, chat2, chat3?

ui/src/layouts/chat.rs
```diff
@@ -29,14 +29,14 @@ pub fn ChatLayout(cx: Scope<Props>) -> Element {
             ChatSidebar {
                 route_info: cx.props.route_info.clone()
             },
-            (state.read().ui.is_minimal_view() && state.read().ui.sidebar_hidden || !state.read().ui.is_minimal_view()).then(|| rsx!(
+            //(state.read().ui.is_minimal_view() && state.read().ui.sidebar_hidden || !state.read().ui.is_minimal_view()).then(|| rsx!(
```
lgmarchi: If we are not going to use, is not better to delete this piece of code?

ui/src/warp_runner/commands.rs
```diff
@@ -36,8 +67,38 @@ pub async fn handle_tesseract_cmd(cmd: TesseractCmd, tesseract: &mut Tesseract)
...
+                        Ok(chat) => {
+                            let _ = all_chats.insert(chat.id, chat);
+                        }
+                        Err(_e) => todo!("log error"),
+                    };
+                }
+                let _ = rsp.send(Ok(all_chats));
+            }
+            Err(_e) => {
+                // do nothing. will cancel the channel
```
lgmarchi, line 93: If comment, let a TODO here, to add log events later.

lgmarchi (commented):
Nice PR. Finished code review, there are some comments about other alternatives and delete unnecessary code. I don't know if you are going to use in the future, but delete println! comments could be nice as well.

ui/src/components/friends/friends_list/mod.rs
sdwoodbury: unfortunately not. `state` doesn't need cloning because it's a `use_shared_state`.  the other variables have to go into a `move` closure and after that they can't be accessed again.

ui/src/layouts/chat.rs
sdwoodbury: normally yes, but in this case we want to fix the commented out code. i'll add an explanation.

ui/src/warp_runner/ui_adapter/mod.rs
sdwoodbury: i wasn't sure about how to make an async iterator so i used `reserve` for now.

dariusc93: A `Stream` is an equiv to an Iterator thats async, but there is really no need for that. Having the `Iterator` here is optional though

sdwoodbury: oh, passing an iterator instead of the `Vec<DID>`. got it.

phillsatellite:
I retested with the most recent commits and still had some issues here and there but I personally think itd be best to create those into a ticket and get this merged in. Still awesome work though, big step forward for us 😄 

The issues I still replicated were the ones mentioned in known bugs and 
- On Users 2nd attempt trying to copy did it will crash 
- Crash on sending the same request twice

phillsatellite (approved):

kit/src/components/user_image_group/mod.rs
```diff
@@ -27,12 +27,11 @@ pub fn emit(cx: &Scope<Props>, e: Event<MouseData>) {
 #[allow(non_snake_case)]
 pub fn UserImageGroup<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
     let pressable = cx.props.onpress.is_some();
-    let count = cx.props.participants.len() as i64 - 3;
+    let count = cx.props.participants.len() as i64 - 3; // todo: why is this  - 3?
```
InfamousVague: We can replace this with a comment if you think it will help future people. As mentioned in private this is used because we render a maximum of 3 participants visually and the rest is added as  a "+n" later

kit/src/components/user_image_group/mod.rs
```diff
@@ -27,12 +27,11 @@ pub fn emit(cx: &Scope<Props>, e: Event<MouseData>) {
...
     let group = cx.props.participants.len() > 2;
     let username = cx.props.with_username.clone().unwrap_or_default();
-    let single_user = &cx.props.participants[0];
 
-    let loading = cx.props.loading.unwrap_or_default();
+    let loading = cx.props.loading.unwrap_or_default() || cx.props.participants.is_empty();
```
InfamousVague, line 36: Looks good for now but wondering if we can have a better process, maybe hoisting this since it uses the loading prop.

kit/src/components/user_image_group/mod.rs
```diff
@@ -45,6 +44,7 @@ pub fn UserImageGroup<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
                 }
             )
         } else {
+            let single_user = &cx.props.participants[0];
```
InfamousVague, line 49: What is this one for?

ui/src/components/friends/friends_list/mod.rs
```diff
@@ -43,14 +91,16 @@ pub fn Friends(cx: Scope) -> Element {
...
+                            let chat2 = chat.clone();
+                            let chat3 = chat.clone();
                             let did_suffix: String = did.to_string().chars().rev().take(6).collect();
-                            let chat_with_friend = state.read().get_chat_with_friend(&friend);
-                            let chat_with_friend_context = state.read().get_chat_with_friend(&friend);
-                            let chat_with_friend_context_clone = chat_with_friend_context.clone();
                             let remove_friend = friend.clone();
-                            let remove_friend_2 = remove_friend.clone();
+                            let remove_friend_2 = friend.clone();
```
InfamousVague, line 99: God I wish we could clean this up, looks good though.

ui/src/layouts/chat.rs
```diff
@@ -29,14 +29,15 @@ pub fn ChatLayout(cx: Scope<Props>) -> Element {
             ChatSidebar {
                 route_info: cx.props.route_info.clone()
             },
-            (state.read().ui.is_minimal_view() && state.read().ui.sidebar_hidden || !state.read().ui.is_minimal_view()).then(|| rsx!(
+            // todo: figure out why this isn't working, fix, and  uncomment. need a special view for mobile devices.
```
InfamousVague: this works on my end, I can try to replicate an issue.

InfamousVague (changes requested):

kit/src/components/user_image_group/mod.rs
sdwoodbury: lines 92-93 for non-group messages. the diff shows that this variable used to be define on line 33. it was moved to only be defined for non-group messages.

kit/src/components/user_image_group/mod.rs
sdwoodbury: done

ui/src/layouts/chat.rs
sdwoodbury: this says that if  not ((mobile view and no sidebar) or (not mobile view) ) then don't render anything...

ui/src/components/friends/add.rs
dariusc93: Okay thats fine then, though it might be best long term to pass the DID or Identity and grab it from there (especially if you will be displaying your own identity in another component)

ui/src/layouts/chat.rs
InfamousVague: The sidebar would be rendered since it's not hidden and minimal_view is true. Looked into this a bit more and the welcome page needs a back button and we need a resize handler on the app. Noted an issue @sdwoodbury was having on linux that i'm not able to replicate but did see.

InfamousVague (approved):

phillsatellite (approved):

PR #149, opened 2023-01-18
feat(Logger): Create logger

### LINKED ISSUE #135: Create dev logger
A utility class called "Logger" which will accept log messages with four different levels (debug, warn, info and error). It should add the errors as structs like Error { level: ErrorLevel, message: String } then push the message to a log in memory. it should also write to a debug.log in .uplink folder. Lastly we should make a component which we can open in a new window which shows the log in a scrolling "pre" element so we can browse the log while running the app

lgmarchi:
### 1. Create a logger to show logs when activate it on developer mode 
a. When open debug logger, automatically it goes to final of the page
b. On new logs, it will go to final of the page. 
https://user-images.githubusercontent.com/63157656/213220989-1b3048a3-627f-4291-8eee-de557f4211e2.mov

### 2. Add menu on developer page to open debug 

image

### 4. MOCK: let mock logs to test this PR, when we enter on general, profile, privacy and audio settiings
- 

- Resolve #135 

@dariusc93 I saw your note about create new file, when file been used reach a size limit, I like this idea, but I think would be better to do on next PR.


phillsatellite:
Sent a DM to Lucas to verify I was testing this correctly, had a crash when I would attempt to open logger 

https://user-images.githubusercontent.com/93608357/213236057-47fad9f8-e5f1-4402-a106-aae40ec86310.mov

ui/src/logger/logger_debug.rs
```diff
@@ -0,0 +1,89 @@
...
+#[inline_props]
+#[allow(non_snake_case)]
+pub fn LoggerDebug(cx: Scope, _drop_handler: WindowDropHandler) -> Element {
+    Logger::activate_logger();
+    let window = use_window(cx);
+
+    let logger = LOGGER.read();
+    let logs = logger.show_log();
+
+    let logs_to_show = use_state(cx, || logs.clone());
```
dariusc93: It might be better to move `logs` into the state by doing `let logs_to_show = use_state(cx, || logger.show_log());`?

ui/src/logger/logger.rs
```diff
@@ -0,0 +1,159 @@
...
+use std::sync::Mutex;
+use warp::sync::RwLock;
+
+use chrono::Local;
+
+static LOG_ACTIVE: Lazy<RwLock<bool>> = Lazy::new(|| RwLock::new(false));
+
+static DEBUG_LOG_PATH: &str = ".uplink/debug.log";
+
+pub static LOGGER: Lazy<RwLock<Logger>> = Lazy::new(|| RwLock::new(Logger::load()));
```
dariusc93: While this is alright, I do wonder if it would be best to use a ATOM_REF instead.

ui/src/logger/logger.rs
```diff
@@ -0,0 +1,159 @@
...
+        let file = match File::open(DEBUG_LOG_PATH) {
+            Ok(log) => log,
+            Err(error) => {
+                Logger::error(format!("Unable to read debug.log file. {error}").as_str());
+                return Vec::new();
+            }
+        };
+
+        let reader = BufReader::new(file);
+        let mut logs: Vec<Log> = vec![];
```
dariusc93: Couldnt you use `log_entries` from `Logger` instead? Would be cheaper than reading the file, especially if its extremely large to read each time.

ui/src/logger/logger_debug.rs
```diff
@@ -0,0 +1,89 @@
...
+    let debug_logger_started_time = use_ref(cx, || formatted_datetime.clone());
+
+    let script = include_str!("./script.js");
+
+    use_future(cx, (), |_| {
+        to_owned![logs_to_show, window, script, logs_on_screen_len];
+        async move {
+            loop {
+                sleep(Duration::from_millis(100)).await;
+                let new_logs = logger.show_log();
```
dariusc93: Since this is being used in a `use_future`, we should 

A. Spawn a blocking task since the operations within `Logger::show_log` are blocking
B. Change the logic of the `Logger` to be async. 

I think A would be better. You would use `tokio::spawn_blocking` with the logger passed into it. That way, its reading operations would not block other operating task in the same thread. Of course, B would be preferred (since its a easy switch) but since its also being used outside of the future, we cannot really do that unless you want to spawn a use_future for the code above in the file, but that may not be desirable. 


```suggestion
                let new_logs = tokio::task::spawn_blocking(move || LOGGER.read().show_log()).await.expect("Should not panic");
```

Note: This should not panic unless there is an issue with the tokio runtime in which case might be a bug if the runtime is still operating. A safer option would be to do something like

```rust
let Ok(new_logs) = tokio::task::spawn_blocking(move || LOGGER.read().show_log()).await else { break };
```

dariusc93 (commented):
Looks like a good start in a logging implementation. I did leave some questions and suggestions and will be taking a closer look shortly

ui/src/logger/logger.rs
lgmarchi: Do you mean pub static LOGGER?

ui/src/logger/logger_debug.rs
lgmarchi: For me no problem.

ui/src/testing/mock.rs
```diff
@@ -60,8 +60,10 @@ pub fn generate_mock() -> State {
     State {
         ui: UI {
             current_call: None,
+            current_debug_logger: None,
             overlays: vec![],
             popout_player: false,
+            debug_logger: false,
```
sdwoodbury: i think this field is redundant and that `current_debug_logger.is_some()` is equivalent to `debug_logger`

sdwoodbury:
if possible, i think it would be useful to have the option to just see logs from the Uplink app. I suspect that for the debug window, we would usually only care about seeing logs from Uplink.

Perhaps the logger could be configured to have different rules for what gets persisted in memory and what gets written to the file. An easy default would be to only display (in the window) logs from Uplink. A more difficult approach would be to allow filtering the log display  by module. 

As for what gets written to disk, I definitely think it would be helpful to persist the logs from Uplink and Warp in separate locations. Perhaps there should be   one directory with just logs from Uplink and another directory with the combined Warp and Uplink logs.

InfamousVague:
> if possible, i think it would be useful to have the option to just see logs from the Uplink app. I suspect that for the debug window, we would usually only care about seeing logs from Uplink.
> 
> Perhaps the logger could be configured to have different rules for what gets persisted in memory and what gets written to the file. An easy default would be to only display (in the window) logs from Uplink. A more difficult approach would be to allow filtering the log display by module.
> 
> As for what gets written to disk, I definitely think it would be helpful to persist the logs from Uplink and Warp in separate locations. Perhaps there should be one directory with just logs from Uplink and another directory with the combined Warp and Uplink logs.

to add to this, I was thinking we could make a simple component that displays the most recent line from the log at the bottom of the app kind of like some IDEs do with compiler errors.

sdwoodbury:
> if possible, i think it would be useful to have the option to just see logs from the Uplink app. I suspect that for the debug window, we would usually only care about seeing logs from Uplink.
> 
> Perhaps the logger could be configured to have different rules for what gets persisted in memory and what gets written to the file. An easy default would be to only display (in the window) logs from Uplink. A more difficult approach would be to allow filtering the log display by module.
> 
> As for what gets written to disk, I definitely think it would be helpful to persist the logs from Uplink and Warp in separate locations. Perhaps there should be one directory with just logs from Uplink and another directory with the combined Warp and Uplink logs.

I think most of this can be disregarded. I mistakenly thought this would capture things logged using the log crate.

ui/src/logger/logger.rs
sdwoodbury: another benefit of using `log_entries` is that it would eliminate the need for regular expressions.

ui/src/logger/logger.rs
```diff
@@ -0,0 +1,159 @@
...
+            };
+            let mut log_entries = self.log_entries.lock().unwrap();
+            log_entries.push(new_log.clone());
+
+            let mut file = OpenOptions::new()
+                .append(true)
+                .open(&self.log_file)
+                .unwrap();
+
+            if let Err(error) = writeln!(file, "{:?}", new_log.clone()) {
```
sdwoodbury: I think it would be better to implement `Display` for `Log`. 
it's common for a logger implementation to display something like this: `<date time> <delimiter> <log level> <delimiter> <message>`.

 for example: `2022-01-19 | 09:41:30.33 | DEBUG | some message` 

another example:  `DEBUG | 2022-01-19 | 09:41:30.33 | module name | some message`

ui/src/logger/logger.rs
```diff
@@ -0,0 +1,159 @@
...
+            Err(error) => {
+                Logger::error(format!("Unable to read debug.log file. {error}").as_str());
+                return Vec::new();
+            }
+        };
+
+        let reader = BufReader::new(file);
+        let mut logs: Vec<Log> = vec![];
+
+        let re_level = Regex::new(r#"level: (.*?),"#).unwrap();
```
sdwoodbury: splitting by a delimiter could also work. though this would require the logs have a different format than relying on the `Debug` macro.

sdwoodbury (changes requested):
I think it's strange that
1.  the `debug`, `warn`, `info`, and `error` functions rely on `LOGGER` but `show_log` doesn't.  (edit: I think `show_log` should probably also use `LOGGER`)

2. `debug`, `warn`, `info`, and `error` call `LOGGER.read()` instead of `LOGGER.wrtie()`

3. `Logger::log` takes a `&self` instead of a `&mut self` and that `LOGGER` uses a `Mutex` internally. If the logging functions called `LOGGER.write()` then `log_entries` wouldn't need a mutex. 

other comments
- `LOGGER` probably shouldn't be public. 
- `LOG_ACTIVE` could be a boolean variable within `Logger`
- `DEBUG_LOG_PATH` should probably be added to `StaticArgs` from `main.rs`. that way if uplink is started with a different path, the logs get written to the correct location.

ui/src/logger/logger_debug.rs
lgmarchi: Darius, I changed code, to read log entries now, and not more read file every time, just in the beginning, when open debug logger.

ui/src/logger/logger.rs
lgmarchi: I did man, much better to avoid read file everytime

ui/src/logger/logger.rs
lgmarchi: I am using with RW_LOCK, and I can save new values already inside logger.rs file, it is working fine

ui/src/logger/logger.rs
lgmarchi: I don't have a display function, but way I save and way I show in UI are differents. 

I used this model to show in UI. 

image

Specially, using separated parts in UI, to be possible color each of them, with different colors. Like Example below: 

image

lgmarchi:
> I think it's strange that
> 
> 1. the `debug`, `warn`, `info`, and `error` functions rely on `LOGGER` but `show_log` doesn't.  (edit: I think `show_log` should probably also use `LOGGER`)
> 2. `debug`, `warn`, `info`, and `error` call `LOGGER.read()` instead of `LOGGER.wrtie()`
> 3. `Logger::log` takes a `&self` instead of a `&mut self` and that `LOGGER` uses a `Mutex` internally. If the logging functions called `LOGGER.write()` then `log_entries` wouldn't need a mutex.
> 
> other comments
> 
> * `LOGGER` probably shouldn't be public.
> * `LOG_ACTIVE` could be a boolean variable within `Logger`
> * `DEBUG_LOG_PATH` should probably be added to `StaticArgs` from `main.rs`. that way if uplink is started with a different path, the logs get written to the correct location.

I was doing some of these changes while you are reviewing, and you did nice requests, so I added them as well in last commits.

lgmarchi:
> > if possible, i think it would be useful to have the option to just see logs from the Uplink app. I suspect that for the debug window, we would usually only care about seeing logs from Uplink.
> > Perhaps the logger could be configured to have different rules for what gets persisted in memory and what gets written to the file. An easy default would be to only display (in the window) logs from Uplink. A more difficult approach would be to allow filtering the log display by module.
> > As for what gets written to disk, I definitely think it would be helpful to persist the logs from Uplink and Warp in separate locations. Perhaps there should be one directory with just logs from Uplink and another directory with the combined Warp and Uplink logs.
> 
> I think most of this can be disregarded. I mistakenly thought this would capture things logged using the log crate.

Ok, but it is totally possible to separate logs from Uplink and Logs from Warp as well, but I think would be better in other PR.

lgmarchi:
> > if possible, i think it would be useful to have the option to just see logs from the Uplink app. I suspect that for the debug window, we would usually only care about seeing logs from Uplink.
> > Perhaps the logger could be configured to have different rules for what gets persisted in memory and what gets written to the file. An easy default would be to only display (in the window) logs from Uplink. A more difficult approach would be to allow filtering the log display by module.
> > As for what gets written to disk, I definitely think it would be helpful to persist the logs from Uplink and Warp in separate locations. Perhaps there should be one directory with just logs from Uplink and another directory with the combined Warp and Uplink logs.
> 
> to add to this, I was thinking we could make a simple component that displays the most recent line from the log at the bottom of the app kind of like some IDEs do with compiler errors.

This is interesting, like when Android Studio is updating something and show progress in this bottom bar. Could be a ticket for this

ui/src/logger/logger.rs
lgmarchi: Following previous answer, we can change way logs are showing in the screen, and with it we can change color of each item. But I think it is not necessary to change way logs are saved on debug.log file.

sdwoodbury:
@lgmarchi  thanks for the changes. You did everything i asked for :)  

I noticed one more thing... It looks like this is intended to be a singleton and I think it departs from the singleton pattern slightly. Look at this example: https://github.com/fadeevab/design-patterns-rust/blob/main/creational/singleton/logger/log.rs 

The `Log` module defines a logger similar to how you do. but the functions to interact with it aren't defined on a struct. I think that `show_log`, `error`, `info`, etc should be accessible at the module level  rather than defined within `impl Logger`. example: `logger::error()`

ui/src/logger/logger.rs
```diff
@@ -0,0 +1,211 @@
...
+    pub fn get_logger() -> Logger {
+        LOGGER.read().clone()
+    }
+
+    pub fn get_log_entries() -> Vec<Log> {
+        LOGGER.read().log_entries.clone()
+    }
+
+    fn load() -> Logger {
+        let logger_path = dirs::home_dir()
```
sdwoodbury: `logger_path` should be `STATIC_ARGS.logger_path`. no need to join to home directory

ui/src/logger/logger.rs
```diff
@@ -0,0 +1,211 @@
...
+                datetime: Local::now().to_string(),
+            };
+
+            let log_to_log_entries = Log {
+                level: new_log.level.clone(),
+                message: new_log.message.clone(),
+                datetime: new_log.datetime[0..19].to_string(),
+            };
+
+            log_entries.push(log_to_log_entries.clone());
```
sdwoodbury: `log_entries.push(log_to_log_entries.clone());` => `log_entries.push(log_to_log_entries);`

ui/src/logger/logger.rs
```diff
@@ -0,0 +1,211 @@
...
+                .open(&self.log_file)
+                .unwrap();
+
+            *LOGGER.write() = Logger {
+                is_log_active: true,
+                log_file: self.log_file.clone(),
+                log_entries,
+            };
+
+            if let Err(error) = writeln!(file, "{:?}", new_log.clone()) {
```
sdwoodbury: `new_log.clone()` => `new_log`

ui/src/logger/logger.rs
```diff
@@ -0,0 +1,211 @@
...
+#[derive(Debug, Clone)]
+pub enum LogLevel {
+    Debug,
+    Info,
+    Warn,
+    Error,
+}
+
+impl LogLevel {
+    pub fn from_str(log_level: &str) -> Self {
```
sdwoodbury: clippy suggested implementing `FromStr` for ``LogLevel`. seems like a good idea

ui/src/testing/mock.rs
lgmarchi: Done!

sdwoodbury:
clippy complained that a module has the same name as its containing module. the contents of `logger.rs` should probably go inside of `logger/mod.rs`

lgmarchi:
> https://github.com/fadeevab/design-patterns-rust/blob/main/creational/singleton/logger/log.rs

Main reason is to keep pattern like log crate? Or do you think is pattern for this kind of file? 

I don't have problems with both, just to understand better, because now we are using just struct and function, not so different 
image

sdwoodbury:
> > https://github.com/fadeevab/design-patterns-rust/blob/main/creational/singleton/logger/log.rs
> 
> Main reason is to keep pattern like log crate? Or do you think is pattern for this kind of file?
> 
> I don't have problems with both, just to understand better, because now we are using just struct and function, not so different image

I think it's confusing to have a struct implementation that references a global variable of the same type. especially if the global variable is used behind the scenes for almost every function

sdwoodbury (approved):

phillsatellite (approved):

Issue #146, opened 2023-01-17
Issues from wiring friends to Warp


phillsatellite:
There were some bugs on Stu's PR for wiring Friends to Warp that we wanted to create into a ticket rather than hold up PR 

- [x] When copying the DID sometimes User will face a crash 

<img width="572" alt="Screenshot 2023-01-17 at 4 01 12 PM" src="https://user-images.githubusercontent.com/93608357/213011479-48b12cd7-06ff-40a5-985a-c6406c9035d0.png">

- [x] When User sends a friend request that they have already sent app will crash 

- [x] Sometimes will face a crash when sending the first friend request 

<img width="567" alt="Screenshot 2023-01-17 at 4 02 49 PM" src="https://user-images.githubusercontent.com/93608357/213011750-a16735c7-2606-445c-b8c6-8baabcfdd666.png">

- [x] UI does not update when someone accepts your friend request 

- [x] adding, removing, and blocking friends updates the UI but doesn't trigger the required warp commands

sdwoodbury:
in `ui/src/components/friends/add.rs` i should probably save the user's ID the first time and not make future calls to warp

sdwoodbury:
i believe this is fixed now

PR #155, opened 2023-01-18
feat(friends): wire up other friend events

sdwoodbury:
### What this PR does 📖

- the following friend related events should now work: send, accept, reject, unfriend
- block is wired up but needs some more work
- fixes a few items from https://github.com/Satellite-im/Uplink/issues/146

- Resolve #

### Additional comments 🎤

sometimes on Linux I get this error: `Failed to create shared memory: Too many open files`. I'm not sure if it's because of warp or saving `state.json` but it interferes with Uplink. eventually we probably do want to use a database instead of a json file. 

no way to unblock a friend
if side A and B are friends, side A then blocks B, A will disappear from B's friends list but B will still have the option to unfriend A
image

since the warp update there's something weird going on when i try to log in

https://user-images.githubusercontent.com/8636602/213306106-32b0a61d-4725-47b1-a01e-e016403518fd.mp4


sdwoodbury:
note that the video of me logging in was me entering the correct password, deleting the last character, then typing the same character again

phillsatellite:
> note that the video of me logging in was me entering the correct password, deleting the last character, then typing the same character again

I can write up ticket for it if we want to fix in another PR 👍

dariusc93:
That too many files error happens because of the file descriptors are low. On the first version of uplink i put fdlimit in to up the file descriptors (see https://github.com/Satellite-im/Uplink-Depricated/blob/dev/src/main.rs#L76). The same could happen here prior to initializing. 

Note: That error doesnt always pertains to files itself but really file descriptors. Networking applications open and closes them and some internally would increase the soft limits temporarily for that process

InfamousVague:
> sometimes on Linux I get this error: `Failed to create shared memory: Too many open files`. I'm not sure if it's because of warp or saving `state.json` but it interferes with Uplink. eventually we probably do want to use a database instead of a json file.

We could run some performance metrics in the future, but I think for the foreseeable future since it's just loading or saving an object straight from file to memory and not querying or anything it should be fine I would think. Serde is pretty quick.

InfamousVague:
also @sdwoodbury if I had to hazard a guess about the "deleting a character logs us in" i would assume it's because of the functionality to "automatically" log us in when we enter the correct pin (basically checking on every keypress after the minimum pin length) is probably firing but the state isn't being triggered properly until you delete the character. This can probably be fixed by switching a `onkeydown` to a `onkeyup` or something similar

sdwoodbury:
> also @sdwoodbury if I had to hazard a guess about the "deleting a character logs us in" i would assume it's because of the functionality to "automatically" log us in when we enter the correct pin (basically checking on every keypress after the minimum pin length) is probably firing but the state isn't being triggered properly until you delete the character. This can probably be fixed by switching a `onkeydown` to a `onkeyup` or something similar

`multipass` wasn't immediately available after decrypting tesseract. adding a delay fixed it.

sdwoodbury:
> That too many files error happens because of the file descriptors are low. On the first version of uplink i put fdlimit in to up the file descriptors (see https://github.com/Satellite-im/Uplink-Depricated/blob/dev/src/main.rs#L76). The same could happen here prior to initializing.
> 
> Note: That error doesnt always pertains to files itself but really file descriptors. Networking applications open and closes them and some internally would increase the soft limits temporarily for that process

i'm pretty certain that the file descriptor limit is being reached because of network activity. I saw some very weird stuff in my `lsof` output. i'm hoping it was related to IPFS. it consisted of repeated attempts to connect to a few odd domains (various VPS) but only happened when running Uplink.

dariusc93:
> > That too many files error happens because of the file descriptors are low. On the first version of uplink i put fdlimit in to up the file descriptors (see https://github.com/Satellite-im/Uplink-Depricated/blob/dev/src/main.rs#L76). The same could happen here prior to initializing.
> > Note: That error doesnt always pertains to files itself but really file descriptors. Networking applications open and closes them and some internally would increase the soft limits temporarily for that process
> 
> i'm pretty certain that the file descriptor limit is being reached because of network activity. I saw some very weird stuff in my `lsof` output. i'm hoping it was related to IPFS. it consisted of repeated attempts to connect to a few odd domains (various VPS) but only happened when running Uplink.

You are correct that its related to network activity from ipfs. Increasing the soft limits will help with the error you are seeing related to too many files opened (using say fdlimit will increase it for the process).

sdwoodbury:
> > note that the video of me logging in was me entering the correct password, deleting the last character, then typing the same character again
> 
> I can write up ticket for it if we want to fix in another PR 👍

 No need. I fixed it

ui/src/components/chat/compose.rs
```diff
@@ -26,11 +26,24 @@ struct ComposeData {
     platform: Platform
 }
 
+impl PartialEq for ComposeData {
+    fn eq(&self, _other: &Self) -> bool {
+        false
```
WanderingHogan, line 30: does this just mean an == returns false no matter what?

looking at this: https://doc.rust-lang.org/std/cmp/trait.PartialEq.html it seems we can compare objects and return true if specific keys are equal, but i am also don't understand this yet

sdwoodbury: good question. this is a hack. to put something in a struct which implements `Props` via `#[derive(Props)]`, it needs to be comparable. Dioxus will re-render an Element if the props change. 

`ComposeData` can't automatically implement `PartialEq` but i know that any time the parent Element is created, I need to re-render the children. I opted to implement PartialEq but make it always false to communicate that the element will always re-render.

ui/src/state/chats.rs
```diff
@@ -54,3 +54,11 @@ pub enum Direction {
     Incoming,
     Outgoing,
 }
+
+impl Chats {
+    pub fn join(&mut self, other: &mut HashMap<Uuid, Chat>) {
```
dariusc93: If youre planning on reusing `other`, you could just have this as a `mut ` instead of `&mut`, though i guess we could technically extend `self.all` with `other` too with `other` being moved to this function

ui/src/state/friends.rs
```diff
@@ -21,3 +21,23 @@ pub struct Friends {
...
 }
+
+impl Friends {
+    pub fn join(&mut self, other: &mut Friends) {
+        for (k, v) in other.all.drain() {
+            self.all.insert(k, v);
+        }
+
+        for v in other.blocked.drain() {
+            self.blocked.insert(v);
```
dariusc93, line 32: ditto. Could have `other` be owned by this function and extend `self.block` (as well as others) with the fields in `other` if `other` isnt going to be reused.

ui/src/warp_runner/mod.rs
```diff
@@ -156,20 +160,37 @@ impl WarpRunner {
...
                     // receive a command from the UI. call the corresponding function
                     opt = rx.recv() => {
                         //println!("got warp_runner cmd");
                         match opt {
                         Some(cmd) => match cmd {
                             WarpCmd::Tesseract(cmd) => handle_tesseract_cmd(cmd, &mut tesseract).await,
-                            WarpCmd::MultiPass(cmd) => handle_multipass_cmd(cmd, &mut tesseract, &mut account).await,
+                            WarpCmd::MultiPass(cmd) => {
+                                // if a command to block a user comes in, need to update the UI because warp doesn't generate an event for a user being blocked.
+                                // todo: ask for that event
```
dariusc93, line 177: In https://github.com/Satellite-im/Warp/pull/148 there will be an event in `MultiPassEventKind` that will notify the client of when they blocked or unblocked a user (eg if you were to use `MultiPass::block` it would emit an event to the stream letting you know they were successfully blocked apart from the function returning `Ok`). In a separate PR once that is merged will notify the stream of when they are blocked/unblocked (though still wondering on if i should allow that to be emitted to the stream or not)

ui/src/main.rs
```diff
@@ -136,17 +132,33 @@ pub enum AuthPages {
...
 struct Args {
+    /// The location to store the .uplink directory, within which a .warp, state.json, and other useful logs will be located
     #[clap(long)]
     path: Option<PathBuf>,
     #[clap(long)]
     experimental_node: bool,
     // todo: when the app is mature, default mock to false
     // there's no way to set --flag=true so for make the flag mean false
+    /// mock data is fake friends, conversations, and messages, which allow for testing the UI.
+    /// may cause crashes when attempting to add/remove fake friends, send messages to them, etc.
```
dariusc93, line 156: Though will be a separate issue, but I feel the mock data should be pushed behind a feature gate in the future instead of a flag, especially when it comes close to production so nobody could easily use this flag can cause conflicts.

dariusc93 (approved):
LGTM. Did leave some questions and comments but nothing blocking to this PR from my scan over it.

PR #156, opened 2023-01-18
chore(inputValidation): Fix input validation

### LINKED ISSUE #80: Account Creation - User can set an emoji as a Username
**Issue:** When User is creating a Username in Account Creation and they make it a single emoji `🥸 ` the input field turns green as if its a accepted Username and when they try to proceed the app will crash

**Screenshot/screen recording**

Screenshot 2023-01-06 at 12 35 53 PM

https://user-images.githubusercontent.com/93608357/211069903-e06de593-9bfd-41cd-a54e-9696547e451c.mov

WanderingHogan:
### What this PR does 📖
-
The if statements would always overwrite the error message variable to be whatever the last if statement was in the validate function, so if the length validation if statement was last, and it was not in error, it would set any errors from the prior validations to be nothing.

- Resolve #80

### DIFF
--- a/kit/Cargo.toml
+++ b/kit/Cargo.toml
@@ -15,7 +15,6 @@ icons = { path="../icons" }
 timeago = "0.4.0"
 humansize = "2.0.0"
 shared = { path = "../shared" }
-unic = "0.9.0"
 
 derive_more = "0.99"
 
--- a/kit/Cargo.toml
+++ b/kit/Cargo.toml
@@ -15,6 +15,7 @@ icons = { path="../icons" }
 timeago = "0.4.0"
 humansize = "2.0.0"
 shared = { path = "../shared" }
+unic = "0.9.0"
 
 derive_more = "0.99"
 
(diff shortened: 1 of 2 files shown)


sdwoodbury (approved):
nice catch

phillsatellite (approved):

dariusc93 (approved):
Good catch there. I do think long term would be to actually make `ValidationError` an real `Error` type rather than a String and in reality, we should return the error when it first happens. Eg we should do this

```rust
pub fn validate(cx: &Scope<Props>, val: &str) -> Option<ValidationError> {

    let options = cx.props.options.unwrap_or_default();

    let validation = options.with_validation.unwrap_or_default();

    if validation.alpha_numeric_only {
        validate_alphanumeric(val)?;
    }

    if validation.no_whitespace {
        validate_no_whitespace(val)?;
    }

    if validation.max_length.is_some() || validation.min_length.is_some() {
        validate_min_max(val, validation.min_length, validation.max_length)?;
    }

    None
}
```

That way, if there is an error, it would return `Some(_)` on the on anything that returns some, but if none of them error it would return `None` at the end. 

EDIT: Actually, as I check over this, it should use a `Result` instead of a `Option`

PR #162, opened 2023-01-21
chore(validation): added ability to ignore colons on inputs, which lets us insert friend codes but still otherwise limit alpha numeric chars

### LINKED ISSUE #157: Add Friend input field only accepting alphanumeric chars
**Issue:** It seems like the rules for setting a Username are being used on the Add Friend feature as well. When User paste DID in input field it shows an error "Only alphanumeric characters are accepted. 

**Screenshot:**

image

WanderingHogan:
### What this PR does 📖
This adds a boolean override to the validation struct on the input that allows you to allow colons while continuing to disallow all the other not alpha numeric inputs. This makes it so the add friend input can have colons and the other inputs with the validator can not.

I tried

- making it an char vector so it was more customizable, but could not figure it out.
- making the override input an Option<String> that i could then split, but you cant derive Copy on an optional string
- checked the char methods available, to see if there was something we could use in place of is_alphanumeric, but didn't find anything
- A few other things i can't recall off hand

if you guys can think of a better way for me to do this, I would be glad to give it another go

- 

- Resolve #157

### DIFF
--- a/kit/src/elements/input/mod.rs
+++ b/kit/src/elements/input/mod.rs
@@ -81,7 +81,7 @@ pub fn validate_no_whitespace(val: &str) -> Option<ValidationError> {
 
 // Default to requireing alpha-numeric inputs, unless ignore_colon override is set on the input field
 pub fn validate_alphanumeric(val: &str, ignore_colon: bool) -> Option<ValidationError> {
-    val_no_colons = val.to_string();
+    let mut val_no_colons = val.to_string();
     if val.contains(":") && ignore_colon {
         val_no_colons = val.replace(":", "a");
     }
--- a/kit/src/elements/input/mod.rs
+++ b/kit/src/elements/input/mod.rs
@@ -81,7 +81,6 @@ pub fn validate_no_whitespace(val: &str) -> Option<ValidationError> {
 
 // Default to requireing alpha-numeric inputs, unless ignore_colon override is set on the input field
 pub fn validate_alphanumeric(val: &str, ignore_colon: bool) -> Option<ValidationError> {
-    let mut val_no_colons: String;
     val_no_colons = val.to_string();
     if val.contains(":") && ignore_colon {
         val_no_colons = val.replace(":", "a");
--- a/kit/src/elements/input/mod.rs
+++ b/kit/src/elements/input/mod.rs
@@ -83,7 +83,7 @@ pub fn validate_no_whitespace(val: &str) -> Option<ValidationError> {
 pub fn validate_alphanumeric(val: &str, ignore_colon: bool) -> Option<ValidationError> {
     let mut val_no_colons: String;
     val_no_colons = val.to_string();
-    if val.contains(":") && ignore_colon == true {
+    if val.contains(":") && ignore_colon {
         val_no_colons = val.replace(":", "a");
     }
     if !val_no_colons.chars().all(char::is_alphanumeric) {
--- a/kit/src/elements/input/mod.rs
+++ b/kit/src/elements/input/mod.rs
@@ -138,7 +138,7 @@ pub fn validate(cx: &Scope<Props>, val: &str) -> Option<ValidationError> {
 
     if validation.alpha_numeric_only 
         && validate_alphanumeric(val, validation.ignore_colons).is_some() {
-            error = validate_alphanumeric(val);
+            error = validate_alphanumeric(val, validation.ignore_colons);
     }
 
     if validation.no_whitespace 
--- a/kit/src/elements/input/mod.rs
+++ b/kit/src/elements/input/mod.rs
@@ -81,11 +81,11 @@ pub fn validate_no_whitespace(val: &str) -> Option<ValidationError> {
 
 // Default to requireing alpha-numeric inputs, unless ignore_colon override is set on the input field
 pub fn validate_alphanumeric(val: &str, ignore_colon: bool) -> Option<ValidationError> {
-    let mut val_no_colons = val.to_string();
-    if val.contains(":") && ignore_colon {
-        val_no_colons = val.replace(':', "a");
+    let mut val = val.to_string();
+    if ignore_colon {
+        val.retain(|c| c != ':');
     }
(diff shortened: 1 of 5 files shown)


kit/src/elements/input/mod.rs
```diff
@@ -78,10 +79,17 @@ pub fn validate_no_whitespace(val: &str) -> Option<ValidationError> {
     None
 }
 
-pub fn validate_alphanumeric(val: &str) -> Option<ValidationError> {
-    if !val.chars().all(char::is_alphanumeric) {
+// Default to requireing alpha-numeric inputs, unless ignore_colon override is set on the input field
+pub fn validate_alphanumeric(val: &str, ignore_colon: bool) -> Option<ValidationError> {
+    let mut val_no_colons: String;
+    val_no_colons = val.to_string();
+    if val.contains(":") && ignore_colon == true {
```
dariusc93:
```suggestion
    if val.contains(":") && ignore_colon {
```

kit/src/elements/input/mod.rs
```diff
@@ -78,10 +79,17 @@ pub fn validate_no_whitespace(val: &str) -> Option<ValidationError> {
     None
 }
 
-pub fn validate_alphanumeric(val: &str) -> Option<ValidationError> {
-    if !val.chars().all(char::is_alphanumeric) {
+// Default to requireing alpha-numeric inputs, unless ignore_colon override is set on the input field
+pub fn validate_alphanumeric(val: &str, ignore_colon: bool) -> Option<ValidationError> {
+    let mut val_no_colons: String;
```
dariusc93:
```suggestion
```

kit/src/elements/input/mod.rs
```diff
@@ -78,10 +79,17 @@ pub fn validate_no_whitespace(val: &str) -> Option<ValidationError> {
...
+    val_no_colons = val.to_string();
```
dariusc93:
```suggestion
    let mut val_no_colons = val.to_string();
```

kit/src/elements/input/mod.rs
```diff
@@ -78,10 +79,17 @@ pub fn validate_no_whitespace(val: &str) -> Option<ValidationError> {
...
+    if val.contains(":") && ignore_colon == true {
+        val_no_colons = val.replace(":", "a");
```
dariusc93:
```suggestion
        val_no_colons = val.replace(':', "a");
```

dariusc93 (commented):
Left some suggestions

WanderingHogan:
@dariusc93 thank you

kit/src/elements/input/mod.rs
```diff
@@ -78,10 +79,16 @@ pub fn validate_no_whitespace(val: &str) -> Option<ValidationError> {
     None
 }
 
-pub fn validate_alphanumeric(val: &str) -> Option<ValidationError> {
-    if !val.chars().all(char::is_alphanumeric) {
+// Default to requireing alpha-numeric inputs, unless ignore_colon override is set on the input field
+pub fn validate_alphanumeric(val: &str, ignore_colon: bool) -> Option<ValidationError> {
+    let mut val_no_colons = val.to_string();
```
sdwoodbury: perhaps `val_no_colons` could be named `val` like it was before, because the value sometimes does contain colons

kit/src/elements/input/mod.rs
```diff
@@ -78,10 +79,16 @@ pub fn validate_no_whitespace(val: &str) -> Option<ValidationError> {
...
+    if val.contains(":") && ignore_colon {
```
sdwoodbury: it might be better to check the boolean condition `ignore_colon` first. that way, if it is false, no further operations happen. 

maybe the statement could be 
```
if ignore_colon {
    val.retain(|c| c != ':');
}

File: kit/src/elements/input/mod.rs [hunk 7]
```diff
@@ -78,10 +79,16 @@ pub fn validate_no_whitespace(val: &str) -> Option<ValidationError> {
...
+        val_no_colons = val.replace(':', "a");
+    }
+    if !val_no_colons.chars().all(char::is_alphanumeric) {
```
sdwoodbury: nice use of `.all()`

sdwoodbury (approved):
approved. please consider my comments as non-blocking suggestions. the code would be fine without them.

dariusc93 (approved):

phillsatellite:
Hmm, not sure if this is just a Phil problem or not but when I tried to compile I would get this error, I had tried to do cargo update/clean and then also cleared cache and would still get error

Screenshot 2023-01-23 at 11 04 51 AM

tooshel:
@phillsatellite I didn't see that on windows but will try on Mac if you keep having that issue after the normal troubleshooting steps!

tooshel:
@phillsatellite Same thing happened when I checkout out the branch on windows 
image

dariusc93:
> Hmm, not sure if this is just a Phil problem or not but when I tried to compile I would get this error, I had tried to do cargo update/clean and then also cleared cache and would still get error
> 
> Screenshot 2023-01-23 at 11 04 51 AM

Pull and try again. It looks like it was missing a boolean argument to that function and I pushed up a commit for that

phillsatellite (approved):

phillsatellite:
Hogan wants to look at this a little more before we merge I put label so no one merges it

kit/src/elements/input/mod.rs
WanderingHogan: good idea, changing now, thank you. I wanted to make sure i wasn't doing something weird in rust like changing the value of the variable outside the function.

kit/src/elements/input/mod.rs
WanderingHogan: that's cleaner, thank you!

WanderingHogan:
@phillsatellite can you test again after I made Stu's changes, I tested and it looks like everything works as expected

phillsatellite (approved):

phillsatellite:
@WanderingHogan Worked good for me!

PR #167, opened 2023-01-23
update(vs-extensions): added Darius recommended vs-code extensions

WanderingHogan:
### What this PR does 📖
Darius pointed out these extensions might help me, so I added them to the .vscode folder to recommend them to anyone using the project with vscode
- 

- Resolve #


dariusc93 (approved):
Looks good to me. Any anybody have any suggestions on what to add can provide it too. 

`ms-vscode.makefile-tools` may not be needed here in this case though but can stay

phillsatellite (approved):

PR #168, opened 2023-01-23
feat(friends): improve friend removal

### LINKED ISSUE #163: improve friends functionality
forgot to implement these before merging the `feat/finish-friends` branch

- [x]  When chatting with a conversation we should remove ourselves from the list of participants when generating the username text, this will allow it to work for both group chats and direct chats with the same code.
- [x] When interacting with mock data, the app crashes. (This one can be another ticket)
- [x]  When removing a user, it should also remove any chats with that user, (this is already a method in state), to clear out all of the favorites, sidebar chats, etc.  If you're in a group chat with a user, it probably doesn't need to remove that. Unless we NEED to be friends with all users in a group.
- [ ] When trying to add a friend that also added you, the request will disappear and we'll add the user, which is totally fine, but maybe we should automatically jump to a conversation when we add a user or accept a request.
- [ ] Sometimes when pasting in a DID key the "add" button doesn't update until we do something else to force update state.

sdwoodbury:
### What this PR does 📖

- when a friend is removed or blocked, delete the conversation if the conversation has only 2 participants (leaves group conversations alone)
- initialize `Account` with a value other than Default. This fixes the problem where the conversation participants include the current user. 

- Resolve #163 #146

### Additional comments 🎤
- wasn't able to reproduce crash when copying DID
- received permission not to make a change for this one: "When trying to add a friend that also added you, the request will disappear and we'll add the user, which is totally fine, but maybe we should automatically jump to a conversation when we add a user or accept a request." because a user may have multiple incoming friend requests.

### DIFF
--- a/ui/src/components/friends/friends_list/mod.rs
+++ b/ui/src/components/friends/friends_list/mod.rs
@@ -31,7 +31,8 @@ enum ChanCmd {
     CreateConversation { recipient: DID, chat: Option<Chat> },
     RemoveFriend(DID),
     BlockFriend(DID),
-    RemoveConversation(DID),
+    // will remove direct conversations involving the friend
+    RemoveDirectConvs(DID),
 }
 
 #[allow(non_snake_case)]
@@ -115,10 +116,10 @@ pub fn Friends(cx: Scope) -> Element {
                             logger::error(&format!("failed to block friend: {}", e));
                         }
                     }
-                    ChanCmd::RemoveConversation(recipient) => {
+                    ChanCmd::RemoveDirectConvs(recipient) => {
                         let (tx, rx) = oneshot::channel::<Result<(), warp::error::Error>>();
                         warp_cmd_tx
-                            .send(WarpCmd::RayGun(RayGunCmd::Remove2PersonConv {
+                            .send(WarpCmd::RayGun(RayGunCmd::RemoveDirectConvs {
                                 recipient: recipient.clone(),
                                 rsp: tx,
                             }))
@@ -206,7 +207,7 @@ pub fn Friends(cx: Scope) -> Element {
                                             text: get_local_text("uplink.remove"),
                                             onpress: move |_| {
                                                 ch.send(ChanCmd::RemoveFriend(remove_friend.did_key()));
-                                                ch.send(ChanCmd::RemoveConversation(remove_friend.did_key()));
+                                                ch.send(ChanCmd::RemoveDirectConvs(remove_friend.did_key()));
                                             }
                                         },
                                         ContextItem {
@@ -215,7 +216,7 @@ pub fn Friends(cx: Scope) -> Element {
                                             text: get_local_text("friends.block"),
                                             onpress: move |_| {
                                                 ch.send(ChanCmd::BlockFriend(block_friend.did_key()));
-                                                ch.send(ChanCmd::RemoveConversation(block_friend.did_key()));
+                                                ch.send(ChanCmd::RemoveDirectConvs(block_friend.did_key()));
                                             }
                                         },
                                     )),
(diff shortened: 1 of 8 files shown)


ui/src/state/mod.rs
```diff
@@ -363,13 +363,11 @@ impl State {
             .cloned()
     }
 
-    pub fn get_without_me(&self, identities: Vec<Identity>) -> Vec<Identity> {
-        let mut set = HashSet::new();
-        set.insert(&self.account.identity);
-
+    pub fn get_without_me(&self, identities: &[Identity]) -> Vec<Identity> {
```
sdwoodbury, line 366: I changed this from a `&Vec` to a `&[T]` because I read [here](https://rust-unofficial.github.io/patterns/idioms/deref.html) that this is better

dariusc93: Yea, in many cases when borrowing something like a `Vec`, its best to have it as a borrowed slice. Like how its better to use `&str` instead of `&String`. If youre using specific functionality from that object though then its best to use it directly than deref, then having to reallocate.

ui/src/warp_runner/commands.rs
```diff
@@ -238,3 +246,42 @@ async fn multipass_initialize_friends(
...
+async fn raygun_remove_2person_conv(
+    recipient: DID,
+    messaging: &mut Messaging,
+) -> Result<(), Error> {
+    match messaging.list_conversations().await {
+        Ok(convs) => {
+            for conv in convs {
+                // check if conversation should be deleted
+                // only consider conversations with 2 participants
+                if conv.recipients().len() == 2 && conv.recipients().contains(&recipient) {
```
dariusc93: Not sure if this matters but you could check the conversation type and if its `ConversationType::Direct` (which would only contains 2 recipients) instead. If its `ConversationType::Group` it can be expected to have more than 2 recipients.

ui/src/components/friends/friends_list/mod.rs
```diff
@@ -121,6 +115,23 @@ pub fn Friends(cx: Scope) -> Element {
                             logger::error(&format!("failed to block friend: {}", e));
                         }
                     }
+                    ChanCmd::RemoveConversation(recipient) => {
+                        let (tx, rx) = oneshot::channel::<Result<(), warp::error::Error>>();
+                        warp_cmd_tx
+                            .send(WarpCmd::RayGun(RayGunCmd::Remove2PersonConv {
```
dariusc93: Could possibly be a check to determine what to send since group conversation being deleted would be broadcasted under the same event while this would only work if its a direct conversation (with only 2 recipients)

dariusc93 (commented):
Looks good overall. A main comment was left regarding `ChanCmd::RemoveConversation`. Will do another check over a little later

ui/src/components/friends/friends_list/mod.rs
sdwoodbury: I made a separate enumeration for removing direct conversations. i don't see a way to delete just the conversation at the moment - just a way to hide the chat. maybe this is good enough for now.

phillsatellite:
Hey Stu! Just a quick question in regards to the hiding the conversation when removing a friend, when I removed and went back to chat it shows forever loading. Should it appear like that or have the button to "start chat" if user has none 
 
https://user-images.githubusercontent.com/93608357/214138759-0ca18da4-e966-4bb7-80da-181c95412554.mov

sdwoodbury:
@phillsatellite 

that was a bug. i believe i just fixed it. 

edit: figured out why it was happening. `get_compose_data` was returning `None` because I wasn't clearing `state.chats.active`  when a conversation was deleted. and `get_compose_data` was trying to use `state.chats.active` to look up the `Conversation`.

phillsatellite:
> @phillsatellite
> 
> that was a bug. i believe i just fixed it.

Awesome checking it out now! Thanks Stu!

phillsatellite (approved):

dariusc93 (approved):

PR #174, opened 2023-01-24
feat(DragAndDrop): Avoid open a file in the app when drop it

### LINKED ISSUE #170: Drag/Dropping a file anywhere on the App will open the file
**Issue:** When you are in the app and you drag a file into the app it will instead open and preview the file 

**Screen recording:**

https://user-images.githubusercontent.com/93608357/214139629-59ea5176-69b2-4b94-b1dc-60a3c88cfb97.mov

lgmarchi:
### 1. Avoid open a file in the app when drop it

https://user-images.githubusercontent.com/63157656/214298242-515f74b9-074a-489f-9416-0985b52b6989.mov

### 2. Drag and drop function will be implemented later
- 

- Resolve #170

### DIFF
--- a/ui/src/main.rs
+++ b/ui/src/main.rs
@@ -270,7 +270,11 @@ fn main() {
     <body style="background-color:rgba(0,0,0,0);"><div id="main"></div></body>
     </html>"#
                     .to_string(),
-            ),
+            )
+            .with_file_drop_handler(|_w, drag_event| {
+                logger::debug(format!("Drag Event: {:?}", drag_event).as_str());
+                true
+            }),
     )
 }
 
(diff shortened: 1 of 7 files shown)


ui/src/main.rs
```diff
@@ -270,7 +270,11 @@ fn main() {
     <body style="background-color:rgba(0,0,0,0);"><div id="main"></div></body>
     </html>"#
                     .to_string(),
-            ),
+            )
+            .with_file_drop_handler(|_w, drag_event| {
```
dariusc93, line 274: Does this work across platform? I remember in the previous implementation of uplink that there was an issue with the file handler while working on linux. Not sure if thats still the case though

dariusc93 (approved):
LGTM. Did leave a question but shouldnt be blocking.

ui/src/main.rs
lgmarchi: Work in all platforms, problems with platforms start when we start to use dragover, drop, dragenter functions from HTML elements.

phillsatellite (approved):

PR #176, opened 2023-01-24
update(ReplyMessage): Fix platform and status for both users when reply message

### LINKED ISSUE #166: User Status will always be "online/mobile" in reply modal
**Issue:** When you are replying to someone, no matter who it is the online status will appear as "Online and on mobile device"

**Screen recording:**

https://user-images.githubusercontent.com/93608357/214102392-e8192d51-a78d-44cd-9e15-7d091d52b6ea.mov

lgmarchi:
### 1. Fix platform and status for both users when reply message

https://user-images.githubusercontent.com/63157656/214344515-cf44c591-4671-4c43-aa35-dfa85eb7221a.mov

- 

- Resolve #166

### DIFF
--- a/ui/src/components/chat/compose.rs
+++ b/ui/src/components/chat/compose.rs
@@ -2,13 +2,13 @@ use std::rc::Rc;
 
 use dioxus::prelude::*;
 
-use kit::{layout::{topbar::Topbar, chatbar::{Chatbar, Reply}}, components::{user_image::UserImage, indicator::{Platform}, context_menu::{ContextMenu, ContextItem}, message_group::{MessageGroup, MessageGroupSkeletal}, message::{Message, Order}, user_image_group::UserImageGroup}, elements::{button::Button, tooltip::{Tooltip, ArrowPosition}, Appearance}, icons::Icon};
+use kit::{layout::{topbar::Topbar, chatbar::{Chatbar, Reply}}, components::{user_image::UserImage, indicator::{Platform, Status}, context_menu::{ContextMenu, ContextItem}, message_group::{MessageGroup, MessageGroupSkeletal}, message::{Message, Order}, user_image_group::UserImageGroup}, elements::{button::Button, tooltip::{Tooltip, ArrowPosition}, Appearance}, icons::Icon};
 
 use dioxus_desktop::use_window;
 use shared::language::get_local_text;
 
 
-use crate::{state::{State, Action, Chat, Identity, self}, components::{media::player::MediaPlayer}, utils::{format_timestamp::format_timestamp_timeago, convert_status, build_participants}, logger};
+use crate::{state::{State, Action, Chat, Identity, self}, components::{media::player::MediaPlayer}, utils::{format_timestamp::format_timestamp_timeago, convert_status, build_participants, build_user_from_identity}, logger};
 
 use super::sidebar::build_participants_names;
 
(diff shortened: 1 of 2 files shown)


phillsatellite (approved):

ui/src/components/chat/compose.rs
```diff
@@ -343,25 +343,34 @@ fn get_chatbar(cx: Scope<ComposeProps>) -> Element {
...
+                let participants = build_participants(&data.active_chat.participants.clone());
+
                 cx.render(rsx!(
-                    active_chat.clone().replying_to.map(|msg| rsx!(
+                    active_chat.clone().replying_to.map(|msg|
+                    {
+                    let our_did = state.read().account.identity.did_key();
+                    let their_did = msg.sender();
+                    let (platform, status) = if our_did == their_did {
+                        (participants[1].platform, participants[1].status)
```
dariusc93: Would be careful when using an index like that. Would recommend using something like `participants.get(1)` and check to see if that is `Some` or `None`. Just mostly to prevent any possible panic due to it being out of bounds for whatever reason (though we can assume/expect it to be more than 1 participant here with any panics being considered a bug, but one can sometimes gracefully handle it without a panic)

dariusc93 (approved):

ui/src/components/chat/compose.rs
```diff
@@ -343,25 +343,34 @@ fn get_chatbar(cx: Scope<ComposeProps>) -> Element {
             )),
             with_replying_to: data.map(|data| {
                 let active_chat = data.active_chat.clone();
+                let participants = build_participants(&data.active_chat.participants.clone());
+
                 cx.render(rsx!(
-                    active_chat.clone().replying_to.map(|msg| rsx!(
+                    active_chat.clone().replying_to.map(|msg|
+                    {
+                    let our_did = state.read().account.identity.did_key();
```
sdwoodbury, line 350: what if instead of checking `our_did` and `their_did`, we took the vector of participants and called `vec.retain(|p| p.did_key() == msg.sender())` and then called `vec.iter().next()` to get the first item in the filtered list. this eliminates the need to know who is where in the vector (which could change, especially for group messages)

ui/src/components/chat/compose.rs
```diff
@@ -343,25 +343,34 @@ fn get_chatbar(cx: Scope<ComposeProps>) -> Element {
...
+
+                    rsx!(
                         Reply {
                             label: get_local_text("messages.replying"),
-                            remote: {
-                                let our_did = state.read().account.identity.did_key();
-                                let their_did = msg.sender();
-                                our_did != their_did
-                            },
+                            remote: our_did != their_did,
```
sdwoodbury: `remote` could become `our_did != msg.sender()` if the variable for `their_did` is done away with

ui/src/components/chat/compose.rs
lgmarchi: Perfect Darius, thanks!

ui/src/components/chat/compose.rs
lgmarchi: Good, I will change it

ui/src/components/chat/compose.rs
lgmarchi: Perfect!

ui/src/components/chat/compose.rs
lgmarchi: Improved!

ui/src/components/chat/compose.rs
lgmarchi: Improved!

PR #177, opened 2023-01-24
feat(input): allow clearing the input element

sdwoodbury:
### What this PR does 📖

- The chatbar needs to be cleared when a message is sent. the Input element doesn't allow for that. this PR fixes that
- the Input element used to have an option for a default value. that field has been replaced with a `UseRef<String>` for the actual value. Anything wishing to create an Input element with a default value should create the `value` field as follows: `use_ref(cx || String::from(<default value>)`

- Resolve #


sdwoodbury:
i tried calling `cx.needs_update()` and `state.write()` in the callbacks used to submit chats but those didn't cause the chatbar to clear. i kind of would have expected the entire UI to re-render and start over with an empty input box but for some reason, that doesn't happen. at least not with mock data.

phillsatellite (approved):

dariusc93 (approved):

PR #190, opened 2023-01-25
feat(chats): send a message

sdwoodbury:
### What this PR does 📖

- enables Uplink to send a message, and receives incoming messages for each conversation. 
- fakes sending messages when mock data is used
- allows the Input element to be cleared via an optional flag 
- turned Chat::messages into a VecDeque in anticipation of pagination, though the version of warp in use only supports fetching all messages (the function to get a range of messages just gets all messages and then discards the unwanted ones). 
- refactored the functions which convert a DID to an Identity to not take ownership of the DID. this eliminates the need to clone the variable. 
- reactions and replies are yet to come. 

- the big thing is this file: `ui/src/warp_runner/conv_stream.rs`. it creates a thread to listen for incoming messages for each conversation. the threads are aborted when `conv_stream::Manager` is dropped 

- Resolve #


sdwoodbury:
old commits in this branch were cherry-picked into other branches. only commits starting from [13c219b](https://github.com/Satellite-im/Uplink/pull/190/commits/13c219bdc1fbb665b9a9569a7877d28f4f2ac557) are relevant

phillsatellite:
This is so cool! I did find some issues here and there that if we want to create into new tickets thats cool with me! 

What I found:
- Sometimes when User A sends 3-4 messages the 3rd or 4th will not appear on User B's side

https://user-images.githubusercontent.com/93608357/214621200-25f14256-3ea9-43dd-b82e-fba877efa52f.mov

- After sending multiple messages will break the typing until user clicks out and back into the chat again 

https://user-images.githubusercontent.com/93608357/214621573-ecff91c7-d708-4ec2-94f3-e21bccc84a67.mov

ui/src/warp_runner/mod.rs
```diff
@@ -242,8 +263,8 @@ async fn warp_initialization(
     .map(|ct| Box::new(ct) as Storage)?;
 
     let mut rg_config = RgIpfsConfig::production(&path);
-    // Used to prevent broadcasting `ConversationCreated` on the sender side
-    rg_config.store_setting.disable_sender_event_emit = true;
+    // setting to `true` will prevent broadcasting `ConversationCreated` on the sender side
+    rg_config.store_setting.disable_sender_event_emit = false;
```
dariusc93: This can be removed if this will be false. Do note if youre getting the conversation from `RayGun::create_conversation` while this is also false will emit `RayGunEventKind::ConversationCreated` unless that been changed so it would be handled in the event stream.

ui/src/components/chat/compose.rs
```diff
@@ -324,19 +326,101 @@ fn get_messages(cx: Scope<ComposeProps>) -> Element {
...
+            while let Some((msg, chat_id)) = rx.next().await {
+                let conv_id = match chat_id {
+                    Some(c) => c,
+                    None => {
+                        continue;
+                    }
+                };
+
+                // don't send empty messages
+                if msg.is_empty() {
```
dariusc93: Could probably merge this with the statement below or remove it altogether since the condition below should also return false on an empty vector

ui/src/components/chat/compose.rs
```diff
@@ -324,19 +326,101 @@ fn get_messages(cx: Scope<ComposeProps>) -> Element {
...
+                    continue;
+                }
+
+                // this is also an empty message
+                if msg.iter().all(|line| line.is_empty())  {
```
dariusc93: Should probably trim whitespace before checking if its empty (though in warp it does check there as well but doesnt hurt as a precaution)
```suggestion
                if msg.iter().all(|line| line.trim().is_empty())  {
```

ui/src/warp_runner/conv_stream.rs
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    pub async fn add_stream(
+        &mut self,
+        conv_id: Uuid,
+        messaging: &mut Messaging,
+    ) -> Result<(), warp::error::Error> {
+        let ch = self.ch.clone();
+        let mut stream = messaging.get_conversation_stream(conv_id).await?;
+        let t = tokio::task::spawn(async move {
```
dariusc93, line 31: :+1:

sdwoodbury:
@phillsatellite  I found an odd crash when using uplink but it was fixed when I deleted `state.json`. i suspect that something got messed up when I used mock data and then switched to not using mock data. after deleting state.json and restarting, i wasn't able to reproduce the bug with the input box. but there may be a bug related to mock data and state.json here for me to fix.

about the messages not being sent - Uplink updates the sender side in response to a warp event. if the message isn't received, that's due to warp/ipfs, not uplink. i'm not sure what else I can do about this one.

InfamousVague:
> @phillsatellite I found an odd crash when using uplink but it was fixed when I deleted `state.json`. i suspect that something got messed up when I used mock data and then switched to not using mock data. after deleting state.json and restarting, i wasn't able to reproduce the bug with the input box. but there may be a bug related to mock data and state.json here for me to fix.
> 
> about the messages not being sent - Uplink updates the sender side in response to a warp event. if the message isn't received, that's due to warp/ipfs, not uplink. i'm not sure what else I can do about this one.

We could ping pong maybe in the future, and if we don't get a pong in `n` where `n` is how long we expect on average a message will take to send (optimally based programmatically on network conditions and mean response times of past messages or something). 

Edit: But this should really be part of Warp. Instead the UI should just ask that Warp supplies the status of a message we can check (to see if it's been received or not) and we can do a simpler curve to poll warp to see if the recipient got the message or not, after `n` we'll quit trying. (i.e. each time we try programmatically we wait a bit longer before trying again so we're not just hammering warp methods.) @dariusc93 tagging for thoughts

dariusc93:
> > @phillsatellite I found an odd crash when using uplink but it was fixed when I deleted `state.json`. i suspect that something got messed up when I used mock data and then switched to not using mock data. after deleting state.json and restarting, i wasn't able to reproduce the bug with the input box. but there may be a bug related to mock data and state.json here for me to fix.
> > about the messages not being sent - Uplink updates the sender side in response to a warp event. if the message isn't received, that's due to warp/ipfs, not uplink. i'm not sure what else I can do about this one.
> 
> We could ping pong maybe in the future, and if we don't get a pong in `n` where `n` is how long we expect on average a message will take to send (optimally based programmatically on network conditions and mean response times of past messages or something).
> 
> Edit: But this should really be part of Warp. Instead the UI should just ask that Warp supplies the status of a message we can check (to see if it's been received or not) and we can do a simpler curve to poll warp to see if the recipient got the message or not, after `n` we'll quit trying. (i.e. each time we try programmatically we wait a bit longer before trying again so we're not just hammering warp methods.) @dariusc93 tagging for thoughts

There is a function call `RayGun::message_status` which would tell one if it been sent (or in this case for now isnt apart of queue since items are queued if they cannot be sent for whatever reason) or not (with another field used to tell if it been delivered or not, but that isnt used internally at this moment but will be in the future) . 

For the case of the video above, since that is done locally, it should be using mdns instead of doing it over the network unless mdns is being blocked by the firewall on the system (which 9/10 isnt the case since many systems would have mdns enabled by default). If it doesnt use mdns, it will attempt to go over the network but because the local ip is apart of it, it may try those addresses when connecting thus bypassing going through any relays. 

I will have a quick look here shortly though to see rather that is the case but @phillsatellite can you try again and switch to another part of the UI when nothing is received on B client and switch back to the chat just to make sure its updating like it should?  

EDIT: Just as a note, the functionality to `RayGun::message_status` may change a bit in the future but could still be used to determine if a message have been sent or not.

phillsatellite:
@dariusc93 Saw your message while testing and gave it a shot, and after going to different page and back UI still didn't get the most recent messages. I also tried to favorite that chat thinking maybe that would get them to appear. 

https://user-images.githubusercontent.com/93608357/214718855-e18774a4-9504-40db-8732-6ef2a4c4699d.mov

and @sdwoodbury I also verified the input bug was fixed with most recent changes 👍 

I also found when you send a long message it will break the UI and let user scroll horizontally in Sidebar and Chat but since its pretty unrelated to the scope of this PR and will need work on mobile as well I figured it'd be best to make that a new issue after this PR is merged just wanted to mention it 👍 
 
https://user-images.githubusercontent.com/93608357/214719924-e1aa6a1b-892d-4b2a-998e-bc1a617d233c.mov

dariusc93:
Done some testing and havent been able to reproduce the issue phil has but after some background testing and talking to him, it may its not using mdns for local discovery on his computer, which might just be a system configuration on his mac, and if it is going through the relay, it may not be establishing a direct connection (although it should be seeing and connecting from the local ips thats on dht, that a separate thing ill be checking with him at a later point to determine that) so it may not skipping out receiving some data since the relay its going through may not allow enough data to go through. 

This isnt a blocker to this pr though.

dariusc93 (approved):

phillsatellite (approved):

phillsatellite:
Still had a few missing messages here and there this morning but since that seems to only be a problem on my end I'll create a ticket to do further investigation after this is merged in

PR #192, opened 2023-01-25
feat(ScrollPosition): Return scroll position

### LINKED ISSUE #178: detect scroll position in Compose window
paginating messages requires knowing the scroll position. Dioxus has an `onscroll` event which returns nothing useful. However, in the callback for `onscroll`, [use_eval](https://docs.rs/dioxus-desktop/latest/dioxus_desktop/fn.use_eval.html) could be used to check the scroll position. 

task: 
1) create a JavaScript function to return the scroll position
2) call this from the `onscroll` callback and obtain the result from the `EvalResult` 
3) for proof of concept, display the result in the terminal via `println!` or equivalent. the callback would be inside of this [div](https://github.com/Satellite-im/Uplink/blob/dev/ui/src/components/chat/compose.rs#L257)

lgmarchi:
### 2. Onscroll function doesn't work on MacOS
- 

- Resolve Satellite-im/Uplink-Backlog#14

### DIFF
--- a/ui/src/components/chat/compose.rs
+++ b/ui/src/components/chat/compose.rs
@@ -247,7 +247,9 @@ fn get_messages(cx: Scope<ComposeProps>) -> Element {
     window.eval(script);
     let scroll_position_script = include_str!("./scroll_position.js");
 
-    use_future(cx, (), |_| {
+    // pass the id of the active_conversation so that the use_future restarts when chats are changed. 
+    let id = cx.props.data.as_ref().map(|d| d.active_chat.id).clone();
+    use_future(cx, &id, |_| {
         to_owned![script, scroll_position_script, window];
         async move {
             window.eval(script.as_str());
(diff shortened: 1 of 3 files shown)


InfamousVague:
for this feature, what requirement are we hoping to use this for? I think we can do this without needing to add JS here

lgmarchi:
> for this feature, what requirement are we hoping to use this for? I think we can do this without needing to add JS here

Scrollposition will be necessary to do pagination

InfamousVague (changes requested):
Let's wait a bit on this, I'm nervous this could lead to performance issues

sdwoodbury:
we could make this a draft and keep it on hand for later

InfamousVague:
Closing the PR, let's keep the branch in case we need it. No need to keep the PR open or in draft though.

PR #196, opened 2023-01-26
chore(docs): Use tasks instead of threads 

dariusc93:
### What this PR does 📖

- Async tasks (either from tokio, or another async runtime) are not the same as system threads. They can be seen as lightweight threads, coroutines (if ever used anything similar in another languages like kotlin, go, etc), but not real threads so we should not confuse the two in this case.

- N/A


sdwoodbury (approved):
Thank you for the correction!

PR #199, opened 2023-01-26
feat(logger): use log crate

sdwoodbury:
### What this PR does 📖

- allows for easier logging. eliminates the need to pass `&format!("...{}", arg)` to a `logger` function. instead can do this: `log::debug!("...{}", arg)`. 
- still captures logs for the `debug_logger` element 

- Resolve #


tooshel:
How do you set the log level? And where do you see the logs, the console?

InfamousVague (approved):

lgmarchi:
This is a good PR, and will ease way to use logs. We are going to have strong logger feature, mixing log crate with our implementation. 
Talking to Stu, save logs to debug.log is not working, so this change was requested.

PR #203, opened 2023-01-27
chore(tesseract-helper): added helper function to see if local accoun…

WanderingHogan:
…t exists

### What this PR does 📖
This add's a helper that just tells us if the local account even exists. I want to use it in the UI in other places but its taking me longer to figure out, so it would be best to get this in if possible.

On the UI side, this method shows or hides the create account button based on if there is a local account somewhere. So if you have the uplink folder, no button, if you clear it out, the create account button is there @phillsatellite 

- Resolve #

### DIFF
--- a/ui/src/warp_runner/manager/mod.rs
+++ b/ui/src/warp_runner/manager/mod.rs
@@ -83,10 +83,8 @@ pub async fn run(mut warp: Warp, notify: Arc<Notify>) {
     logger::debug("terminating warp_runner thread");
 }
 
-const KEYSTORE_PATH_NAME: &str = ".keystore";
-
 fn init_tesseract() -> Tesseract {
-    let tess_path = STATIC_ARGS.warp_path.join(&KEYSTORE_PATH_NAME);
+    let tess_path = STATIC_ARGS.warp_path.join(".keystore");
     match Tesseract::from_file(&tess_path) {
         Ok(tess) => tess,
         Err(_) => {
--- a/ui/src/warp_runner/manager/mod.rs
+++ b/ui/src/warp_runner/manager/mod.rs
@@ -99,10 +99,6 @@ fn init_tesseract() -> Tesseract {
     }
 }
 
-pub fn account_exists() -> bool {
-    return STATIC_ARGS.warp_path.join(&KEYSTORE_PATH_NAME).exists();
-}
-
 async fn warp_initialization(
     tesseract: Tesseract,
     experimental: bool,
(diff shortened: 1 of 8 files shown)


sdwoodbury:
checking for the keystore file seems fine. instead of adding functions in the `manager` and `WarpRunner `structs, i suggest having a function within the `warp_runner` module (outside of any struct). just because it's a helper function that only relies on static data and doesn't use anything from `WarpRunner` or `Manager`. 

it could be called like this: `warp_runner::account_exists()`

phillsatellite (approved):

phillsatellite:
Lucas found a bug that went right over my head when testing, after creating an account if you log out and then log back in again using a different PIN (and hitting the enter key) it will take you to create username page

dariusc93:
As a temporary measure, what could be done is instead of trying for the file, to check for a specific key to see if it exist. In the old variant of uplink, i suggested using `Tesseract::exist` to see if `keypair` exist. If this key exist, then its a strong possibility that an account exist. I say "strong possibility" though because although the keypair does exist wouldnt always mean the account itself has been created. In this case though, checking it might be a better option to determine if an account needs to be created or not. 

EDIT: Here is how it was done in the old version https://github.com/Satellite-im/Uplink-Deprecated/blob/dev/src/components/prelude/unlock/mod.rs . Do note that later it still does check to see if `MultiPass::get_own_account` would return a error later on (which if it does after unlocking tesseract, even with that field existing, then one would need to use `MultiPass::create_identity`

ui/src/warp_runner/manager/mod.rs
```diff
@@ -83,8 +83,10 @@ pub async fn run(mut warp: Warp, notify: Arc<Notify>) {
     logger::debug("terminating warp_runner thread");
 }
 
+const KEYSTORE_PATH_NAME: &str = ".keystore";
```
sdwoodbury: this variable is no longer necessary now that warp_runner is being used to check if an account exists

WanderingHogan: gratuitous variable removed

WanderingHogan:
@dariusc93 - @sdwoodbury already had a helper function, `TesseractCmd::KeyExists` that made `Tesseract::exist` available. He went and made the changes here.

sdwoodbury (approved):

phillsatellite (approved):

InfamousVague (approved):

dariusc93 (approved):

PR #205, opened 2023-01-27
feat(reactions): add reactions to warp_runner

sdwoodbury:
### What this PR does 📖

- this PR allows for adding and removing reactions from messages
- the UI isn't done for this yet, so it must be tested in the debugger UI or by running uplink with the `debug` option. ex: `target/debug/ui --no-mock debug`

- Resolve #

### DIFF
--- a/ui/src/state/mod.rs
+++ b/ui/src/state/mod.rs
@@ -505,69 +505,63 @@ impl State {
 
     pub fn add_message_reaction(&mut self, chat_id: Uuid, message_id: Uuid, emoji: String) {
         let user = self.account.identity.did_key();
-        if let Some(conv) = self.chats.all.get_mut(&chat_id) {
-            // using any() here for an easy way to apply the Fn to the entire list
-            let res = conv.messages.iter_mut().any(|msg| {
-                if msg.id() != message_id {
-                    return false;
+        let conv = match self.chats.all.get_mut(&chat_id) {
+            Some(c) => c,
+            None => {
+                log::warn!("attempted to add reaction to nonexistent conversation");
+                return;
+            }
+        };
+
+        for msg in &mut conv.messages {
+            if msg.id() != message_id {
+                continue;
+            }
+
+            let mut has_emoji = false;
+            for reaction in msg.reactions_mut() {
+                if !reaction.emoji().eq(&emoji) {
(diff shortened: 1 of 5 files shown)


phillsatellite (approved):

ui/src/state/mod.rs
```diff
@@ -501,6 +502,68 @@ impl State {
...
+        let user = self.account.identity.did_key();
+        let conv = match self.chats.all.get_mut(&chat_id) {
+            Some(c) => c,
+            None => {
+                log::warn!("attempted to add reaction to nonexistent conversation");
+                return;
+            }
+        };
+
+        for msg in &mut conv.messages {
```
dariusc93, line 516: As I look at this section, You could probably get away with doing something like 

```rust
if let Some(msg) = conv
            .messages
            .iter()
            .position(|msg| msg.id() == message_id)
            .and_then(|index| conv.messages.get_mut(index))
        {
            let mut has_emoji = false;
            for reaction in msg.reactions_mut() {
                if !reaction.emoji().eq(&emoji) {
                    continue;
                }
                if !reaction.users().contains(&user) {
                    reaction.users_mut().push(user.clone());
                    has_emoji = true;
                }
            }

            if !has_emoji {
                let mut r = Reaction::default();
                r.set_emoji(&emoji);
                r.set_users(vec![user]);
                msg.reactions_mut().push(r);
            }
        }
```

Or instead you could omit `and_then` and switch the iterator to mutable and use `find` which should return a mutable message, if it is found. Eg

```rust
if let Some(msg) = conv
            .messages
            .iter_mut()
            .find(|msg| msg.id() == message_id)
        {
            let mut has_emoji = false;
            for reaction in msg.reactions_mut() {
                if !reaction.emoji().eq(&emoji) {
                    continue;
                }
                if !reaction.users().contains(&user) {
                    reaction.users_mut().push(user.clone());
                    has_emoji = true;
                }
            }

            if !has_emoji {
                let mut r = Reaction::default();
                r.set_emoji(&emoji);
                r.set_users(vec![user]);
                msg.reactions_mut().push(r);
            }
        }
```

dariusc93 (approved):
LTGM. Left some comments but its non blocking

PR #208, opened 2023-01-30
fix(login): reinitialize warp with unlocked tesseract 

sdwoodbury:
### What this PR does 📖

- i suddenly wasn't able to log into Uplink due to a cryptography error. it turns out that warp needs to be initialized with an unlocked tesseract (this may be new behavior). the fix was kind of  a hack - re-initializing warp in response to the Unlock and CreateAccount commands. if anyone has a better idea i'm open to suggestions I say it's a hack because the current design is potentially confusing: warp_runner initializes Warp with a locked tesseract and then accepts commands which will all fail until it's reinitialized with an unlocked tesseract. 
- I also improved the way logs are displayed in the terminal. this shouldn't affect how they are saved to file and thus have no effect on the debug_logger element. 

- Resolve #

### DIFF
--- a/ui/Cargo.toml
+++ b/ui/Cargo.toml
@@ -41,6 +41,7 @@ soloud = "1.0.2"
 open = "3.2.0"
 wry = { version = "0.23.4" }
 derive_more = "0.99"
+colored = "2.0.0"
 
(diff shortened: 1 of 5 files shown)


sdwoodbury:
nvm. not ready yet

PR #210, opened 2023-01-30
fix(login): fix "invalid signature" error on login

sdwoodbury:
### What this PR does 📖

- sometimes i got that error when logging in, especially when logging in a second time and especially when loading multiple conversations. it seems that the other devs didn't have that error today - no one else mentioned it. 
- I changed the flow of `warp_runner` to first only respond to commands related to login. 
- I got paranoid about how multipass and raygun worked with teseract, and made sure they were only initialized with an unlocked tesseract. 
- i changed the logic for logins, to check for a "keypair" in tesseract when not creating a new account. 
- I  improved the way logs are displayed in the terminal. this shouldn't affect how they are saved to file and thus have no effect on the debug_logger element.
- I made the `trace` debugging option save warp logs in a single file. it isn't ideal but helps debug problems involving warp. 

- Resolve #

### DIFF
--- a/ui/Cargo.toml
+++ b/ui/Cargo.toml
@@ -41,6 +41,7 @@ soloud = "1.0.2"
 open = "3.2.0"
 wry = { version = "0.23.4" }
 derive_more = "0.99"
+colored = "2.0.0"
 
 notify-rust = { version = "4.6.0", default-features = false, features = ["d"] }
 once_cell = "1.13"
--- a/ui/src/logger/mod.rs
+++ b/ui/src/logger/mod.rs
@@ -46,11 +46,13 @@ impl std::fmt::Display for Log {
 #[derive(Debug, Clone)]
 pub struct Logger {
     log_file: String,
+    warp_file: String,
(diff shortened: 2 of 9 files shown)


phillsatellite (approved):

phillsatellite:
Reason for removing QA Approved was because when I tested, I tested with no mock data and it saved all my messages and friends I had added but I didnt know Matt wanted it to also save everything in the mock data as well

sdwoodbury:
> Reason for removing QA Approved was because when I tested, I tested with no mock data and it saved all my messages and friends I had added but I didnt know Matt wanted it to also save everything in the mock data as well

I think the mock data is a separate issue. This would address a bug where chats and friends fail to load (due to invalid signature), which would send the user to a never-ending loading page.

dariusc93:
Just as a little note, MultiPass (or in this case, warp-mp-ipfs) is the only one that interacts with tesseract. RayGun doesnt make use of it.

dariusc93 (approved):

PR #211, opened 2023-01-31
feat(typing): add backend for typing indicator

sdwoodbury:
### What this PR does 📖

- sends and receives events to trigger the typing indicator 
- introduces a task to refresh message timestamps (unrelated to the typing indicator)
- adds documentation to `main::app()`
- copied the `scss` code for the typing indicator animation from the old Uplink app. 

- Resolve #

### DIFF
--- a/ui/src/components/chat/compose.rs
+++ b/ui/src/components/chat/compose.rs
@@ -397,7 +397,20 @@ fn get_chatbar(cx: Scope<ComposeProps>) -> Element {
     let input = use_ref(cx, Vec::<String>::new);
     let should_clear_input = use_state(cx,|| false);
     let active_chat_id = data.as_ref().map(|d| d.active_chat.id);
+
+    // todo: use this to render the typing indicator
+    let  _users_typing = match active_chat_id.and_then(|id| state.read().chats.all.get(&id).cloned()) {
+        Some(chat) => {
+            let users_typing = chat.participants.iter().filter(|x| chat.typing_indicator.contains_key(&x.did_key())).collect::<Vec<_>>();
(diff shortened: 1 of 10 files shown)


ui/src/components/chat/compose.rs
```diff
@@ -382,8 +397,21 @@ fn get_chatbar(cx: Scope<ComposeProps>) -> Element {
     let input = use_ref(cx, Vec::<String>::new);
     let should_clear_input = use_state(cx,|| false);
     let active_chat_id = data.as_ref().map(|d| d.active_chat.id);
+
+    // todo: use this to render the typing indicator
+    let  _users_typing = match active_chat_id.and_then(|id| state.read().chats.all.get(&id).cloned()) {
```
dariusc93: Not sure if its intended to have this expanded for future implementation of the rendering of the indicator but you could do (though may  be in a cleaner format once rustfmt is used)

```rust
let  _users_typing = active_chat_id.and_then(|id| state.read().chats.all.get(&id).cloned()).map(|chat| chat.participants.iter().filter(|x| chat.typing_indicator.contains_key(&x.did_key())).map(|x| x.username()).collect::<Vec<_>>());
```

dariusc93 (approved):

phillsatellite (approved):

ui/src/components/chat/compose.rs
sdwoodbury: i did it for readability. though now i noticed that my `cargo fmt` on save stopped working.

sdwoodbury: i'll have to fmt everything in a separate PR. i've been thinking about doing that for a while now.

PR #212, opened 2023-01-31
feat(Constellation): Start constellation in Uplink adding new folder

lgmarchi:
### 2. 2 first functions from Constellation, is to get items from current directory and add new folder 
a. We can use mock data and add new folder just to UI, but it will lost data when we restart app 

https://user-images.githubusercontent.com/63157656/215869276-c3138011-177a-4cda-8c08-05a53769f07c.mov

b. We can use real data, adding new folder will upload this to FileSystem, and show it again when restart app

https://user-images.githubusercontent.com/63157656/215869094-9d461f49-77a7-46d0-8239-4452b98bf92a.mov

- 

- Resolve #


kit/src/elements/folder/mod.rs
```diff
@@ -78,15 +75,19 @@ pub fn Folder<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
...
                 },
                 with_rename.then(|| rsx! (
-                    Input {
-                        disabled: disabled,
-                        placeholder: placeholder,
-                        // todo: use is_valid
-                        onreturn: move |(s, _is_valid)| emit(&cx, s)
-                    }
+                        Input {
+                            id: Uuid::new_v4().to_string(),
```
sdwoodbury, line 94: there might not be a need to generate a random id. without this, does the list of files not update? 
if it works without the `id` field, that field should be removed since we're not using the id to prevent re-renders

ui/src/state/mod.rs
```diff
@@ -53,6 +54,8 @@ pub struct State {
     #[serde(default)]
     pub friends: friends::Friends,
     #[serde(default)]
+    pub items: items::Items,
```
sdwoodbury: i suggest picking a different name for items. i think storage would be more descriptive.  same thing for the file `state/items.rs` -> `state/storage.rs`

also, until there is data which persists, this field could be `#[serde(skip)]` so long as it implements `Default`

also, unrelated - further down `hooks` is `#[serde(skip_serializing, skip_deserializing)]`. i know this wasn't you. i thought I changed that to `#[serde(skip)]`, which i believe is equivalent. we may want to take this opportunity to change it back

phillsatellite:
Hi Lucas! Awesome step forward! Just a few things I found on here that I think would perfect it. If you want to fix these in a new PR just lemme know and I'll add some tickets for them! 

1. The name of the folder sits to the left when you save it, I think it might be nice to have the title of the folder appear centered. 

Screenshot 2023-01-31 at 3 17 17 PM

2. We might want to limit chars for the names of folders, right now User can add as many as theyd like and it will extend across the entire screen.

https://user-images.githubusercontent.com/93608357/215874158-8dc5be4e-4269-49ad-8b53-79edc1d5dd9b.mov

ui/src/state/items.rs
```diff
@@ -0,0 +1,27 @@
...
+
+// TODO: Properly wrap data which is expected to persist remotely in options, so we can know if we're still figuring out what exists "remotely", i.e. loading.
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+pub struct Items {
+    // becomes true when the items fields have been retrieved from Warp
+    #[serde(skip)]
+    pub initialized: bool,
+    // All items in the current directory.
+    #[serde(skip)]
+    pub all: Vec<warp::constellation::item::Item>,
```
sdwoodbury: I think that `all` may be unnecessary. it's only used in `join`, which could instead merge `self.directories` with `other.directories` and `self.files` with `other.files`. 

i looked closer at the `join` function for `chats` and `friends` and i believe it's now  unnecessary. it was useful at one time but should have been deleted. for example, `friends_init` defaults to true when mock data is enabled, causing that use_future would exit early, and `Friends::join` would never be called. 

it turns out that `Items::join` is also unnecessary.

ui/src/layouts/files.rs
```diff
@@ -30,13 +45,100 @@ pub fn FilesLayout(cx: Scope<Props>) -> Element {
...
+    }
+
+    let ch = use_coroutine(cx, |mut rx: UnboundedReceiver<ChanCmd>| {
+        to_owned![items_state, directories_list, files_list];
+        async move {
+            let warp_cmd_tx = WARP_CMD_CH.tx.clone();
+            while let Some(cmd) = rx.next().await {
+                match cmd {
+                    ChanCmd::AddNewFolder(folder_name) => {
+                        if STATIC_ARGS.use_mock {
```
sdwoodbury: to reduce nesting, i recommend handling the mock case for `AddNewFolder` outside of the `use_coroutine`

ui/src/layouts/files.rs
```diff
@@ -30,13 +45,100 @@ pub fn FilesLayout(cx: Scope<Props>) -> Element {
...
+                                }
+                                Err(e) => {
+                                    log::error!("failed to add new folder conversation: {}", e);
+                                    continue;
+                                }
+                            }
+                        }
+                    }
+                    ChanCmd::GetItemsFromCurrentDirectory => {
+                        if STATIC_ARGS.use_mock {
```
sdwoodbury: same

ui/src/main.rs
```diff
@@ -507,6 +510,50 @@ fn app(cx: Scope) -> Element {
...
+                        if STATIC_ARGS.use_mock {
+                            state.write().items.join(items);
+                        } else {
+                            state.write().items = items.clone();
+                        }
+
+                        needs_update.set(true);
+                    }
+                    Err(e) => {
+                        logger::error(&e.to_string());
```
sdwoodbury: this works but we no longer need to do this. `log::error!("<some message> {}", e);` is now possible.

ui/src/layouts/files.rs
```diff
@@ -141,27 +243,52 @@ pub fn FilesLayout(cx: Scope<Props>) -> Element {
...
+                            aria_label: dir.name(),
+                        })
+                    }),
+                    files_list.read().iter().map(|file| {
+                        rsx!(File {
+                            text: file.name(),
+                            aria_label: file.name(),
+                        })
+                    }),
+                    if STATIC_ARGS.use_mock {
```
sdwoodbury: i'm fine with this but suspect that the mock data generation should instead go in the `testing/mock.rs` module

sdwoodbury (changes requested):
good work! I suggested some small changes.

ui/src/layouts/files.rs
```diff
@@ -30,13 +45,100 @@ pub fn FilesLayout(cx: Scope<Props>) -> Element {
...
+        async move {
+            let warp_cmd_tx = WARP_CMD_CH.tx.clone();
+            while let Some(cmd) = rx.next().await {
+                match cmd {
+                    ChanCmd::AddNewFolder(folder_name) => {
+                        if STATIC_ARGS.use_mock {
+                            directories_list
+                                .with_mut(|i| i.insert(0, Directory::new(&folder_name)));
+                        } else {
+                            let (tx, rx) = oneshot::channel::<Result<(), warp::error::Error>>();
```
dariusc93: Not a blocker but you usually dont need to define a type a head of time since the compiler would detect it when using it (eg `tx` is used in `rsp` field so it would detect the type that way).

ui/src/state/items.rs
```diff
@@ -0,0 +1,27 @@
+use serde::{Deserialize, Serialize};
+use warp::{constellation::directory::Directory, constellation::file::File};
+
+// TODO: Properly wrap data which is expected to persist remotely in options, so we can know if we're still figuring out what exists "remotely", i.e. loading.
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+pub struct Items {
```
dariusc93: Could you explain the need for `all` while also using `files` and `directories`? I wonder if some of the fields is necessary since in constellation, you can use `Directory` and get the list the items and resolve it into either a `File` or `Directory`. Containing a list of files, directories, and items may not be ideal since youre holding unnecessary data.

ui/src/warp_runner/manager/commands/constellation_commands.rs
```diff
@@ -0,0 +1,65 @@
...
+    let _ = storage.create_directory(folder_name, true).await?;
+    log::debug!("New directory created: {:?}", folder_name);
+    Ok(())
+}
+
+fn initialize_items(storage: &mut Storage) -> Result<Items, Error> {
+    let current_dir = storage.current_directory()?;
+    let items = current_dir.get_items();
+
+    let mut directories: Vec<Directory> = Vec::new();
```
dariusc93: if `directories` and `files` are to be used, you could iterate over `items` by doing something along the lines of

```rust
let directories = items.iter().filter_map(|item| item.get_directory().ok()).collect::<Vec<_>>();
```

Same with `files`, although I do feel splitting them up like this is counterproductive and would retain unnecessary data if you are also holding onto `items` too.

dariusc93 (commented):
Left some comments and a question

kit/src/elements/folder/mod.rs
lgmarchi: It is necessary, because when we open this field, there is a javascript code to focus on this input field, and for this, is necessary the id

ui/src/state/items.rs
lgmarchi: done

ui/src/state/items.rs
lgmarchi: I would just let a vec with all items as well to maybe in the future, presentation of files and folders be different, but for now I will delete this field

ui/src/warp_runner/manager/commands/constellation_commands.rs
lgmarchi: Done!

ui/src/state/mod.rs
lgmarchi: Done!

ui/src/layouts/files.rs
lgmarchi: It is asking to let type because of rx

ui/src/layouts/files.rs
lgmarchi: Done!

lgmarchi:
> Hi Lucas! Awesome step forward! Just a few things I found on here that I think would perfect it. If you want to fix these in a new PR just lemme know and I'll add some tickets for them!
> 
> 1. The name of the folder sits to the left when you save it, I think it might be nice to have the title of the folder appear centered.
> 
> Screenshot 2023-01-31 at 3 17 17 PM
> 
> 2. We might want to limit chars for the names of folders, right now User can add as many as theyd like and it will extend across the entire screen.
> 
>  Screen.Recording.2023-01-31.at.3.20.02.PM.mov

Phill, I changed to be 64 chars maximum, but folder name will be formatted to not over other names.

sdwoodbury (approved):
looks good to me. thanks for the changes!

ui/src/layouts/storage.rs
```diff
@@ -0,0 +1,297 @@
...
+    let free_space_text = get_local_text("files.free-space");
+    let total_space_text = get_local_text("files.total-space");
+    let storage_state: &UseState<Option<Storage>> = use_state(cx, || None);
+    let directories_list = use_ref(cx, || state.read().storage.directories.clone());
+    let files_list = use_ref(cx, || state.read().storage.files.clone());
+
+    let add_new_folder = use_state(cx, || false);
+
+    if let Some(storage) = storage_state.get().clone() {
+        if STATIC_ARGS.use_mock == false {
```
dariusc93:
```suggestion
        if !STATIC_ARGS.use_mock {
```

dariusc93 (approved):
LTGM. Left a suggestion but its non-blocking

phillsatellite (approved):

phillsatellite:
Hi Lucas! I noticed something new when I was re-testing everything. When I was going to switch to mobile view I noticed that when you turn on Dev Mode all the folders will be missing, is this intended or a bug? 

https://user-images.githubusercontent.com/93608357/216103590-0e643f01-8c14-403f-acf8-27105b70a694.mov

ui/src/layouts/storage.rs
lgmarchi: Man, I was thinking about this, is not more readable to use == false? 

I think it is easier to understand, I am not sure what way is better in good practices, i want your opinion too on this

lgmarchi:
> Hi Lucas! I noticed something new when I was re-testing everything. When I was going to switch to mobile view I noticed that when you turn on Dev Mode all the folders will be missing, is this intended or a bug?
> 
>  Screen.Recording.2023-02-01.at.11.28.30.AM.mov

I fixed bottom nav bar my friend!

lgmarchi:
> Hi

Fix folders missing when we update app.

phillsatellite (approved):

phillsatellite:
Put missing dev review for Lucas's most recent changes he added. Approved on my side 👍

kit/src/elements/folder/mod.rs
```diff
@@ -25,8 +26,21 @@ pub struct Props<'a> {
     loading: Option<bool>,
 }
 
-pub fn get_text(cx: &Scope<Props>) -> String {
-    cx.props.text.clone().unwrap_or_default()
+pub fn get_text(cx: &Scope<Props>) -> (String, String) {
+    let folder_name = cx.props.text.clone().unwrap_or_default();
+    let mut folder_name_formatted = folder_name.clone();
+
+    if folder_name_formatted.len() > 15 {
```
InfamousVague: can we make this > 15 a FOLDER_LEN constant at the top of the file?

ui/src/layouts/storage.rs
```diff
@@ -0,0 +1,302 @@
+use std::time::Duration;
+
```
InfamousVague, line 2: Is this file renamed to Storage for any specific reason or just matching warp? I'm okay with whatever just want to be consistent across the app.

InfamousVague (approved):

InfamousVague:
LGTM after the mentioned comments

ui/src/layouts/storage.rs
lgmarchi: Is just to use better name than I was using before, but we can change for another as well without problems, like FilesBrowser

kit/src/elements/folder/mod.rs
lgmarchi: Absolutely!

phillsatellite (approved):

PR #213, opened 2023-01-31
Top to bottom Notification system

InfamousVague:
WIP

- [x] **Bind `Configuration` to state.**  - This will allow us to update things in real time like we do with normal state values, whilst benefiting from the compatibility and ease of use of the config.
- [x] **Ensure that the config is properly loaded into state when application refreshes.**
- [x] **Add `Notification` numbers to state.** - This adds generic `uint`s to state representing the count of notifications that are "unseen" coming from a given featureset. This does not replace things like the unread indicator because without a very granular layout something like this will be un-needed and overly complex. This is used for things like application badges to represent a global number of things a user should check on in the app. We also shouldn't abuse this by default like most apps, otherwise it becomes meaningless and it's "always red".
- [x] **Wire `Notification` actions to make correct internal calls for sounds & alerts.**
- [x] **Update App Badge on Mac to display the count represented in `state.ui.notifications.total()`**
- [x] **Wire config settings to only notify when the user wants.**
- [ ] **Wire "Request Access" button in Notification Settings to prompt user for permissions to use notifications.** There is a prompt similar to this for Apple Ventura Desktop which will be the target for this PR. 
- [x] **Add a global toggle for all notifications - i.e. notifications on or off.**
- [x] **Add button under developer settings to push a test notification in 3 seconds.** (to give us time to focus some other app or minimize to test things)


InfamousVague:
>  Wire "Request Access" button in Notification Settings to prompt user for permissions to use notifications. There is a prompt similar to this for Apple Ventura Desktop which will be the target for this PR. 

Moving this last one to a new issue for now until I can learn more about how to handle this. The rest is ready after conflicts are resolved.

InfamousVague:
@sdwoodbury @dariusc93 I went ahead and merged in with dev but there are a lot of additions to things that have been moved so I'm not sure if I broke something but signup is not working for me on mac anymore so I likely broke something but if you two could give it a quick glance over it might be obvious.

phillsatellite:
The Sidebar is so cool lol I did find a few bugs though lemme know if you want any certain ones made into tickets and I'll write em up

1. Sometimes toggles arent saving when the user logs out and back in again 

https://user-images.githubusercontent.com/93608357/216237380-47b957b1-8212-4856-a004-92514db542c6.mov

2. User can also toggle on and off message/Friends/Settings even if the main notification toggle is off

https://user-images.githubusercontent.com/93608357/216237511-5f2d971b-ae4a-495c-9a53-e0ed9d7fb9dc.mov

3. This one is pretty weird. There seems to be 2 click areas for the sidebar. The one to the right thats right overtop the border of the sidebar is pretty  small and hard to click 

https://user-images.githubusercontent.com/93608357/216237667-ba7ec265-670f-45ae-b601-f2b782c9a946.mov

4. If you drag the sidebar out enough it will render behind the Uplink Logo. While its behind the logo it will not be clickable 

https://user-images.githubusercontent.com/93608357/216237751-360b7af9-1a93-4d02-b376-21319792aeb8.mov

5. It looks like on every page except for the chat page there is a border that stops the user from dragging the sidebar across the entire screen. I assumed the chat page not having a border was a bug.

https://user-images.githubusercontent.com/93608357/216237922-735f2e7a-440f-4cc3-8aeb-fe9601a20029.mov

6. I think this could be fixed in the ticket we have to have the user not be able to select the entire screen. When you drag the sidebar when there is text on the screen you will usually highlight all of the text.

https://user-images.githubusercontent.com/93608357/216238104-fead593d-f5f5-4352-9cee-d0facb2609b1.mov

7. And the last one is more of a question, do we want the size of the sidebar to persist through all the pages if the user adjusts it?

https://user-images.githubusercontent.com/93608357/216238462-6c53bdb6-a838-4e9e-a7dd-bd0541a47760.mov

InfamousVague:
> This one is pretty weird. There seems to be 2 click areas for the sidebar. The one to the right thats right overtop the border of the sidebar is pretty small and hard to click
This is because the "background" for the scrollbar is transparent, but what's actually happening is you're moving from the left to the right edge of the scrollbar track. At least that's my hypothesis, it could be checked out with a different theme created with non transparent scrollbars

https://user-images.githubusercontent.com/1770198/216239787-bd1ffbfc-4084-4003-b7cb-2ce9b0aaa5a3.mov



> 
    And the last one is more of a question, do we want the size of the sidebar to persist through all the pages if the user adjusts it?

Yes, let's add another ticket for this, Lucas is helping me out with that!

> 
    I think this could be fixed in the ticket we have to have the user not be able to select the entire screen. When you drag the sidebar when there is text on the screen you will usually highlight all of the text.
New ticket would be good, just assign it to me i can knock it out in a less unruly pr.

InfamousVague:
> 
    User can also toggle on and off message/Friends/Settings even if the main notification toggle is off

should be resolved

InfamousVague:
For the rest of the issues that are not related to the notifications itself we should consider moving to a new backlogged ticket, the resizing is just nice extra functionality and the dev logger itself is just for people doing dev work so we can put up with some small bugs or fix it ourselves.

shared/locales/en-US/main.ftl
```diff
@@ -116,6 +116,16 @@ settings-files = Files Settings
     .open-sync-folder = Open Sync Folder
     .open-sync-folder-description = Open the folder where your files are synced to.
 
+settings-notifications = Notifications
```
sdwoodbury, line 119: this should probably be implemented for the other languages

ui/src/components/settings/sidebar.rs
```diff
@@ -21,6 +21,7 @@ pub enum Page {
     Files,
     Privacy,
     Profile,
+    Notifications,
 }
 
 impl FromStr for Page {
```
sdwoodbury, line 27: will implementing `FromStr` for `Page` work for non english languages?

ui/src/components/settings/sidebar.rs
```diff
@@ -58,22 +60,17 @@ pub fn emit(cx: &Scope<Props>, e: Page) {
 pub fn Sidebar<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
```
sdwoodbury, line 60: I didn't realize al these routes were here. in `main.rs` there's a variable called `UPLINK_ROUTES` which I used to store the `to` fields for all routes. that way there's no need to hunt for routes or remember magic strings. It may be better to use that variable here.

sdwoodbury, line 60: it may be better to copy what `main.rs` does with the `UPLINK_ROUTES` variable - make a static struct in `sidebar.rs` with a field for each route, and use that in `FromStr` and for the `UIRoute`s.

also, instead of calling `emit(cx, ...)`, I think it's better to put the `match` statement in the `onnavigate` callback to get a `Page`, and pass that to the `onpress` event. i think that would be cleaner and easier to follow.

shared/locales/en-US/main.ftl
sdwoodbury, line 119: this should probably be implemented for the other languages

ui/src/layouts/settings.rs
```diff
@@ -64,12 +67,12 @@ pub fn SettingsLayout(cx: Scope<Props>) -> Element {
                     id: "content",
                     class: "full-width",
                     match to.get() {
```
sdwoodbury, line 69: you could get rid of some nesting here by moving the match statement out of the `cx.render()` call. the statement could return a `rsx!(SettingsPage {})` which could then be used in the `cx.render()` call,  replacing the `match` statement on line 69

ui/src/main.rs
```diff
@@ -167,7 +170,7 @@ struct Args {
 }
 
 fn copy_assets() {
-    let themes_dest = STATIC_ARGS.uplink_path.join("themes");
+    let themes_dest = STATIC_ARGS.uplink_path.join("");
```
sdwoodbury: joining an empty string does nothing

ui/src/main.rs
```diff
@@ -417,11 +426,35 @@ fn app(cx: Scope) -> Element {
             style { "{UIKIT_STYLES} {APP_STYLE} {theme}" },
             div {
                 id: "app-wrap",
+                onmouseenter: move |_| {
```
sdwoodbury: - why resize the window on mouse enter?
- may want to check that the config has changed before writing it out to a file
- use `log` instead of `println`
- `minimal_view` criteria on line 445 is different from the criteria on line 392

ui/src/state/notifications.rs
```diff
@@ -0,0 +1,148 @@
...
+pub struct Notifications {
+    pub friends: u32, // For notifications about new friends, friend requests and related CTAs.
+    pub messages: u32, // For notifications about new messages, mentions.
+    pub settings: u32, // For notifications about updates, issues and more.
+}
+
+impl Notifications {
+    pub fn new() -> Self {
+        // By default we'll say there are no notifications.
+        Self {
```
sdwoodbury, line 25: you could use `Default::default()` here.

ui/src/state/notifications.rs
```diff
@@ -0,0 +1,148 @@
...
+            crate::utils::sounds::Play(Sounds::Notification);
+        }
+    }
+
+    // Removes notification(s) from the specified kind.
+    pub fn decrement(&mut self, kind: NotificaitonKind, count: u32) {
+        match kind {
+            NotificaitonKind::FriendRequest => {
+                // Prevent underflow.
+                if count > self.friends {
```
sdwoodbury, line 79: this is fine. if you really want to be fancy, you can make it a one liner using [this](https://doc.rust-lang.org/std/primitive.u32.html#method.checked_sub)

`self.friends = self.friends.checked_sub(count).unwrap_or_default(); `

ui/src/main.rs
```diff
@@ -78,12 +80,13 @@ pub static STATIC_ARGS: Lazy<StaticArgs> = Lazy::new(|| {
...
         uplink_path: uplink_path.clone(),
         cache_path: uplink_path.join("state.json"),
+        mock_cache_path: uplink_path.join("mock-state.json"),
         config_path: uplink_path.join("Config.json"),
         warp_path: uplink_path.join("warp"),
         logger_path: uplink_path.join("debug.log"),
         typing_indicator_refresh: 5,
         typing_indicator_timeout: 6,
-        use_mock: !args.no_mock,
+        use_mock: args.no_mock,
```
sdwoodbury: this is a bug. it doesn't make mock be off by default. it inverts the meaning of the `--no-mock` flag.

ui/src/state/mod.rs
```diff
@@ -726,31 +719,41 @@ impl State {
...
+    fn load_mock() -> Self {
+        let contents = match fs::read_to_string(&STATIC_ARGS.mock_cache_path) {
+            Ok(r) => r,
+            Err(_) => {
+                return generate_mock();
+            }
+        };
+        serde_json::from_str(&contents).unwrap_or_else(|_| generate_mock())
     }
-}
```
sdwoodbury, line 752: looks like commit `5ac67e4adee9171ee682caa40b4f0952ef31294e` removed the handling for `MessageEvent::ReactionAdded` and `MessageEvent::ReactionRemoved`

ui/src/state/mod.rs
```diff
@@ -96,6 +101,104 @@ impl State {
         State::default()
     }
 
+    pub fn mock() -> Self {
```
sdwoodbury: since commit `11b88c3050e9278c39de38871df27168de6a14a7`, `State::new()` and `State::mock` are unnecessary. all that was handled by the `load()` function. it got broken during a merge with dev. using the `load()` function for everything prevents bugs like the one mentioned in `main.rs`

ui/src/main.rs
```diff
@@ -359,16 +363,20 @@ pub fn app_bootstrap(cx: Scope) -> Element {
     let mut state = if STATIC_ARGS.use_mock {
```
sdwoodbury: use `State::load()` instead of this `if` statement. looks like a merge conflict caused a reversion.

ui/src/main.rs
```diff
@@ -359,16 +363,20 @@ pub fn app_bootstrap(cx: Scope) -> Element {
...
         State::mock()
```
sdwoodbury: this will make mock data not persist across reloads.

sdwoodbury (changes requested):

phillsatellite:
Also experienced some bugs when retesting, after clearing data (and of course cargo update/clean) when user hits enter after creating username app will just close without any panics in terminal 

https://user-images.githubusercontent.com/93608357/216390556-1ee394b3-b14e-48cf-b49c-94451c7da56c.mov

ui/src/state/mod.rs
sdwoodbury: and merging with dev fixed it. this one is resolved.

sdwoodbury:
> @sdwoodbury @dariusc93 I went ahead and merged in with dev but there are a lot of additions to things that have been moved so I'm not sure if I broke something but signup is not working for me on mac anymore so I likely broke something but if you two could give it a quick glance over it might be obvious.

logging in on dev works and logging in in this branch works as of commit `dcd5a57fd8e9ec6e07bc8df73c5ac0afc4bf916f`, which was before that merge with dev. it might be worth checking out commit `dcd5a57fd8e9ec6e07bc8df73c5ac0afc4bf916f`, creating a new branch, and trying the merge again.  be sure to first run `cargo fmt` and fix all clippy warnings. that should reduce the number of conflicts.

PR #222, opened 2023-02-01
fix(logging): don't log passphrase

sdwoodbury:
### What this PR does 📖

- implement `Display` for `MultiPassCmd` so that passwords don't show up in debug logs 
- don't implement `Debug` for `MultiPassCmd` for the same reason
- implement `Display` for `WarpCmd`  and unimplement `Debug`
- fix clippy warnings project wide
- remove all `.expect()` calls when a `WarpCmd` couldn't be sent, because this would require `Debug`
- improve logic for friends and chats initialization 

- Resolve #


phillsatellite (approved):

ui/src/warp_runner/manager/commands/multipass_commands.rs
```diff
@@ -60,6 +60,21 @@ pub enum MultiPassCmd {
     },
 }
 
+// hide sensitive information from debug logs
+impl std::fmt::Display for MultiPassCmd {
```
dariusc93: You should manually implement `Debug` too since that would also show the password.

dariusc93 (changes requested):
In this case, we should probably implement `Debug` manually.

ui/src/warp_runner/manager/commands/multipass_commands.rs
dariusc93: Just as a note, this (`Debug` derive) is implemented on line 14.

sdwoodbury: this turned out to be a lot easier than i thought it would be . `derive_more` is such an awesome crate.

dariusc93 (approved):

PR #236, opened 2023-02-02
fix(sidebar): fix friends sidebar

### LINKED ISSUE #217: Mobile - Switching to mobile view on Friends will break UI
**Issue:** If the user is on Friends page and they switch to mobile view, all of the icons will still appear like the desktop icons

**Screen recording:**

https://user-images.githubusercontent.com/93608357/216107076-75f86060-0c37-40c2-9a1d-079ed0d88c63.mov

sdwoodbury:
### What this PR does 📖

- removes a buggy hook used to hide the sidebar in friends. it was called `first_render`. instead, default the sidebar to be hidden when using a minimal_view and use separate functions for regular and minimal views: `FriendsLayout` and `MinimalFriendsLayout`. 
- we may want to adopt this change in the other layouts, which still use the `first_render` hook. 

- Resolve #217

### DIFF
--- a/ui/src/layouts/friends.rs
+++ b/ui/src/layouts/friends.rs
@@ -7,6 +7,7 @@ use kit::{
     layout::topbar::Topbar,
 };
 use shared::language::get_local_text;
+use warp::logging::tracing::log;
 
 use crate::{
     components::{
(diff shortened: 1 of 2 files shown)


Jekrimo (approved):

dariusc93 (approved):

phillsatellite (approved):

InfamousVague (approved):
I think for now this is fine honestly, it removed the hacky code; Ideally in the future we can find a way to do this without code reuse.

PR #251, opened 2023-02-04
Add resize bindings for better "responsive" use cases. 

InfamousVague:
Also updates the focus state of the app when we click on and off. debugging issue with size not accurately updating.


InfamousVague:
This does not address anything in particular, rather it sets us up for the infamous "resize observer" we've been missing as well as a helpful `bool` representing weather or not the application is in focus or not which is super helpful for dispatching notifications.  I assume the issue with the resized size always being the same is an issue outside of the newly added code either in dioxus or wry.

ui/src/components/settings/sidebar.rs
```diff
@@ -146,6 +146,9 @@ pub fn Sidebar<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
                     routes: cx.props.route_info.routes.clone(),
                     active: cx.props.route_info.active.clone(),
                     onnavigate: move |route| {
+                        if state.read().configuration.config.audiovideo.interface_sounds {
```
sdwoodbury, line 149: this would be nicer if `configuration` has a function we could call instead of `configuration.config.audiovideo.interface_sounds`

sdwoodbury (changes requested):
looks fine. i highly suggest either 
1) adding some convenience functions to the config struct
2)  not wrapping the `Configuration` from `src/config.rs` inside of the `Configuration` struct from `src/state/configuration.rs`. 

calling `state.read().configuration.config` feels redundant.

ui/src/components/settings/sidebar.rs
InfamousVague: Let's think over this a bit more in the future, down to refactor this but I don't know if this is the PR for that. This simply adds helpers into state for real time events.

sdwoodbury (approved):
approving and letting the change to `config` be handled in another PR.

phillsatellite (approved):

WanderingHogan:
@InfamousVague this merge makes the sidebar hide/only render when you make the screen much larger

default:
Screenshot 2023-02-06 at 2 47 27 PM

resized:
Screenshot 2023-02-06 at 2 48 31 PM

PR #252, opened 2023-02-06
feat(Constellation): Upload files using file browser

lgmarchi:
### 1. Upload a file or multiple files using files browser 

https://user-images.githubusercontent.com/63157656/217002461-ba181a2b-ae0b-4b94-8afc-9078cbe144cb.mov

- 

- Resolve #

### DIFF
--- a/kit/src/elements/file/mod.rs
+++ b/kit/src/elements/file/mod.rs
@@ -1,9 +1,12 @@
+use std::{ffi::OsStr, path::PathBuf};
+
 use dioxus::prelude::*;
 
 use crate::{
     elements::input::Input,
     icons::{Icon, IconElement},
 };
+const MAX_LEN_TO_FORMAT_NAME: usize = 15;
 
 #[derive(Props)]
 pub struct Props<'a> {
@@ -23,8 +26,34 @@ pub struct Props<'a> {
     loading: Option<bool>,
 }
 
-pub fn get_text(cx: &Scope<Props>) -> String {
-    cx.props.text.clone().unwrap_or_default()
+pub fn get_text(cx: &Scope<Props>) -> (String, String) {
+    let file_name = cx.props.text.clone().unwrap_or_default();
+    let mut file_name_formatted = cx.props.text.clone().unwrap_or_default();
+    let file_extension = std::path::Path::new(&file_name)
+        .extension()
+        .unwrap_or_else(|| std::ffi::OsStr::new(""))
+        .to_str()
+        .unwrap()
+        .to_string();
+    let item = PathBuf::from(&file_name);
+    let file_stem = item
+        .file_stem()
+        .and_then(OsStr::to_str)
+        .map(str::to_string)
+        .unwrap_or_default();
+
+    if file_stem.len() > MAX_LEN_TO_FORMAT_NAME {
+        file_name_formatted = match &file_name.get(0..7) {
+            Some(name_sliced) => format!(
+                "{}...{}.{}",
+                name_sliced,
+                &file_stem[file_stem.len() - 2..].to_string(),
+                file_extension
+            ),
+            None => file_name.clone(),
+        };
+    }
+    (file_name, file_name_formatted)
 }
 
 pub fn get_aria_label(cx: &Scope<Props>) -> String {
@@ -45,9 +74,9 @@ pub fn emit_press(cx: &Scope<Props>) {
 
 #[allow(non_snake_case)]
 pub fn File<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
-    let text = get_text(&cx);
+    let (file_name, file_name_formatted) = get_text(&cx);
     let aria_label = get_aria_label(&cx);
-    let placeholder = text.clone();
+    let placeholder = file_name.clone();
     let with_rename = cx.props.with_rename.unwrap_or_default();
     let disabled = cx.props.disabled.unwrap_or_default();
 
@@ -79,7 +108,8 @@ pub fn File<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
                 )),
                 (!with_rename).then(|| rsx! (
                     label {
-                        "{text}"
+                        class: "file-name",
+                        "{file_name_formatted}"
                     }
                 ))
             }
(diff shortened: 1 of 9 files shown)


phillsatellite (approved):

phillsatellite:
This is awesome! I'm going to make a few tickets that I think we could use for Files

1. Add Upload bar
2. Add File Preview 
3. Add scrolling for files 
4. Im making a ticket for the files grid breaking while uploading files. I replicated this with mock data on dev by adding a bunch of folders

Screenshot 2023-02-06 at 11 25 22 AM

kit/src/elements/file/mod.rs
```diff
@@ -23,8 +26,34 @@ pub struct Props<'a> {
     loading: Option<bool>,
 }
 
-pub fn get_text(cx: &Scope<Props>) -> String {
-    cx.props.text.clone().unwrap_or_default()
+pub fn get_text(cx: &Scope<Props>) -> (String, String) {
+    let file_name = cx.props.text.clone().unwrap_or_default();
```
sdwoodbury, line 30: `Props::text` should not be an `Option`. you can't open a file without a name. though that mistake was not caused by this PR, now is a good time to fix it.

ui/src/layouts/storage.rs
```diff
@@ -108,7 +110,31 @@ pub fn FilesLayout(cx: Scope<Props>) -> Element {
...
+                        let (tx, rx) = oneshot::channel::<Result<Storage, warp::error::Error>>();
+
+                        if let Err(e) = warp_cmd_tx.send(WarpCmd::Constellation(
+                            ConstellationCmd::UploadFiles {
+                                files_path,
+                                rsp: tx,
+                            },
+                        )) {
+                            log::error!("failed to upload files {}", e);
+                            return;
```
sdwoodbury, line 128: this `return` should be a `continue`

ui/src/layouts/storage.rs
```diff
@@ -176,6 +202,13 @@ pub fn FilesLayout(cx: Scope<Props>) -> Element {
                                         text: get_local_text("files.upload"),
                                     }
                                 ))
+                                onpress: move |_| {
+                                    let files_local_path = match FileDialog::new().set_directory(".").pick_files() {
+                                        Some(path) => path,
+                                        None => return
```
sdwoodbury, line 208: this `return` should be a `continue`

ui/src/components/debug_logger/mod.rs
```diff
@@ -44,7 +41,7 @@ pub fn DebugLogger(cx: Scope) -> Element {
                         let log_datetime = fields.next().unwrap_or_default();
                         let log_level = fields.next().unwrap_or_default();
                         let log_message = fields.next().unwrap_or_default();
-                        let log_color = logger::get_color_string(Level::from_str(log_level).unwrap_or(Level::Debug));
+                        let log_level_string = format!("{}", log_level.trim().to_lowercase().as_str());
```
sdwoodbury, line 44: using `log_level.trim().to_lowercase();` instead of `format!` compiles just fine.

ui/src/warp_runner/manager/commands/constellation_commands.rs
```diff
@@ -64,3 +88,184 @@ fn initialize_items(warp_storage: &mut warp_storage) -> Result<uplink_storage, E
...
+                }
+                log::info!("{:?} file uploaded!", &filename);
+            }
+            Err(error) => log::error!("Error when upload file: {:?}", error),
+        }
+    }
+    get_items_from_current_directory(warp_storage)
+}
+
+fn verify_duplicate_name(
```
sdwoodbury, line 196: perhaps `verify_duplicate_name` should be renamed to `rename_if_duplicate`

ui/src/warp_runner/manager/commands/constellation_commands.rs
```diff
@@ -64,3 +88,184 @@ fn initialize_items(warp_storage: &mut warp_storage) -> Result<uplink_storage, E
...
+}
+
+async fn set_thumbnail_if_file_is_image(
+    warp_storage: warp_storage,
+    filename_to_save: String,
+) -> Result<String, Box<dyn std::error::Error>> {
+    let item = warp_storage
+        .current_directory()?
+        .get_item(&filename_to_save)?;
+    let parts_of_filename: Vec<&str> = filename_to_save.split('.').collect();
```
sdwoodbury, line 237: if a file name has no extension, i wonder if this will yield the file name instead. perhaps a `Path::extension()` should be used instead of  `name.split('.')`

ui/src/warp_runner/manager/commands/constellation_commands.rs
```diff
@@ -64,3 +88,184 @@ fn initialize_items(warp_storage: &mut warp_storage) -> Result<uplink_storage, E
...
+            "png" => IMAGE_PNG.to_string(),
+            "jpg" => IMAGE_JPEG.to_string(),
+            "jpeg" => IMAGE_JPEG.to_string(),
+            "svg" => IMAGE_SVG.to_string(),
+            &_ => "".to_string(),
+        },
+        None => "".to_string(),
+    };
+
+    if !file.is_empty() || !mime.is_empty() {
```
sdwoodbury, line 262: should this be `!file.is_empty() && !mime.is_empty()`?

ui/src/warp_runner/manager/commands/constellation_commands.rs
```diff
@@ -64,3 +88,184 @@ fn initialize_items(warp_storage: &mut warp_storage) -> Result<uplink_storage, E
...
+        let prefix = format!("data:{};base64,", mime);
+        let base64_image = base64::encode(&file);
+        let img = prefix + base64_image.as_str();
+        item.set_thumbnail(&img);
+        Ok(format_args!("{} thumbnail updated with success!", item.name()).to_string())
+    } else {
+        Err(Box::from(Error::InvalidItem))
```
sdwoodbury, line 269: suggest logging the reason for the error - either invalid mime or an empty file.

ui/src/warp_runner/manager/commands/constellation_commands.rs
sdwoodbury, line 196: i wonder if it's better to notify the user that the file exists and then 
- ask them if the file should be overwritten
- if yes: overwrite the file
- if no, fail the operation (let the user rename the file themself before retrying)

otherwise, if there is a file named "file.txt", the first duplicate would be "file 1.txt", next would be "file 1 2.txt", etc.

sdwoodbury (changes requested):

PR #271, opened 2023-02-07
fix(misc): Increase file desc limit

dariusc93:
### What this PR does 📖

- Fixes the "too many file" error that users might see on unix-like systems (eg linux, mac, etc). 

- N/A

### Additional comments 🎤
The error is due to the soft file descriptor limit being hit. Some systems like mac may have low soft limits (of around 256 or 512) while others may have it a bit higher. On linux, it varies on the distribution but the avg soft limit is around 2048.  With network applications, they will open a file descriptor for connections and sometimes may hold that open, which is why it may show that error. Some applications get around this by increasing the soft limits for that process only, which is what we are doing here in this case.

### DIFF
--- a/ui/src/main.rs
+++ b/ui/src/main.rs
@@ -191,6 +191,10 @@ fn copy_assets() {
 }
 
 fn main() {
+    // Attempts to increase the file desc limit on unix-like systems
+    // Note: Will be changed out in the future 
+    if fdlimit::raise_fd_limit().is_none() {}
+
     // configure logging
     let args = Args::parse();
     let max_log_level = if let Some(profile) = args.profile {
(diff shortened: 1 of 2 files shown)


tooshel (approved):

ui/src/main.rs
```diff
@@ -191,6 +191,10 @@ fn copy_assets() {
 }
 
 fn main() {
+    // Attempts to increase the file desc limit on unix-like systems
+    // Note: Will be changed out in the future
```
InfamousVague, line 195: Do you mind changing this to a `// TODO:` explaining why we should change it out? Just so others can know in the future. Looks good otherwise!

InfamousVague (changes requested):

ui/src/main.rs
dariusc93: Its more of a note than a todo because it may stay in. Changing it to a `TODO` would mean it will while there are still other factors to take into consideration before the approach i may do

PR #273, opened 2023-02-07
chore(sounds): comment out the use to fix the warning

tooshel:
### What this PR does 📖

- fix the new warning I ended up adding with my last PR

### DIFF
--- a/ui/src/main.rs
+++ b/ui/src/main.rs
@@ -42,7 +42,6 @@ use crate::layouts::unlock::UnlockLayout;
 use crate::state::ui::WindowMeta;
 use crate::state::Action;
 use crate::state::{friends, storage};
-//use crate::utils::sounds::Sounds;
 use crate::warp_runner::{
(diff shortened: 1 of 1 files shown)


ui/src/main.rs
```diff
@@ -42,7 +42,7 @@ use crate::layouts::unlock::UnlockLayout;
 use crate::state::ui::WindowMeta;
 use crate::state::Action;
 use crate::state::{friends, storage};
-use crate::utils::sounds::Sounds;
+//use crate::utils::sounds::Sounds;
```
InfamousVague: Let's just remove this line for now, We can always add it back in if we need it in the future. Especially for imports.

InfamousVague (changes requested):

ui/src/main.rs
tooshel: totally get it! fixed!

InfamousVague (approved):

dariusc93 (approved):

InfamousVague (approved):

InfamousVague:
I know I was preaching about process, but I don't know that we need @phillsatellite to review this one.

PR #274, opened 2023-02-07
chore(Extensions): Started bootstrapping extensions

InfamousVague:
### What this PR does 📖

- Adds ability for basic chatbar extensions to be added via a dynamic library.
- Should get us primed to build things like the emojipicker.

### Additional comments 🎤

There are a few key items this PR will address that needs to be completed:

- [x] Define Extension API 
- [x] Define Extension Example
- [x] Allow dropping of extensions into the extensions folder (`.uplink/extensions`).
- [x] Update the UI automatically to pull information from an extension into the settings.
- [ ] Enable ability to toggle extension on or off from the settings.
- [ ] Ensure there are proper methods for our extension to interact with state in a responsible mannor.


extension_example/src/lib.rs
```diff
@@ -0,0 +1,45 @@
+use dioxus::prelude::*;
+use extensions::*;
+use kit::{elements::button::Button, icons::Icon};
+
+// Exports the plugin using the registrar
+// You don't need to really worry about this but it is required.
+// Just change emoji's to the name of your extension in alpha-numeric snake case.
```
sdwoodbury: "emojis" 
`// Just change "emojis"  to the name of your extension in alpha-numeric snake case.`

extensions/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+        Self::default()
+    }
+
+    pub fn locate(&mut self, path: PathBuf) -> &Self {
+        log::debug!("Locating extensions");
+
+        // TODO: Search the extensions folder for files. Load them into self
+        let _ = fs::create_dir_all(&path);
+
+        let paths = fs::read_dir(&path).expect("Directory is empty");
```
sdwoodbury: 1) `read_dir` doesn't fail due to an empty directory. see error conditions here: `https://doc.rust-lang.org/std/fs/fn.read_dir.html`
2) `fs::create_dir_all` can create a directory, in which case the new directory would be empty.

extensions/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+
+#[derive(Default)]
+pub struct Librarian {}
+
+impl Librarian {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn locate(&mut self, path: PathBuf) -> &Self {
```
sdwoodbury: this should return a `Result` instead of panicking.

extensions/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+}
+
+impl fmt::Display for ExtensionProxy {
+    // This trait requires `fmt` with this exact signature.
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        // Write strictly the first element into the supplied output
+        // stream: `f`. Returns `fmt::Result` which indicates whether the
+        // operation succeeded or failed. Note that `write!` uses syntax which
+        // is very similar to `println!`.
+        write!(f, ".")
```
sdwoodbury, line 120: instead of printing ".", perhaps print `self.details()`?

ui/src/components/settings/sub_pages/extensions.rs
```diff
@@ -19,13 +20,22 @@ pub fn ExtensionSettings(cx: Scope) -> Element {
...
             },
-            ExtensionSetting {
-                title: placeholder,
-                author: "Nobody#1345".into(),
-                description: "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.".into(),
-                Switch {}
-            }
+            state.read().ui.extensions.values().map(|ext| {
+                let details = ext.extension.details();
+                println!("Deets {:?}", &details.meta.pretty_name);
```
sdwoodbury: this will eventually need to be replaces with a `log` macro

ui/src/extensions/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    /// # Safety
+    ///
+    /// An extension **must** be implemented using the
+    /// [`extensions::export_extension!()`] macro. Trying manually implement
+    /// a plugin without going through that macro will result in undefined
+    /// behaviour.
+    pub unsafe fn load<P: AsRef<OsStr>>(&mut self, library_path: P) -> io::Result<()> {
+        // load the library into memory
+        let library = Rc::new(Library::new(library_path).unwrap());
```
sdwoodbury, line 51: once `Library::new()` returns a `Result` (really an `io::Result`), `unwrap()` can be replaced with a `?`

ui/src/extensions/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+
+        let extension_info = library
+            .get::<unsafe extern "C" fn() -> *mut Core>(b"extension_entry")
+            .unwrap();
```
sdwoodbury, line 55: don't use `unwrap` in a function that returns a `Result`

ui/src/main.rs
```diff
@@ -85,8 +89,10 @@ pub static STATIC_ARGS: Lazy<StaticArgs> = Lazy::new(|| {
     };
     StaticArgs {
         uplink_path: uplink_path.clone(),
+        // TODO: This isn't quite accurate since uplink path is technically what we consider the cache dir.
```
sdwoodbury: sounds like `.uplink` should have a sub-directory for the cache? that would definitely help when overwriting an account - it would let the `themes` directory not get deleted.

ui/src/main.rs
```diff
@@ -399,6 +405,38 @@ pub fn app_bootstrap(cx: Scope) -> Element {
...
+    // load any extensions, we currently don't care to store the result of located extensions since they are stored by the librarian.
+    // We should however ensure we use the same librarian across the app so they should probably live in a globally accessible place
+    // that updates when they have new info, i.e. state.
+    fs::create_dir_all(&STATIC_ARGS.extensions_path).unwrap();
+    let paths = fs::read_dir(&STATIC_ARGS.extensions_path).expect("Directory is empty");
+    let mut extensions_library = AvailableExtensions::new();
+
+    for entry in paths {
+        let path = entry.unwrap().path();
+        println!("{:?}", path);
```
sdwoodbury: why use `println!` instead of `log`? executing `ui` with the `debug`, `trace`, or `trace2` flag causes the log macros to send output to the terminal. you can view available options like this: `ui --help`

sdwoodbury:
@InfamousVague 
I just tested the extension_example. this is pretty cool. 

I think this needs developer facing documentation. i see that the `Library` crate accepts a path to a library and then `Library::get` attempts to retrieve a function. and 
the `extension_entry` function returns a `Core` struct. But then there's the interaction between the `Librarian`, `ExtensionRegistrar`, `ExtensionProxy`, and `AvailableExtensions`. It's a lot to follow and some high level documentation would be helpful. But before we do that, it's probably worth iterating on the design. 

I'm guessing this was designed and built on the spur of the moment, as the team learned how to build and load a dynamic library. Now that we know how to make it work, we can probably make it better. I bet the design could be easily refined and simplified. This would make documentation and maintenance easier.

ui/src/extensions/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+            _lib: Rc::clone(&self.lib),
+        };
+        self.extensions.insert(name.to_string(), proxy);
+    }
+}
+
+#[derive(Default)]
+pub struct AvailableExtensions {
+    pub extensions: HashMap<String, ExtensionProxy>,
+    pub libraries: Vec<Rc<Library>>,
```
sdwoodbury: does this really need a `Vec` of `Rc<Library>`? `ExtensionProxy` contains a `Rc<Library>` too.

InfamousVague:
> @InfamousVague I just tested the extension_example. this is pretty cool.
> 
> I think this needs developer facing documentation. i see that the `Library` crate accepts a path to a library and then `Library::get` attempts to retrieve a function. and the `extension_entry` function returns a `Core` struct. But then there's the interaction between the `Librarian`, `ExtensionRegistrar`, `ExtensionProxy`, and `AvailableExtensions`. It's a lot to follow and some high level documentation would be helpful. But before we do that, it's probably worth iterating on the design.
> 
> I'm guessing this was designed and built on the spur of the moment, as the team learned how to build and load a dynamic library. Now that we know how to make it work, we can probably make it better. I bet the design could be easily refined and simplified. This would make documentation and maintenance easier.

We're still a ways off from it being where it needs to be; I'd suggest we focus on the core of things and worry about making extensions better after the core is rock solid.

ui/src/main.rs
InfamousVague: debugging, pr's still not ready.

ui/src/extensions/mod.rs
InfamousVague: PR in draft, not ready for review.

InfamousVague:
looks good to me, the goal I spoke with stu about in private is to get this merged in for now, since the extensions stuff is mostly abstracted out it _shouldn't_ cause any new issues, but of course we should have QA test still.

ui/src/components/chat/compose.rs
```diff
@@ -573,6 +571,18 @@ fn get_chatbar(cx: Scope<ComposeProps>) -> Element {
     let msg_valid =
         |msg: &[String]| !msg.is_empty() && msg.iter().any(|line| !line.trim().is_empty());
 
+    let extensions = &state.read().ui.extensions;
+
+    let ext_renders = {
```
sdwoodbury, line 576: either comment out rendering this for now, or filter based on the extensions location. that is, only render extensions intended for the compose window.

ui/src/main.rs
```diff
@@ -316,6 +292,34 @@ fn auth_wrapper(cx: Scope, page: UseState<AuthPages>, pin: UseRef<String>) -> El
     ))
 }
 
+fn get_extensions() -> HashMap<String, ExtensionProxy> {
+    // load any extensions, we currently don't care to store the result of located extensions since they are stored by the librarian.
+    // We should however ensure we use the same librarian across the app so they should probably live in a globally accessible place
+    // that updates when they have new info, i.e. state.
+    fs::create_dir_all(&STATIC_ARGS.extensions_path).unwrap();
+    let paths = fs::read_dir(&STATIC_ARGS.extensions_path).expect("Directory is empty");
```
sdwoodbury, line 300: the `expect()` call has an incorrect error message. `read_dir` doesn't fail when the directory is empty, but when the directory can't be read. 

ideally this wouldn't panic either.

phillsatellite:
The only issue that this caused onto the Dev branch is the theme dropdown no longer shows the list of themes to select from, this happens on Windows and MacOS but since this is important I can write up a ticket if we dont want it blocking the PR 

Screenshot 2023-02-21 at 4 52 02 PM

ui/src/main.rs
```diff
@@ -361,6 +368,57 @@ fn app(cx: Scope) -> Element {
     let desktop = use_window(cx);
     let state = use_shared_state::<State>(cx)?;
 
+    // Automatically select the best implementation for your platform.
+    let inner = state.inner();
```
sdwoodbury, line 372: i recommend moving this and the use_future to the end of this function. the function was laid out so the `use_future`s were declared after `main_element`, which allowed a dev to see the end result of this function without having to scroll past all the `use_future`s

PR #276, opened 2023-02-08
chore(Sounds): Started wiring notification sounds

InfamousVague:
Started wiring up the notification sounds but @sdwoodbury might have a better idea where the code should go (Ideally we can keep it out of components but we need access the app context to reference state.

### DIFF
--- a/ui/src/state/configuration.rs
+++ b/ui/src/state/configuration.rs
@@ -46,6 +46,11 @@ impl Configuration {
         let _ = self.config.save();
     }
 
+    pub fn set_message_sounds(&mut self, status: bool) {
+        self.config.audiovideo.message_sounds = status;
+        let _ = self.config.save();
+    }
+
     pub fn set_friends_notifications(&mut self, friends_notifications: bool) {
         self.config.notifications.friends_notifications = friends_notifications;
         let _ = self.config.save();
--- a/ui/src/state/mod.rs
+++ b/ui/src/state/mod.rs
@@ -19,6 +19,7 @@ pub use friends::Friends;
 pub use identity::Identity;
 pub use route::Route;
 pub use settings::Settings;
+use shared::language::get_local_text;
 pub use ui::{Theme, ToastNotification, UI};
 
 use crate::{
@@ -788,7 +789,7 @@ impl State {
 
                 if !self.ui.metadata.focused && notifications_enabled {
                     crate::utils::notifications::push_notification(
-                        "New friend request!".into(),
+                        get_local_text("friends.new-request"),
                         format!("{} sent a request.", identity.username()),
                         Some(crate::utils::sounds::Sounds::Notification),
                         notify_rust::Timeout::Milliseconds(4),
@@ -881,7 +882,7 @@ impl State {
                         None
                     };
                     crate::utils::notifications::push_notification(
-                        "New friend request!".into(),
+                        get_local_text("friends.new-request"),
                         format!("{} sent a request.", "NOT YET IMPL"),
                         sound,
                         notify_rust::Timeout::Milliseconds(4),
--- a/ui/src/state/mod.rs
+++ b/ui/src/state/mod.rs
@@ -833,6 +833,34 @@ impl State {
             } => {
                 // todo: don't load all the messages by default. if the user scrolled up, for example, this incoming message may not need to be fetched yet.
                 self.add_msg_to_chat(conversation_id, message);
+
+                // TODO: Get state available in this scope.
+                // Dispatch notifications only when we're not already focused on the application.
+                let notifications_enabled = self
+                    .configuration
+                    .config
+                    .notifications
+                    .messages_notifications;
+                let should_play_sound =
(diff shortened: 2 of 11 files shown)


ui/src/warp_runner/ui_adapter/message_event.rs
```diff
@@ -43,6 +43,37 @@ pub async fn convert_message_event(
...
+                notifications_enabled && !state.read().ui.metadata.focused;
+
+            state.write().mutate(state::Action::AddNotification(
+                state::notifications::NotificaitonKind::Notification,
+                1,
+            ));
+
+            if should_dispatch_notification {
+                crate::utils::notifications::push_notification(
+                    "New friend request!".into(),
```
sdwoodbury: needs translation

InfamousVague:
This is almost ready, there is just a remaining bug where the sound is played twice when not focusing the app, and for some reason it's played even when the new message is coming from the active chat

InfamousVague:
@phillsatellite this PR brings the ability to get notification sounds when you get a friend request and when messages come in.

ui/src/state/mod.rs
```diff
@@ -771,7 +771,29 @@ impl State {
...
+                // Dispatch notifications only when we're not already focused on the application.
+                let notifications_enabled = self
+                    .configuration
+                    .config
+                    .notifications
+                    .friends_notifications;
+
+                if !self.ui.metadata.focused && notifications_enabled {
+                    crate::utils::notifications::push_notification(
+                        "New friend request!".into(),
```
sdwoodbury: needs translations

ui/src/state/mod.rs
```diff
@@ -833,6 +855,41 @@ impl State {
...
+
+                // This should be called if we have notifications enabled for new messages
+                if should_dispatch_notification {
+                    let sound = if self.configuration.config.audiovideo.message_sounds {
+                        Some(crate::utils::sounds::Sounds::Notification)
+                    } else {
+                        None
+                    };
+                    crate::utils::notifications::push_notification(
+                        "New friend request!".into(),
```
sdwoodbury: same

sdwoodbury (changes requested):

phillsatellite:
@InfamousVague Yo! So I had a few issues,

This I replicated on both Windows and MacOS, it seems like _Message Sounds_ toggle is missing a click 
https://user-images.githubusercontent.com/93608357/217663577-b30cc504-a476-44ae-ada6-6165b9fdfbef.mov

The second issue @tooshel had on dev branch yesterday but I still havent been able to replicate on there. When I have a Windows User message a Mac User everytime the Mac User receives a message the text input box shoots to the top of the page

https://user-images.githubusercontent.com/93608357/217667045-b49334ae-f512-4066-a65f-1f98a6a2be6d.mov

InfamousVague:
> @InfamousVague Yo! So I had a few issues,
> 
> This I replicated on both Windows and MacOS, it seems like _Message Sounds_ toggle is missing a click https://user-images.githubusercontent.com/93608357/217663577-b30cc504-a476-44ae-ada6-6165b9fdfbef.mov
> 
> The second issue @tooshel had on dev branch yesterday but I still havent been able to replicate on there. When I have a Windows User message a Mac User everytime the Mac User receives a message the text input box shoots to the top of the page
> edit.mov

This could be related to line endings but I don't think it's related to this ticket since it just sends out the notification arbitrarily when we get the associated events. 

I will wire up the sound now though!

InfamousVague:
@phillsatellite should all be good to go now, I was able to hear the flip sound on my end but only in one direction but I did resolve a different bug around that area. If the thing doesn't make the sound twice for right now I wouldn't sweat it, we might just end up having to rethink some of the stuff in the sound util in the future if it becomes a problem.

ui/src/state/configuration.rs
```diff
@@ -46,6 +46,11 @@ impl Configuration {
         let _ = self.config.save();
     }
 
+    pub fn set_message_sounds(&mut self, status: bool) {
+        self.config.audiovideo.message_sounds = status;
+        let _ = self.config.save();
```
sdwoodbury, line 51: should only save if `status != self.config.audiovideo.message_sounds`

sdwoodbury (dismissed):
the requested change from last time wasn't addressed. only the English translation was added. Translations should be added for all languages in the `shared/locales` folder

InfamousVague:
> the requested change from last time wasn't addressed. only the English translation was added. Translations should be added for all languages in the `shared/locales` folder

I don't think we should require this, it's a little outta scope and I want to avoid crappy google translate translations by non native speaking developers. It should fallback just fine to the english version pending proper translation.

ui/src/state/configuration.rs
InfamousVague: Let's address this in a future PR as every configuration method currently doesn't use this logic and it's out of scope of this PR.

sdwoodbury (approved):
now it compiles

PR #279, opened 2023-02-08
fix(login): ensure tesseract is locked prior to login and add retry

sdwoodbury:
### What this PR does 📖

- what the title says. 
- the intent of this PR is to resolve the bug where Uplink stops recognizing your password. 

- Resolve #


ui/src/warp_runner/mod.rs
```diff
@@ -201,7 +201,28 @@ async fn login(
     new_account: bool,
 ) -> Result<manager::Warp, warp::error::Error> {
     log::debug!("login");
-    tesseract.unlock(passphrase.as_bytes())?;
+
+    if tesseract.is_unlock() {
+        log::warn!("login with tesseract alread unlocked. locking teseract");
+        tesseract.lock();
+        while tesseract.is_unlock() {
```
dariusc93: Was there an instance where calling `Tesseract::lock` still showed it was unlocked for some brief millis? Though nothing is wrong with this check since guessing that this is used to make sure that it is locked, but only curious.

sdwoodbury: nope, just being paranoid.

dariusc93 (dismissed):

phillsatellite:
I was able to replicate this each time only by adding a Mac user and a Windows user 

1. Create an account on MacOS 
2. Add Mac user as a friend on Windows 
3. Spam both accounts in chat by sending a lot of messages 
4. Once this happens sig will show as invalid

I also noticed when a Windows User messages a Mac user the Mac's text input will shoot to the top each time a message comes in 

And in last video you can see sig is invalid and I go to login and create account button never appears 

https://user-images.githubusercontent.com/93608357/218153617-a970967d-a642-4f58-8b4a-1b1ca08ab0b6.mov

https://user-images.githubusercontent.com/93608357/218153445-797f3c16-42cd-42c5-b6ef-8f34185ebbbb.mov

ui/src/warp_runner/mod.rs
```diff
@@ -109,110 +122,162 @@ async fn handle_login(notify: Arc<Notify>) {
...
-                            log::error!("create_identity failed. should never happen: {}", e);
+
+                        if let Err(e) = warp.tesseract.unlock(passphrase.as_bytes()) {
+                            log::info!("unlock failed: {:?}", e);
                             let _ = rsp.send(Err(e));
+                            continue;
+                        };
+                        match warp.multipass.create_identity(Some(&username), None).await {
+                            Ok(_id) =>  match wait_for_multipass(&mut warp, notify.clone()).await {
+                                Ok(_) => match save_tesseract(&warp.tesseract) {
```
dariusc93, line 152: Any reason to save when this is done internally once the autosave flag is set?

sdwoodbury: i was being paranoid

sdwoodbury:
@phillsatellite  I think we fixed the remaining problem. i think getting rid of the `lock()` call at the end of `warp_runner::manager::run` fixed the problem with the tesseract file being corrupted. Uplink now lets the user create a new account and overwrite the old one. 

the issue with friend requests not succeeding is a separate issue (as long as that doesn't cause the keystore to be corrupted)

ui/src/warp_runner/mod.rs
sdwoodbury: using a special save function lets me flush the file too, which seems to be needed now.

phillsatellite:
Finishing up testing one more thing on Darius's branch then switching over 👍

phillsatellite:
Retested again with all fresh accounts and still had the same issue as soon as you add a Windows user you it will break and when you try to log back in will skip to create account 

Macbook air account 
macair

Windows account
Widows

This video shows the Macbook pro account. I demonstrate opening and closing totally fine until once the Windows User is added 

This is the macbook Pro account, in video I demonstrate closing and opening app totally fine with another Mac User added and messages sent back and fourth, then once I add the Windows User I get the signature error before Windows even send Mac a message 

https://user-images.githubusercontent.com/93608357/218182754-86a6823a-e4d4-4e4d-9ddd-7c113e3e7ecd.mov

ui/src/warp_runner/mod.rs
```diff
@@ -222,10 +306,20 @@ async fn warp_initialization(
     experimental: bool,
 ) -> Result<manager::Warp, warp::error::Error> {
     log::debug!("warp initialization");
+    let repo_lock_path = format!("{}/repo_lock", &STATIC_ARGS.warp_path.to_string_lossy());
```
dariusc93: I wouldnt advise removing the lock. The point behind it is to make 2 instances cannot share the same directory at the same time. Although i do plan on hooking it something different than this file, removing it would likely cause other instances thats being initialized to share the directory and cause conflicts and corruption

ui/src/warp_runner/mod.rs
```diff
@@ -256,3 +350,30 @@ async fn warp_initialization(
...
+            log::error!("failed to open tesseract keystore for saving: {}", e);
+            return Err(Error::CorruptedDataStore);
+        }
+    };
+    if let Err(e) = tesseract.to_writer(&mut file) {
+        log::error!("tesseract.to_writer() failed: {}", e);
+        return Err(e);
+    }
+
+    if let Err(e) = file.sync_all() {
```
dariusc93, line 363: This will be done in `Tesseract::to_file` in https://github.com/Satellite-im/Warp/pull/166

ui/src/warp_runner/mod.rs
```diff
@@ -110,110 +123,185 @@ async fn handle_login(notify: Arc<Notify>) {
...
+        if let Err(e) = std::fs::remove_dir_all(&STATIC_ARGS.uplink_path) {
+            log::warn!("failed to delete uplink directory: {}", e);
+        }
+
+        // create directories
+        if let Err(e) = std::fs::create_dir_all(&STATIC_ARGS.warp_path) {
+            log::warn!("failed to create warp directory: {}", e);
+        }
+
+        if let Err(e) = std::fs::create_dir_all(&STATIC_ARGS.themes_path) {
```
dariusc93:
```suggestion
        if let Err(e) = tokio::fs::create_dir_all(&STATIC_ARGS.themes_path).await {
```

phillsatellite (approved):

phillsatellite:
Logged in and out almost a thousand times using the X and Control+C with MacOS + Windows users added and didnt get forced to create a new account a single time 🔥  2 dev reviews and this is good to go on my side!

ui/src/warp_runner/mod.rs
```diff
@@ -256,3 +340,30 @@ async fn warp_initialization(
         constellation: storage,
     })
 }
+
+pub fn save_tesseract(tesseract: &warp::tesseract::Tesseract) -> Result<(), Error> {
```
dariusc93, line 344: Here, this could be replaced internally with 

```rust
tesseract.to_file()?;
```

ui/src/warp_runner/mod.rs
```diff
@@ -110,110 +123,181 @@ async fn handle_login(notify: Arc<Notify>) {
...
+        tesseract.set_autosave();
+        tesseract
+    };
+
+    // this code path addresses cross-platform issues involving account recreation.
+    // the tesseract file was being overwritten incorrectly.
+    // to fix this, manually delete the file and re-create it.
+    if overwrite_old_account {
+        // delete old account data
+        if let Err(e) = std::fs::remove_dir_all(&STATIC_ARGS.uplink_path) {
```
dariusc93, line 259: Not required since these are short lived, but may want to use `tokio::fs` for these since its inside of an async function

ui/src/warp_runner/mod.rs
```diff
@@ -110,110 +123,181 @@ async fn handle_login(notify: Arc<Notify>) {
...
+            log::warn!("failed to delete uplink directory: {}", e);
+        }
+
+        // create directories
+        if let Err(e) = std::fs::create_dir_all(&STATIC_ARGS.warp_path) {
```
dariusc93, line 264: ditto

ui/src/warp_runner/mod.rs
```diff
@@ -110,110 +123,181 @@ async fn handle_login(notify: Arc<Notify>) {
...
+            log::warn!("failed to create warp directory: {}", e);
+        }
+
+        // create the tesseract key file so it can be saved later
+        if let Err(e) = std::fs::File::create(&STATIC_ARGS.tesseract_path) {
```
dariusc93, line 269: ditto

ui/src/warp_runner/mod.rs
```diff
@@ -110,110 +123,181 @@ async fn handle_login(notify: Arc<Notify>) {
...
-        log::info!("string keypair not found in tesseract");
-        return Err(warp::error::Error::IdentityNotCreated);
+        return Ok(configure_tesseract(Tesseract::default()));
     }
-    let res = warp_initialization(tesseract, false).await;
-    tokio::time::sleep(std::time::Duration::from_millis(100)).await;
-    res
+
+    // open existing file or create new one
+    let tesseract = match std::fs::File::open(&STATIC_ARGS.tesseract_path) {
```
dariusc93, line 278: This could be replaced with `Tesseract::open_or_create`, which would open the file if it exist, otherwise it would attempt to create one if it doesnt. Do note, it would only attempts this if the file doesnt exist. If it does exist, it may throw other errors ( (eg doesnt have permissions, OOM, etc), in which case those errors should be brought up stream

ui/src/warp_runner/mod.rs
lgmarchi, line 264: Warp directory has not been created automatically when we are creating an account? I mean, this code is really necessary?

ui/src/warp_runner/mod.rs
```diff
@@ -110,110 +123,181 @@ async fn handle_login(notify: Arc<Notify>) {
...
+        tesseract
+    };
+
+    // this code path addresses cross-platform issues involving account recreation.
+    // the tesseract file was being overwritten incorrectly.
+    // to fix this, manually delete the file and re-create it.
+    if overwrite_old_account {
+        // delete old account data
+        if let Err(e) = std::fs::remove_dir_all(&STATIC_ARGS.uplink_path) {
+            log::warn!("failed to delete uplink directory: {}", e);
```
lgmarchi, line 260: I think these logs should be an error, than warn, what do you think?

ui/src/warp_runner/mod.rs
```diff
@@ -110,110 +123,181 @@ async fn handle_login(notify: Arc<Notify>) {
...
-    while !tesseract.is_unlock() {
-        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
-    }
+// don't set file or autosave until tesseract is unlocked
+// assumes that all anyone needs from tesseract is "keypair"
+// otherwise, Tesseract::to_file probably needs to call file.sync_all()
+async fn init_tesseract(overwrite_old_account: bool) -> Result<Tesseract, Error> {
+    log::trace!("initializing tesseract");
+
+    let configure_tesseract = |tesseract: Tesseract| {
```
lgmarchi, line 247: After all changes, deleting repositories, this code should be necessary yet?

ui/src/main.rs
```diff
@@ -155,6 +159,8 @@ enum LogProfile {
     Debug,
     /// print everything including tracing logs to the terminal
     Trace,
+    /// like trace but include warp logs
+    Trace2,
```
lgmarchi, line 163: Would be a better name, like TracePlus or TraceWithWarp? Just for the name be clearer.

ui/src/logger/mod.rs
```diff
@@ -220,7 +205,7 @@ pub fn init_with_level(level: LevelFilter) -> Result<(), SetLoggerError> {
     log::set_max_level(level);
     log::set_boxed_logger(Box::new(LogGlue::new(level)))?;
     if level == LevelFilter::Trace {
```
lgmarchi, line 207: Here, should be LevelFilter::Trace2, right?

lgmarchi (approved):
This PR looks awesome! Let some comments, but I think none of them is a blocking for this important PR.

ui/src/warp_runner/mod.rs
dariusc93: Yea internally it does check for the directories and create them if it doesnt exist.

ui/src/warp_runner/mod.rs
sdwoodbury: i'm fine with either. i thought to use warn because the program continued after the error, instead of returning.

ui/src/warp_runner/mod.rs
sdwoodbury: i believe we definitely want to delete the old warp directory and `state.json` when creating a new account. but if we use tesseract::from_file then we might not need the `confiure_tesseract` closure anymore

ui/src/main.rs
sdwoodbury: idk, it's all documented via the `--help` option and without running that command, no one would know about the different trace options anyway. i had the same thoughts and considered something like "supertrace". but felt that it would be more work and less convenient, in exchange for minimal gain

ui/src/logger/mod.rs
sdwoodbury: no, the command line option enables trace logs from another module, the `LevelFilter` is still `Trace`, it doesn't get any higher then that.

ui/src/warp_runner/mod.rs
sdwoodbury: sounds like this depends on whether Tesseract is created `from_file` or `from_reader`

sdwoodbury:
i'll address the feedback in another branch. if I pushed any changes, all the testing and reviews would need to be redone, per the current github configuration.

PR #280, opened 2023-02-08
chore(login): add superfluous login button if user is returning, also…

### LINKED ISSUE #260: Login - Update Login button
**Task:** I was talking with Matt and he was thinking we need to update the login button. When you have a fresh account the Create Account button will be there when you are creating a PIN. When you are trying to log back in again there will be no button there and after you type in your PIN you are logged right in. 

What Matt suggested we do instead is:

- If a User opens the Uplink app and there is no account sensed, then the Create Account button should be visible (this is all good on dev right now 😄 )

- If a User opens Uplink and there is already an account sensed, then the button should be visible still but instead of saying _Create Account_ it should say something like "Log in"

**Screenshots:**

https://user-images.githubusercontent.com/93608357/217070598-e4b26178-190d-4d64-88da-34c1fcf398b8.mov

Screenshot 2023-02-06 at 2 46 11 PM

WanderingHogan:
… add different message to unlock screen

### What this PR does 📖
Adds in the button that isn't actually tied to anything. On input we check to see if the pin is valid. Needs input to be fixed to show error message on unlock (already a ticket for it)
Screenshot 2023-02-08 at 1 32 10 PM
Screenshot 2023-02-08 at 1 31 51 PM

- 

- Resolve #260

### DIFF
--- a/shared/locales/en-US/main.ftl
+++ b/shared/locales/en-US/main.ftl
@@ -167,8 +167,10 @@ remote-controls = Remote Controls
 unlock = Unlock
     .warning1 = Your password is used to encrypt your data. It is never sent to any server. You should use a strong password that you don't use anywhere else.
     .warning2 = If you forget this password we cannot help you retrieve it.
+    .welcome-back = Welcome back! Please enter your pin to unlock your account.
     .enter-pin = Enter Pin
     .create-account = Create Account
+    .unlock-account = Unlock Account
 
 auth = Create Account 
     .enter-username = Enter Username
--- a/shared/locales/es-MX/main.ftl
+++ b/shared/locales/es-MX/main.ftl
@@ -151,8 +151,10 @@ remote-controls = Remote Controls
 unlock = Unlock
     .warning1 = Su contraseña se utiliza para cifrar sus datos. Nunca se envía a ningún servidor. Debe usar una contraseña fuerte que no use en ningún otro lugar.
     .warning2 = Si olvida esta contraseña, no podremos ayudar a recuperarla.
+    .welcome-back = ¡Bienvenido de nuevo! Ingrese su pin para desbloquear su cuenta.
     .enter-pin = Ingresar Pin
     .create-account = Crear Cuenta
+    .unlock-account = Desbloquear Cuenta
 
 auth = Create Account 
     .enter-username = Ingresar Nombre de Usuario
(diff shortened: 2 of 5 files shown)


InfamousVague:
We will want to actually wire this up; 

Right now the way it functions is as intended but it's missing functionality. 

The flow should be 

A user first visits the pin page, it will give them info since it's their first time about what the pin is and how we can't recover it. They pick a pin and as long as the pin is longer than the required minimum we allow them to hit enter, or the button, to continue to the next page. (Hitting enter on mobile devices may not be an option)

Next we go to login again and we don't really need to tell them everything about a pin again we just say "enter your pin to unlock" and give them an input and a disabled button. once the pin is over the required minimum as a convenience we check if the entered pin is correct on every keypress after the minimum so we can save them having to hit the button. But in the case that they enter there pin wrong and can't press enter, there is currently no way to indicate they entered the wrong pin since they can't tap the button. 

We should just make it so that after the minimum required length the button shows up and when they tap it, it should try our whatever pin they have entered and return any errors like pressing enter does, I guess just a proxy to hitting enter.

InfamousVague:
I think looking at the code, it should just be doing whatever code happens on enter inside the
 ```
 onpress: move |_| {
// do nothing, it will unlock if the pin is correct, and should show the error message if it is not
}
```

LGTM though after that

WanderingHogan:
So - right now, every single input the validation is ran.

Do you want to remove this and make it only fire on enter/button press?

There is another ticket to fix the wrong pin/no account/other tesseract errors and make them show up on the input. By the time the button is visible we actually already know if the pin is bad or its some other issue, do you want us to check again even without an input change?

InfamousVague:
> So - right now, every single input the validation is ran.
> 
> Do you want to remove this and make it only fire on enter/button press?
> 
> There is another ticket to fix the wrong pin/no account/other tesseract errors and make them show up on the input. By the time the button is visible we actually already know if the pin is bad or its some other issue, do you want us to check again even without an input change?

I think it'd be best if we made it try to login / decrypt automatically every keypress after the minimum key length. However we should only validate the pin on enter or hitting the button. So it will not say invalid pin or incorrect pin until we hit return or hit the button, but if we enter the correct pin we'll still log in automatically.

WanderingHogan:
@InfamousVague ok cool i think i got it, thank you

PR #282, opened 2023-02-09
chore(themes): fix themes dropdown for windows and sorts items alphabetically 

### LINKED ISSUE #261: Theme dropdown will have double the amount of Themes
**Issue:** Seems to be on Windows only. I had this issue a single time with MacOS but then was fixed in a matter of minutes and couldnt replicate again after that. 

- first issue is the list of themes are duplicated when you open the dropdown 

-The second issue is the names of the themes are showing the path instead of just the name of the theme. Example (Themes \ default )

- Default will also appear twice in the dropdown once as "Default" the second as "Themes \ Default" 

-And I'm not sure if this part really matters or not but I noticed I have 2 folders labeled Theme one of them in 

`Uplink/ui/extra/themes` 

and also 

`.uplink/themes`

**So during the call Sheldon and I had he noticed in `mod.rs ui/src/utils` the code, line 18 `-join(path: ".uplink/"` was searching for .uplink and inside the folder we had a Themes folder and then a list of the themes as well so thats why in Settings the User would still have a duplicated list of themes

**Screenshot:**

image

Themes & Themes \ Default 

https://user-images.githubusercontent.com/93608357/217906000-53bb526d-7501-448d-8e7a-d03235b575d9.mp4

image

tooshel:
### What this PR does 📖

- fixes the theme dropdown for windows and sorts items alphabetically 
1) the path was pulling from .uplink, the default, instead of using the static arg
2) the name of the theme was being created by splitting the path on / which doesn't work right on windows

- Resolve #261

Please make sure the path fix and theme dropdown works correctly on Mac!

### DIFF
--- a/ui/src/utils/mod.rs
+++ b/ui/src/utils/mod.rs
@@ -23,18 +23,15 @@ pub fn get_available_themes() -> Vec<Theme> {
         .filter_map(|file| file.ok())
     {
         if file.metadata().unwrap().is_file() {
-            let theme = file.path().display().to_string();
-
-            let mut theme_str = theme.replace("\\", "/");
-            theme_str = theme_str.split('/').last().unwrap().to_string();
-
-            let pretty_theme_str = &theme_str.replace(".scss", "");
+            let theme_path = file.path().display().to_string();
+            let theme_name_str = file.path().iter().last().unwrap();
+            let pretty_theme_str = &theme_name_str.to_string_lossy().replace(".scss", "");
             let pretty_theme_str = titlecase(pretty_theme_str);
 
-            let styles = fs::read_to_string(&theme).unwrap_or_default();
+            let styles = fs::read_to_string(&theme_path).unwrap_or_default();
 
             let theme = Theme {
-                filename: theme_str.to_owned(),
+                filename: theme_path.to_owned(),
                 name: pretty_theme_str.to_owned(),
                 styles,
             };
--- a/ui/src/utils/mod.rs
+++ b/ui/src/utils/mod.rs
@@ -3,7 +3,10 @@ use std::fs;
 use titlecase::titlecase;
 use walkdir::WalkDir;
 
-use crate::state::{self, Theme};
+use crate::{
+    state::{self, Theme},
+    STATIC_ARGS,
+};
 use kit::User as UserInfo;
 
 pub mod format_timestamp;
@@ -13,10 +16,7 @@ pub mod sounds;
 pub fn get_available_themes() -> Vec<Theme> {
     let mut themes = vec![];
 
-    let theme_path = dirs::home_dir()
-        .unwrap_or_default()
-        .join(".uplink/")
-        .join("themes");
+    let theme_path = STATIC_ARGS.themes_path.clone();
 
     for file in WalkDir::new(theme_path)
         .into_iter()
@@ -25,7 +25,9 @@ pub fn get_available_themes() -> Vec<Theme> {
         if file.metadata().unwrap().is_file() {
             let theme = file.path().display().to_string();
 
-            let theme_str = theme.split('/').last().unwrap();
+            let mut theme_str = theme.replace("\\", "/");
+            theme_str = theme_str.split('/').last().unwrap().to_string();
+
             let pretty_theme_str = &theme_str.replace(".scss", "");
             let pretty_theme_str = titlecase(pretty_theme_str);
 
@@ -40,6 +42,7 @@ pub fn get_available_themes() -> Vec<Theme> {
             themes.push(theme);
         }
     }
+    themes.sort_by_key(|theme| theme.name.clone());
 
     themes
 }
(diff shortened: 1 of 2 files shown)


phillsatellite (dismissed):

phillsatellite:
Awesome fix!

ui/src/utils/mod.rs
```diff
@@ -25,7 +25,9 @@ pub fn get_available_themes() -> Vec<Theme> {
         if file.metadata().unwrap().is_file() {
             let theme = file.path().display().to_string();
 
-            let theme_str = theme.split('/').last().unwrap();
+            let mut theme_str = theme.replace("\\", "/");
+            theme_str = theme_str.split('/').last().unwrap().to_string();
+
```
InfamousVague: Since this is a windows only issue, let's wrap this in the os tag. 

> #[cfg(target_os = "windows")]
> theme_str = theme_str...

so that we leave it relatively untouched on other systems to prevent causing bugs on other systems. (while unlikely, just nice to be sure)

InfamousVague (changes requested):
LGTM after the requested change

ui/src/utils/mod.rs
InfamousVague: Also I'd toss in a little comment above the code to explain why we do that on windows so people like me (slightly oblivious) don't look over it in the future and break it in some other way.

ui/src/utils/mod.rs
```diff
@@ -25,7 +25,9 @@ pub fn get_available_themes() -> Vec<Theme> {
         if file.metadata().unwrap().is_file() {
             let theme = file.path().display().to_string();
 
-            let theme_str = theme.split('/').last().unwrap();
+            let mut theme_str = theme.replace("\\", "/");
+            theme_str = theme_str.split('/').last().unwrap().to_string();
```
dariusc93: Instead, you should be able to use `file.path().file_name()`  instead of splitting.  You can also iterate over the path and get the last item if you prefer that

eg

```rust
let theme_str = file.path().file_name().map(|s| s.to_string_lossy().to_string()).unwrap();
```

or 

```rust
let theme_str = file.path().iter().last().unwrap();
```

Note you do want to avoid unwrapping unless youre completely sure it wont panic.

dariusc93 (commented):
Left a suggestion but besides that it looks fine to me

tooshel:
> Left a suggestion but besides that it looks fine to me

I'm on it! I will try both and see which I like better.

ui/src/utils/mod.rs
tooshel: I like that! I do wish there were just "path" tools like there are for urls and other languages that abstract that stuff away. I will look around a bit. And maybe Darius's suggestion will help.

ui/src/utils/mod.rs
tooshel: I liked the second one because it was clearer what it was returning . . . but I did have to make a few other changes since it was returning an osstr instead of just a str.

ui/src/utils/mod.rs
tooshel: @InfamousVague I ended up taking those two lines out and replacing them with 1 that uses those path tools that I was hopping existed. The did end up existing but instead of str you get osstr and it was a bit of madness so I changed a few other things but we are still new one less line of code so I say it's a win.

sdwoodbury:
hey @tooshel i committed a function that might be better for calculating the theme name. try it out and if you don't like it, feel free to delete the function. right now the function is unused except for in a unit test.

tooshel:
> hey @tooshel i committed a function that might be better for calculating the theme name. try it out and if you don't like it, feel free to delete the function. right now the function is unused except for in a unit test.

Nice! I'll check it out!

ui/src/utils/mod.rs
```diff
@@ -13,37 +16,43 @@ pub mod sounds;
...
         .into_iter()
         .filter_map(|file| file.ok())
     {
         if file.metadata().unwrap().is_file() {
-            let theme = file.path().display().to_string();
-
-            let theme_str = theme.split('/').last().unwrap();
-            let pretty_theme_str = &theme_str.replace(".scss", "");
+            let theme_path = file.path().display().to_string();
+            let theme_name_str = file.path().iter().last().unwrap();
```
lgmarchi, line 27: To avoid panic, would be better to use 
`
unwrap_or("default")`

@dariusc93 @sdwoodbury ?

lgmarchi (approved):
LGTM. Approved!

phillsatellite (approved):

phillsatellite:
Perfect on my side awesome job @tooshel

ui/src/utils/mod.rs
tooshel: @lgmarchi . . . thanks for bringing that up! Turns out Stuart added the function but isn't using it. I can use it and replace that line.

InfamousVague (approved):

PR #298, opened 2023-02-14
feat(Thumbnail): Add thumbnails for images and video files

lgmarchi:
### 1. Now, if file is an image or video, it will show its thumbnail 

https://user-images.githubusercontent.com/63157656/218764985-cf959f49-2aab-4aa7-ad5d-2ff1cc70760d.mov

### 2. To extract thumbnail from a video, I used ffmpeg with commands 
a. For it works, we need to install ffmpeg -> `brew install ffmpeg` for macOS 
And for windows, I followed steps on this site [here](https://www.geeksforgeeks.org/how-to-install-ffmpeg-on-windows/)
b. Commands are: 
image

- **.arg("-i")**: Specifies the input file for FFmpeg.

- **.arg(file_path)**: Specifies the path to the input file that you want to convert.

- **.arg("-vf")**: Specifies the video filter to apply.

- **.arg("select=eq(pict_type\\,I)")**: Filters the input video stream to only include key frames (I-frames).

- **.arg("-q:v")**: Specifies the video quality.

- **.arg("2")**: Sets the video quality to 2. The lower the value, the higher the quality.

- **.arg("-f")**: Specifies the output format.

- **.arg("image2")**: Sets the output format to image2, which is a format for image sequences.

- **.arg("-update")**: Sets FFmpeg to update the output file, even if it already exists.

- **.arg("1")**: Specifies that only one image should be written to the output file.

- **.arg(path.clone())**: Specifies the output file path.

- **.stdout(Stdio::piped())**: Redirects the standard output to a pipe.

- **.stderr(Stdio::null())**: Redirects the standard error to /dev/null to suppress any error messages that FFmpeg may output.

- Resolve #

### DIFF
--- a/ui/src/warp_runner/manager/commands/constellation_commands.rs
+++ b/ui/src/warp_runner/manager/commands/constellation_commands.rs
@@ -278,13 +278,13 @@ async fn upload_files(
                     }
                 }
                 match set_thumbnail_if_file_is_image(warp_storage, filename.clone()).await {
-                    Ok(success) => log::info!("{:?}", success),
+                    Ok(_) => log::info!("Image Thumbnail uploaded"),
                     Err(error) => log::error!("Error on update thumbnail for image: {:?}", error),
                 }
                 match set_thumbnail_if_file_is_video(warp_storage, filename.clone(), file_path)
                     .await
                 {
-                    Ok(success) => log::info!("Video Thumbnail: {:?}", success),
+                    Ok(_) => log::info!("Video Thumbnail uploaded"),
                     Err(error) => log::error!("Error on update thumbnail for video: {:?}", error),
                 }
                 log::info!("{:?} file uploaded!", filename);
@@ -335,6 +335,7 @@ async fn set_thumbnail_if_file_is_video(
     file_path: PathBuf,
 ) -> Result<(), Box<dyn std::error::Error>> {
     let video_formats = VIDEO_FILE_EXTENSIONS.to_vec();
+
     let file_extension = std::path::Path::new(&filename_to_save)
         .extension()
         .and_then(OsStr::to_str)
@@ -358,12 +359,7 @@ async fn set_thumbnail_if_file_is_video(
 
     let temp_dir = TempDir::new()?;
 
-    let path = format!(
-        "{}/{}{}.jpg",
-        temp_dir.path().to_string_lossy(),
-        file_stem,
-        uuid::Uuid::new_v4()
-    );
+    let temp_path = temp_dir.path().join(file_stem);
 
     let output = Command::new("ffmpeg")
         .arg("-i")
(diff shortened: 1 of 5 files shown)


dariusc93:
The problem is that not all systems have ffmpeg installed so calling the command to execute it would yield an error if it doesnt exist in the system. For images, you could use the `image` crate.

ui/src/warp_runner/manager/commands/constellation_commands.rs
```diff
@@ -319,10 +329,76 @@ fn rename_if_duplicate(
...
+        .file_stem()
+        .and_then(OsStr::to_str)
+        .map(str::to_string)
+        .unwrap_or_default();
+
+    let temp_dir = TempDir::new()?;
+
+    let temp_path = temp_dir.path().join(file_stem);
+
+    let output = Command::new("ffmpeg")
```
dariusc93, line 362: Since this is done in a async function, you would want to use the `Command` from tokio to prevent blocking the current thread.

lgmarchi:
> The problem is that not all systems have ffmpeg installed so calling the command to execute it would yield an error if it doesnt exist in the system. For images, you could use the `image` crate.

But if do we compile the app, as an executable, it will not work as well?

ui/src/warp_runner/manager/commands/constellation_commands.rs
lgmarchi: Actually, I removed async of this function, if I change command for tokio, I start to have other problems

phillsatellite:
Small issue on Windows and MacOS: 

I couldn't seem to get any video thumbnails to appear and also had trouble with any .heic photo 

windows

Screenshot 2023-02-14 at 2 29 20 PM

lgmarchi:
> Small issue on Windows and MacOS:
> 
> I couldn't seem to get any video thumbnails to appear and also had trouble with any .heic photo
> 
> windows
> 
> Screenshot 2023-02-14 at 2 29 20 PM

Heic format for pictures, could be in other, is not common format to work, and crate for it can provoke performance issues. 

About video thumbnails, is necessary to install ffmpeg in both machines, MacOS or Windows.

And I added a code now, if ffmpeg fails to get video thumbnails, will show document with play symbol, to let user know that it is a video file. 

image

phillsatellite:
got ffmpeg working on my Windows, thumbnail appearing on Mac and Windows on my side 🔥

phillsatellite (dismissed):

ui/src/warp_runner/manager/commands/constellation_commands.rs
```diff
@@ -319,10 +327,76 @@ fn rename_if_duplicate(
...
+        .arg("-i")
+        .arg(file_path)
+        .arg("-vf")
+        .arg("select=eq(pict_type\\,I)")
+        .arg("-q:v")
+        .arg("2")
+        .arg("-f")
+        .arg("image2")
+        .arg("-update")
+        .arg("1")
```
InfamousVague: not your fault, but is this really the only way they allow us to have multiple args? It's very verbose

InfamousVague (dismissed):
LGTM

ui/src/warp_runner/manager/commands/constellation_commands.rs
lgmarchi: image

There is this way, using like a vec of args. I will change!

lgmarchi:
Re tested PR after changes made. 

https://user-images.githubusercontent.com/63157656/219172069-47a3e0ee-987d-4168-be55-8bf365f9555f.mov

phillsatellite (approved):

phillsatellite:
I retested again and was good on my side as well 🔥

InfamousVague (approved):

PR #311, opened 2023-02-17
fix(config): fix default values and saving 

sdwoodbury:
### What this PR does 📖
- addresses the following problems:
1. changing settings within state didn't trigger a re-render or cause state to be saved
2. State was expected to have a default configuration that matched what Serde would generate from this: "{}" 
3. when the configuration was saved, `Config.json` didn't match the `configuration` entry in `State.json` 

- moves `config.rs` into `state/configuration.rs` and forces configuration to be changed via `state::mutate()`, same as everything else in state. 
- shadows changes to `AudioVideo` config so that the unlock and create_account pages can play the sound (to maintain previous behavior). 

- fixes some misc clippy warnings and a minor bug in logger

### Tests: 
a. A way to test it, is creating new account, going to .uplink folder, and verify if state.json file was created with correct default values, some of them true, some of them false 
b. Other way, is testing if sound in app is working if you receive new message, without need to go to settings page and verify if the option is true 
c. Other test would be see state.json file, go to settings on Uplink, change one of the options there, and observe if values has been changed in state.json 

- Resolve #

### DIFF
--- a/ui/src/state/action.rs
+++ b/ui/src/state/action.rs
@@ -165,26 +165,26 @@ pub enum Action {
 
 #[derive(Display)]
 pub enum ConfigAction {
-    #[display(fmt = "NotificationsEnabled {_0}")]
-    NotificationsEnabled(bool),
-    #[display(fmt = "Theme {_0}")]
-    Theme(String),
-    #[display(fmt = "OverlayEnabled {_0}")]
-    OverlayEnabled(bool),
-    #[display(fmt = "DevModeEnabled {_0}")]
-    DevModeEnabled(bool),
-    #[display(fmt = "InterfaceSoundsEnabled {_0}")]
-    InterfaceSoundsEnabled(bool),
-    #[display(fmt = "MediaSoundsEnabled {_0}")]
-    MediaSoundsEnabled(bool),
-    #[display(fmt = "MessageSoundsEnabled {_0}")]
-    MessageSoundsEnabled(bool),
-    #[display(fmt = "FriendsNotificationsEnabled {_0}")]
-    FriendsNotificationsEnabled(bool),
-    #[display(fmt = "MessagesNotificationsEnabled {_0}")]
-    MessagesNotificationsEnabled(bool),
-    #[display(fmt = "SettingsNotificationsEnabled {_0}")]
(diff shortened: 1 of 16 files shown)


phillsatellite (dismissed):

phillsatellite:
Everything seemed good on my end! 🔥

ui/src/state/configuration.rs
```diff
@@ -1,68 +1,205 @@
...
+        // Create a default configuration here
+        // For example:
+        Self::default()
     }
+}
+
+impl Configuration {
+    pub fn handle_action(&mut self, action: ConfigAction) {
+        let old_audiovideo = self.audiovideo;
+        match action {
```
lgmarchi, line 148: I leave it as a suggestion to change the actions here, because it does not look like a writing that demonstrates an action being performed

ui/src/state/action.rs
```diff
@@ -159,6 +159,32 @@ pub enum Action {
     MockSend(Uuid, Vec<String>),
     #[display(fmt = "ClearUnreads")]
     ClearUnreads(Chat),
+    #[display(fmt = "Config {_0}")]
+    Config(ConfigAction),
+}
+
+#[derive(Display)]
+pub enum ConfigAction {
```
lgmarchi, line 167: Same about name of the actions

lgmarchi (commented):
Left some comments about names used on actions. I think is good to change them.

ui/src/state/configuration.rs
sdwoodbury: What do you recommend the code be renamed to?

lgmarchi: Like, SetNotificationsUsage or SetNotificationsAvailability. Because it is an action, so I think is better for legibility to use a verb.

ui/src/state/action.rs
sdwoodbury: image

i think calling this a `ConfigAction` is close enough considering that it's inside of an `Action` enum. i can rename the function from `handle_action` to `mutate` to make it mimic what `Action` does. i can prefix the enumeration values with `Set` to make it sound more action-y too.

InfamousVague (approved):

PR #312, opened 2023-02-17
feat(Constellation): Rename files and folders

lgmarchi:
### 1. You can rename a file or a folder
a. It will keep extension on name 
b. Can not have file or folder, with same name inside same folder
c. Using Esc Key, you can close input field 
d. `/` and `\\` are not allowed as well on rename files 

https://user-images.githubusercontent.com/63157656/219701906-ce5df258-d275-4e8b-874d-e76796b5c301.mov

- Resolve #


phillsatellite:
Worked good! Thumbnails worked, could rename files but couldnt rename files with 2  of the same names and rename persisted after closing out and re opening the app ✅

kit/src/elements/file/mod.rs
```diff
@@ -87,9 +87,19 @@ pub fn emit_press(cx: &Scope<Props>) {
     }
 }
 
+pub fn get_file_extension(file_name: String) -> String {
```
sdwoodbury, line 87: i see this is only used for `get_text`. if you want this to be a function so you can unit test it, then that's cool. but I think it's a mistake to make `get_text` take a second argument for `file_extension`. if `get_text` needs the file extension, it can call `get_file_extension` on its own. I think this should be changed back before merging.

kit/src/elements/file/mod.rs
```diff
@@ -135,14 +145,43 @@ pub fn File<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
...
-                with_rename.then(|| rsx! (
-                    Input {
-                        disabled: disabled,
-                        placeholder: placeholder,
-                        // todo: use is_valid
-                        onreturn: move |(s, _is_valid)| emit(&cx, s)
+                with_rename.then(||
+                    {
+                    let chars_to_remove = vec!['\\', '/'];
+                    let mut special_chars = SPECIAL_CHARS.to_vec();
```
sdwoodbury: i know this filtering of `SPECIAL_CHARS` is needed for the input validation. I think it's better to  add a `chars_to_allow` field on the input box. in this case it would be `vec!['\\', '/']`. the use of `SPECIAL_CHARS` could be moved inside of the `Input` element. 

without doing this, the logic for filtering allowed characters and then passing them to `Input` will be duplicated a lot. i think the sooner this gets changed, the better.

sdwoodbury:
review in progress

kit/src/elements/folder/mod.rs
```diff
@@ -114,9 +116,9 @@ pub fn Folder<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
                                     ..Options::default()
                                 }
                                 // todo: use is_valid
-                                onreturn: move |(s, _is_valid)| {
+                                onreturn: move |(s, _is_valid, key_code)| {
```
sdwoodbury: the comment can be deleted. `_is_valid` doesn't need the underscore anymore

ui/src/layouts/storage.rs
```diff
@@ -342,42 +375,91 @@ pub fn FilesLayout(cx: Scope<Props>) -> Element {
...
                         let folder_name = dir.name();
-                        rsx!(Folder {
-                            text: dir.name(),
-                            aria_label: dir.name(),
-                            onpress: move |_| {
-                                ch.send(ChanCmd::OpenDirectory(folder_name.clone()));
-                            }
-                        })
+                        let folder_name2 = dir.name();
+                        let is_renaming = use_state(cx, || false);
```
sdwoodbury: calling `use_state` or any other hook inside of a loop, iterator, or conditional statement is bad. don't ever do this. if the files list changes,  then one invocation of `FilesLayout` will produce a different set of hooks then the next. 

This violates the [rules of hooks](https://dioxuslabs.com/docs/0.3/guide/en/interactivity/hooks.html?highlight=hooks#rules-of-hooks). 

instead of using a `ContextMenu` you need to make a new element which contains a `ContextItem` and a `File`, create the hook in the top level element, and pass it to both child elements.

ui/src/layouts/storage.rs
```diff
@@ -342,42 +375,91 @@ pub fn FilesLayout(cx: Scope<Props>) -> Element {
...
+                        }})
                     }),
-                    files_list.read().iter().map(|file| {
-                        rsx!(File {
-                            text: file.name(),
-                            aria_label: file.name(),
-                            thumbnail: file.thumbnail(),
-                        })
+                   files_list.read().iter().map(|file| {
+                        let is_renaming = use_state(cx, || false);
```
sdwoodbury: ditto

kit/src/elements/file/mod.rs
sdwoodbury: also (not the fault of this PR), could the input validation just filter out everything in `chars_to_allow` before checking if alpha numeric? using `SPECIAL_CHARS` seems  unnecessary.

kit/src/elements/file/mod.rs
```diff
@@ -75,9 +75,9 @@ pub fn get_aria_label(cx: &Scope<Props>) -> String {
     cx.props.aria_label.clone().unwrap_or_default()
 }
 
-pub fn emit(cx: &Scope<Props>, s: String) {
+pub fn emit(cx: &Scope<Props>, s: String, key_code: Code) {
```
sdwoodbury, line 75: this is the right way to extend the `emit` function, though i really don't like them. i think they make it hard to follow what's happening. i think they also shouldn't be `pub`. that's my 2 cents about that. not a blocking request.

sdwoodbury (changes requested):

ui/src/layouts/storage.rs
sdwoodbury: sorry if i sounded overly bossy in the first paragraph.

lgmarchi: Very good man, I learned one more with that! 

I used rules like site is saying!

kit/src/elements/file/mod.rs
lgmarchi: I let as pub because it I followed pattern of each element and functions inside them, functions are pub

kit/src/elements/file/mod.rs
lgmarchi: I used in other places, and using in get_text as well. Maybe this function and file_stem() we should move to utils soon

kit/src/elements/file/mod.rs
lgmarchi: First question we can improve. 

Second question, we can have just alpha numeric, and allow a special char like @, or _ only, so I think special chars is necessary depending on the case yet

lgmarchi:
With new updates in hooks, now, if we click out file or folder, it will close input element, and it will not allow to open 2 input elements at same time. 

https://user-images.githubusercontent.com/63157656/220179611-d98b2e26-3b87-4aec-8583-ad02a4d56c4f.mov

ui/src/layouts/storage.rs
```diff
@@ -496,3 +501,14 @@ fn update_items_with_mock_data(
     };
     storage_state.set(Some(storage_mock));
 }
+
+fn check_true_value_in_renamig(is_renaming_map: &UseRef<HashMap<Uuid, bool>>) {
```
sdwoodbury: misspelled: `check_true_value_in_renaming`

sdwoodbury (dismissed):
Before merging, please run `cargo clippy --fix`. Doing that *shouldn't* require re-test or re-review since it doesn't change behavior. 

One observation - it appears that `is_renaming_map` is supposed to have at most one element in it at a time. In that case, the code could be simplified by changing it to an `Option<Uuid>`. For the `File` and `Folder` elements, `with_rename` could be set to `item_being_renamed == Some(file.id())`. Clearing the map could be replaced with `item_being_renamed = None`.

ui/src/layouts/storage.rs
lgmarchi: Done!

sdwoodbury (dismissed):

dariusc93 (dismissed):
LGTM

phillsatellite (dismissed):

sdwoodbury (approved):

dariusc93 (approved):

dariusc93:
Actually, if there is no change besides the imports, this could get merged

PR #326, opened 2023-02-22
feat(replies): add back end for displaying message replies and reactions

sdwoodbury:
### What this PR does 📖

- the `Message` element in `kit` now has access to the message being replied to and the message reactions
- After this PR, uplink will be ready to have the UI added for these areas. 
- each `raygun::Message` is wrapped in a struct which contains part of the message being replied to. this required minor updates across the app to switch from using `raygun::Message` to `ui_adapter::Message`

- Resolve #


common/src/warp_runner/ui_adapter/mod.rs
```diff
@@ -84,10 +110,17 @@ pub async fn conversation_to_chat(
...
     // todo: warp doesn't support paging yet. it also doesn't check the range bounds
     let unreads = messaging.get_message_count(conv.id()).await?;
-    let messages: VecDeque<raygun::Message> = messaging
+    let messages = messaging
         .get_messages(conv.id(), MessageOptions::default().set_range(0..unreads))
-        .await?
-        .into();
+        .await?;
+
+    let s = stream! {
```
dariusc93: Any reason to use the stream macro over `FuturesOrdered` or iterate over each one and store the results into a vec? Though in this case it might be better to use `FuturesOrdered` (or unordered if the order doesnt matter). Eg something like this should work

```rust
    let messages = FuturesOrdered::from_iter(
        messages
            .iter()
            .map(|message| convert_raygun_message(messaging, message).boxed()),
    )
    .collect()
    .await;
```

common/src/warp_runner/ui_adapter/mod.rs
```diff
@@ -10,8 +10,12 @@ pub use message_event::{convert_message_event, MessageEvent};
 pub use multipass_event::{convert_multipass_event, MultiPassEvent};
 pub use raygun_event::{convert_raygun_event, RayGunEvent};
 
-use std::collections::{HashMap, VecDeque};
-
+use crate::state::{self, chats};
+use async_stream::stream;
+use futures::pin_mut;
+use futures_util::stream::StreamExt;
```
dariusc93: You should be able to use `futures::StreamExt` instead since that is a reexport of this import.

ui/src/components/chat/sidebar.rs
```diff
@@ -188,8 +190,8 @@ pub fn Sidebar(cx: Scope<Props>) -> Element {
 
                     let last_message = chat.messages.iter().last();
                     let unwrapped_message = match last_message {
-                        Some(m) => m,
-                        None => &default_message,
+                        Some(m) => m.inner.clone(),
+                        None => raygun::Message::default(),
```
dariusc93, line 195: Might want to add a comment or note here that to possibly render an error instead of showing a `Message::default()`. While it shouldnt ever get to this point, if it does, its best to make sure that something is displayed for it (not relevant to this PR but just a note)

dariusc93 (commented):
Everything LGTM. I did leave a question/suggestion related to the async-stream use.

common/src/warp_runner/ui_adapter/mod.rs
sdwoodbury: this is awesome! I used the stream macro because i didn't know any better.

phillsatellite (approved):

dariusc93 (approved):

PR #335, opened 2023-02-23
feat(chats): wire up UI for replies, typing indicator, and attachments

sdwoodbury:
### What this PR does 📖

- typing indicator: shows in chatbar and sidebar. displays the user's icon rather than the username. multiple typing indicators, such as for group messages, have not yet been implemented. 
- replies: see pictures 
- attachments: allow the user to attach and send multiple files. prior to sending, attachments may be removed via the trashcan icon. 

- Resolve #

### Additional comments 🎤

image
image
image

### DIFF
--- a/common/src/warp_runner/manager/commands/raygun_commands.rs
+++ b/common/src/warp_runner/manager/commands/raygun_commands.rs
@@ -44,7 +44,7 @@ pub enum RayGunCmd {
         conv_id: Uuid,
         msg_id: Uuid,
         file_name: String,
-        directory: String,
+        directory: PathBuf,
         rsp: oneshot::Sender<Result<ConstellationProgressStream, warp::error::Error>>,
     },
     #[display(fmt = "DeleteMessage {{ conv_id: {conv_id}, msg_id: {msg_id} }} ")]
@@ -137,7 +137,7 @@ pub async fn handle_raygun_cmd(
             directory,
             rsp,
         } => {
-            let pb = PathBuf::from(directory).join(&file_name);
+            let pb = directory.join(&file_name);
             let r = messaging.download(conv_id, msg_id, file_name, pb).await;
             let _ = rsp.send(r);
         }
--- a/common/src/warp_runner/manager/commands/raygun_commands.rs
+++ b/common/src/warp_runner/manager/commands/raygun_commands.rs
@@ -3,6 +3,7 @@ use futures::channel::oneshot;
 use std::{collections::HashMap, path::PathBuf};
 use uuid::Uuid;
 use warp::{
+    constellation::ConstellationProgressStream,
     crypto::DID,
     error::Error,
     logging::tracing::log,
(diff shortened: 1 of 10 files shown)


InfamousVague:
there are a few small things left to wire but I think we can safely merge this in when you're ready @sdwoodbury

common/src/warp_runner/manager/commands/raygun_commands.rs
```diff
@@ -106,8 +116,29 @@ pub async fn handle_raygun_cmd(
...
+            let _ = rsp.send(r);
+        }
+        RayGunCmd::DownloadAttachment {
+            conv_id,
+            msg_id,
+            file_name,
+            directory,
+            rsp,
+        } => {
+            let pb = directory.join(&file_name);
```
dariusc93, line 140: It might be better for `DownloadAttachment` to use a single `PathBuf` representing the location and file where it would be stored at. I do see the reason for it and its not blocking to the PR, but just generally a suggestion.

dariusc93 (dismissed):
Looked it over and overall looks good.

common/src/warp_runner/manager/commands/raygun_commands.rs
```diff
@@ -106,8 +116,29 @@ pub async fn handle_raygun_cmd(
...
+        RayGunCmd::SendMessage {
+            conv_id,
+            msg,
+            attachments,
+            rsp,
+        } => {
+            let r = if attachments.is_empty() {
+                messaging.send(conv_id, None, msg).await
+            } else {
+                messaging.attach(conv_id, attachments, msg).await
```
dariusc93, line 128: Just as a quick note, since I just noticed it, messages here in the attach function are optional. So down the road, might want to lift it up so it could send an attachment even if a message is not present for `Raygun::attach`. Doesnt block this PR though unless its something you wish to quickly do.

common/src/warp_runner/manager/commands/raygun_commands.rs
sdwoodbury: Agreed. Especially if the user wants to rename the file when downloading it.

phillsatellite:
Just a couple things I found when testing, since this is adding in a lot as it is if it'd be easier to fix this in a new PR just lemme know and I will write up a ticket. Other than these small issues this PR is super cool!! 🎉 

(These are on Windows and MacOS)
1. When the user adds attachments they can't actually send the message unless they type some text into the input box

https://user-images.githubusercontent.com/93608357/221927355-eed33484-4ea0-49ec-8f75-f04c08a8fb1f.mov

2. When someone is on a smaller monitor or laptop, the message you type up with the attachments can look a bit messy but I do not think this is block worthy of the PR 

Screenshot 2023-02-28 at 12 12 23 PM

dariusc93 (approved):

phillsatellite (approved):

phillsatellite:
Awesome fix! 🔥

PR #339, opened 2023-02-24
fix(windows): remove transparency support . . . known issue on windows

tooshel:
### What this PR does 📖

Removes OS based transparency which we were not using anyway and was causing unbearable slowness for Windows 10 even on powerful computers.  It's also an issue on certain builds of Windows 11.

- none! Fixed it too fast! :)

I was the only one seeing the bug, likely because I'm the only one on Windows 10. According to the docs it's a known issue on Windows 10 and 11.  [Here is a link for more.](https://github.com/tauri-apps/window-vibrancy)

### DIFF
--- a/ui/Cargo.toml
+++ b/ui/Cargo.toml
@@ -21,7 +21,6 @@ extensions = { path = "../extensions" }
 tokio-util = { version = "0.7", features = ["full"] }
 arboard = "3.1.0"
 humansize = "2.1.3"
-window-vibrancy = "0.3.2"
 uuid = { version = "1.0", features = ["serde", "v4"] }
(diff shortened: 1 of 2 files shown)


phillsatellite (approved):

phillsatellite:
Couldnt replicate before and still couldn't replicate so good on my side! 👍

dariusc93 (approved):

dariusc93:
Just a note for future implementation of effects, instead of having them hard coded in and enabled byd efault, its best to look into making it toggle-able in some way. Either by a configuration flag in command line, in the config, feature flag, etc.

lgmarchi (approved):

PR #344, opened 2023-02-25
feat(DeleteItems): Delete a file or entire directory

lgmarchi:
### 1. Remove one file or entire directory tree in Storage 
a. Removing a file 

https://user-images.githubusercontent.com/63157656/221327764-2570e9f3-f914-455c-a90b-e869ab5daf21.mov

b. Removing a entire directory tree

https://user-images.githubusercontent.com/63157656/221327914-b4f6be17-173e-471f-a6bd-bb4479fd8cde.mov

- 

- Resolve #


common/src/warp_runner/manager/commands/constellation_commands.rs
```diff
@@ -114,7 +123,114 @@ pub async fn handle_constellation_cmd(cmd: ConstellationCmd, warp_storage: &mut
...
+    // Code keeps here just if item is a directory
+    let first_dir = warp_storage.current_directory()?;
+    let mut current_dirs_opened = get_directories_opened();
+    current_dirs_opened.push(first_dir.clone());
+
+    let mut dirs: Vec<Directory> = current_dirs_opened;
+
+    match warp_storage.select(&item.name()) {
+        Ok(_) => log::debug!("Selected new dir: {:?}.", item.name()),
+        Err(error) => log::error!("Error to select new dir: {:?}.", error),
```
dariusc93: may want to have it return an error here or possibly block continuation in some way to prevent any unexpected behaviour.

common/src/warp_runner/manager/commands/constellation_commands.rs
```diff
@@ -114,7 +123,114 @@ pub async fn handle_constellation_cmd(cmd: ConstellationCmd, warp_storage: &mut
...
             let _ = rsp.send(r);
         }
+        ConstellationCmd::DeleteItems { item, rsp } => {
+            let r = delete_items(warp_storage, item).await;
+            let _ = rsp.send(r);
+        }
+    }
+}
+
+async fn delete_items(
```
dariusc93, line 133: Something that may want to be done is push this off into its own task. Doesnt have to be apart of this PR since it may require tracking the actual task itself. but ideally you want to make sure it doesnt block the response. Eg, if somebody has 1000 files uploaded, and begin deleting them, while its async, other task wouldnt be blocked (since its all in a scheduler), it would block a oneshot receiver in warp runner event loop until this function actually returns successfully or with an error. Maybe could set a limit to determine if a task should be spawned or continue as is with the result returning an `Either`.

Maybe @sdwoodbury would want to chime in on this with his thoughts. 

For now, not really a blocker to this PR but just a thought for the future

common/src/warp_runner/manager/commands/constellation_commands.rs
```diff
@@ -114,7 +123,114 @@ pub async fn handle_constellation_cmd(cmd: ConstellationCmd, warp_storage: &mut
...
+    warp_storage: &mut warp_storage,
+    item: Item,
+) -> Result<uplink_storage, Error> {
+    // If is file, just a small function solve it
+    if item.is_file() {
+        let file_name = item.name();
+        match warp_storage.remove(&file_name, true).await {
```
dariusc93: Just an fyi, In https://github.com/Satellite-im/Warp/tree/warp/fs-removal, I will be working on allowing deleting directories and its item within (if recursive is used), so as a precaution, it may be best to set this recursive to `false`.

common/src/warp_runner/manager/commands/constellation_commands.rs
```diff
@@ -68,6 +72,11 @@ pub enum ConstellationCmd {
         local_path_to_save_file: PathBuf,
         rsp: oneshot::Sender<Result<(), warp::error::Error>>,
     },
+    #[display(fmt = "RemoveItems {{ item: {item:?} }} ")]
```
sdwoodbury: `RemoveItems` ought to be `DeleteItems` so that we don't get confused about what happened, when reading the log.

common/src/warp_runner/manager/commands/constellation_commands.rs
sdwoodbury: The scheduler for `warp_runner`  passes the same mutable reference to each potential task. The commands for Constellation doesn't seem to need Multipass or Raygun, so perhaps it could be split off onto its own task. It might not be so bad to have the user wait until files are removed. We would at least want to prevent the user from attaching a file which is queued to be deleted. Even if Constellation is moved to its own task, allowing the UI to show alerts for new messages and friend requests, it still may be wise to have the user be stuck on a "loading" page until the delete operation is either cancelled or completed. 

image

phillsatellite (dismissed):

phillsatellite:
Worked well on my side @lgmarchi really neat!

common/src/warp_runner/manager/commands/constellation_commands.rs
sdwoodbury, line 133: i tested this function on Linux and it seems to work. much simpler:
```
fn delete_items(
    warp_storage: &mut warp_storage,
    item: Item,
) -> Result<uplink_storage, Error> {
    let cwd = warp_storage.current_directory()?;
    cwd.remove_item(&item.name())?;
    get_items_from_current_directory(warp_storage)
}
```

can we use this instead? requesting changes here but will approve if for some reason this doesn't work on a different platform.

sdwoodbury (changes requested):

common/src/warp_runner/manager/commands/constellation_commands.rs
lgmarchi: It just remove item from that directory, but don't delete them from File System. Because of this is necessary to run through all directories and delete each item per time

sdwoodbury (dismissed):

dariusc93 (dismissed):

phillsatellite (approved):

common/src/warp_runner/manager/commands/constellation_commands.rs
lgmarchi: Ticket opened about this issue. 

https://github.com/Satellite-im/Uplink/issues/345

PR #354, opened 2023-03-01
fix(input): make separate elements for input and textarea

sdwoodbury:
### What this PR does 📖

- `type: password` has no effect on `textarea`. the unlock page needs an `input` element. 
image

- also, when displaying part of a message in the sidebar, only show the first line. 
image

- Resolve #

### Additional comments 🎤
I checked out commit `aa92a35ce0f14c1497efb299847995458acd4bfa`, copied `kit/src/elements/input`, checked out  `dev`, renamed `kit/src/elements/input` to `kit/src/elements/textarea`, and then added the old `input` folder. 

I then made the following modifications
- for clean up
    - removed the redundant `#[props(optional)]` but that has no effect. 
    - made the helper functions in `input` and `textarea` private
- behavior changes
    - removed the `type` field from `textarea` and the `is_password` prop.  
    - removed the validation from `textarea` -> was duplicate code and I think it may only be needed for the regular `input`


phillsatellite:
Hey Stu! Working pretty well so far! Was curious though should we show a message in the sidebar if the latest message is an attachment? I think I recall in Core-PWA we had "Username has sent an attachment" 

Screenshot 2023-03-01 at 11 47 19 AM

phillsatellite:
Attachment message looks good! I noticed one small thing with it, in the video I am on my account TestingMac and whenever I send attachments the sidebar will say that the Windows user has sent attachments even though it was TestingMac who sent them. This is the same behavior for the Windows account whenever they send bulk attachments it will say that TestingMac has sent attachments 

https://user-images.githubusercontent.com/93608357/222221295-1ef5b6ce-3a13-4645-8989-a778b3f0b40b.mov

InfamousVague (changes requested):
My biggest complaint here is that I propose we instead add a "multi-line" or some other prop to the existing input component in order to determine if we render a textarea or an input. I'm not a fan of duplicating the entire element into a new module for such a small thing because it adds two places we'll need to maintain in the future. I think I need to understand a bit better why the input broke in the first place to propose the best solution, though.

InfamousVague (approved):

phillsatellite (approved):

phillsatellite:
Tested every input we have available and the chat and replies are the only ones with the mullti-line support 👍  This also seems to fix a bug where the input boxes in Settings would adjust sizes as you start to type that I was about to make a ticket for. Awesome PR! 🔥

Issue #204, opened 2023-01-27
Friends - Add error message when User tries to add themselves or already friends


phillsatellite:
**Task:** If we copy our own did and then try to add ourselves or someone we are already friends with we should have a error message appear. I think for chore I remember us having "You cant add yourself, silly goose" 

https://user-images.githubusercontent.com/93608357/215173715-c68f5f87-1ea3-48ab-af29-1dc5dfaeb157.mov

dariusc93:
Resolved in #357

PR #357, opened 2023-03-01
fix(error): toast error alert on add friend errors

Jekrimo:
### What this PR does 📖

- Trying to add ourselves or someone we are already friends with we should have a error message appear.

- Resolve #

#204 

### Additional comments 🎤

Stu is the goat


ui/src/components/friends/add.rs
```diff
@@ -89,12 +102,19 @@ pub fn AddFriend(cx: Scope) -> Element {
                         | Error::BlockedByUser
                         | Error::InvalidIdentifierCondition
                         | Error::CannotSendSelfFriendRequest => {
```
dariusc93: The other error conditions would be irrelevant (though it looks like it been there for awhile before this PR). May be best to remove all except  `Error::CannotSendSelfFriendRequest` unless there is a reason on why those would need to be there, and if it does need to remain, then the text below should change to something else to add more context.

ui/src/components/friends/add.rs
```diff
@@ -89,12 +102,19 @@ pub fn AddFriend(cx: Scope) -> Element {
                         | Error::BlockedByUser
                         | Error::InvalidIdentifierCondition
```
sdwoodbury: I agree that some of the match arms can be deleted so that the error would fall through to the  `_ => ` arm. It might be helpful to have a custom toast to notify the user that they are blocked.

phillsatellite:
The function works really good! And the error for when you try to add yourself looks perfect! The only thing I noticed when testing was when you try to add someone that you are already friends with the error message says "Adding friend failed" I think it might be better for the user if it said something like "You are already friends!" or something similar. Pretty much what Darius had mentioned above just wanted to leave my notes from my findings as well 🙂

Screenshot 2023-03-02 at 2 26 12 AM

ui/src/components/friends/add.rs
Jekrimo: Totally agree, I'll clean it

ui/src/components/friends/add.rs
Jekrimo: I could see some people not wanting someone to know that they are blocked though right? Like some stalker checking to see if someone is on the app or something. Maybe make it more vague, idk?

Jekrimo:
> 

totally! updated it! I think we will definitely want to go though all of them and get all the wording across the app feeling like it's coming from the same source if that makes sense?

InfamousVague (approved):
Looks good to me, can we add a comment above the if let some statement describing it a bit? What is the 5?

phillsatellite (approved):

phillsatellite:
> > 
> 
> totally! updated it! I think we will definitely want to go though all of them and get all the wording across the app feeling like it's coming from the same source if that makes sense?

Sounds like a plan to me! We can discuss kinda what we want all the errors to say and I can make a checklist style ticket to confirm each of those errors is added/appearing correctly 😄

dariusc93 (approved):

PR #361, opened 2023-03-02
Art unlock

InfamousVague:
image


InfamousVague:
"Unknown" needs to be implemented but our release date but I figured it's obvious enough when left there, that we won't forget. If anyone knows how to get this working happy to add it to the PR! It can say something like "Let's choose your password" if there is no known person.

phillsatellite:
Awesome PR 🔥  Besides the Unknown at login and the Welcome message in Settings persisting after you click Dismiss, I only found one other thing so far that I couldn't replicate on dev branch 

If you switch to mobile view it doesn't fit to screen and allows the user the ability to horizontally scroll on that page 

https://user-images.githubusercontent.com/93608357/222345304-47d60f47-4279-4bbc-88ce-01e2113e3b53.mov

InfamousVague:
I tried the best I could to clean up the CSS to be responsive hopefully this helps enough.

InfamousVague:
I also made various other small fixes along the way that were too small to break out into their own thing.

phillsatellite:
Per Matts request I'm gonna create a ticket it does fit better for mobile view now but I think it would look best with the Dismiss button underneath of the Welcome description 

image

phillsatellite (dismissed):

ui/src/layouts/unlock.rs
```diff
@@ -172,6 +169,7 @@ pub fn UnlockLayout(cx: Scope, page: UseState<AuthPages>, pin: UseRef<String>) -
                         options: Options {
                             with_validation: Some(pin_validation),
                             with_clear_btn: true,
+                            with_label: Some("Welcome back, UNKNOWN"), // TODO: Implement this.
```
lgmarchi: It is missing translation here

ui/src/main.rs
```diff
@@ -371,6 +371,7 @@ pub fn app_bootstrap(cx: Scope) -> Element {
     // todo: perhaps when the user resizes the window, store that in State, and load that here
     let desktop = use_window(cx);
     desktop.set_inner_size(LogicalSize::new(950.0, 600.0));
+    desktop.set_min_inner_size(Some(LogicalSize::new(300.0, 500.0)));
```
lgmarchi, line 376: min_inner_size should be (300.0, 350.0)?

common/locales/en-US/main.ftl
```diff
@@ -87,7 +87,10 @@ settings-profile = Profile Settings
     .change-avatar = Change avatar
     .edit-button = Edit
     .save-button = Save
-
+    .welcome = Your new profile!
+    .welcome-desc = Tell the world all about yourself, well tell them as much as you can while we're still under construction, at least.
+    .welcome-cta = First step, pick out a profile picture and maybe even a banner too!
```
lgmarchi, line 93: Would be nice, to add this variables to other languages as well, for us not lost which new variables were created.

ui/src/components/settings/sub_pages/profile/mod.rs
```diff
@@ -48,12 +48,46 @@ pub fn ProfileSettings(cx: Scope) -> Element {
...
+                    div {
+                        class: "welcome",
+                        img {
+                            src: "./ui/extra/images/mascot/working.png"
+                        },
+                    },
+                    div {
+                        class: "welcome-content",
+                        Button {
+                            text: "Dismiss".into(),
```
lgmarchi: Missing translation here

lgmarchi (commented):
Did some comments about small changes

ui/src/layouts/unlock.rs
InfamousVague: I could not get this to work properly so left it as a TODO for now

ui/src/main.rs
InfamousVague: Not in this instance since we're on the main routes, that is too small for us.

lgmarchi:
Now is possible to see difference in message when you are creating account or doing a login: 

image

phillsatellite (approved):

Issue #366, opened 2023-03-02
Global - Logo PNG missing if you do cargo build instead of run


phillsatellite:
**Issue:** Like the title says, if you do cargo build to open the app instead of doing cargo run the welcome PNG will be missing on the landing page

**STR:**
Run cargo build 
Open the Uplink app
PNG will be missing 

**Screenshot:**

![image](https://user-images.githubusercontent.com/93608357/222523514-898bcb34-d9de-48ee-ac18-6ae6c4580c09.png)

tooshel:
It seems like dioxus is recommending we take the images and make them part of the install process. Right now I was able to use the new "dioxus build" command to generate an MSI and soon I'll be able to bundle assets (like images) with that installer.

tooshel:
This is related to how we build the apps and Stuart is looking into that now. It's not just the images that are missing . . . fonts are missing too. All of those assets are being referenced in the code with links relative to the root of the uplink project and we need to copy those files and folders over to wherever uplink is installed.

PR #372, opened 2023-03-03
chore(Emotes): Added mascot emotes now so we don't lose them

InfamousVague:
### What this PR does 📖

Added mascot emotes now so we don't lose them


dariusc93 (approved):
Though not related to this PR, might be best to make sure apart of the binary by statically included into the UI (to prevent something like #366 down the road)

PR #375, opened 2023-03-03
chore(cargo): Update dependency

dariusc93:
### What this PR does 📖

- Refactors `RayGun::send` into 2 functions `RayGun::send` and `RayGun::edit` (which should prevent possibly confusion when it comes to implementing editing functionality)
- Added `Tesseract::subscribe` to get a stream to determine if the store been locked or unlocked (useful without needing to poll/call `Tesseract::is_unlocked` in a loop, though there is no performance impact if done)
- Bug fixes related to profile picture and banner (discovered while helping @sdwoodbury in `feat/profile`) 

### Which issue(s) this PR fixes 🔨

- N/A

### Special notes for reviewers 🗒️
Though should not be any difference (since the change is literally minor), test unlocking with both debug and release profile.

Issue #362, opened 2023-03-02
Friends - Green indicator should clear after request is sent 


phillsatellite:
**Issue:** After you send a friend request and the DID clears from the input box, the green outline that appeared should clear out with the DID key 

**STR:** 
-Send friend request 

**Screen recording:**

https://user-images.githubusercontent.com/93608357/222485576-1733fe0b-0c29-48fb-9088-7c46ddd5fe71.mov

PR #378, opened 2023-03-03
feat(profile): wire up profile to warp

sdwoodbury:
### What this PR does 📖

- allow user to update Identity: status, profile picture, banner 
- identities for friends are refreshed every minute
- the profile picture and status appear in the top bar 
- the pfp in the sidebar isn't implemented yet, and should probably be held off on until we decide if we will change the data model in `state` per discussions on the Uplink private discord channel. 
- same goes for other places that the profile picture and online status need to be wired up

- Resolve #

### Additional comments 🎤
- the input box for user status remains green after "enter" is pressed. this eventually needs to be changed. 
- the user's Identity is now fetched from Warp (minor change). seems the other stuff - whether the user is online and their platform - can be determined at runtime and doesn't need to be saved.


dariusc93:
~~Note: This should not merge until #375 has been merged into dev~~

phillsatellite (dismissed):

phillsatellite:
Approved on my side, I did have some issues here and there where a friend would go missing and not appear for about 3 mins then it seemed like Warp finally loaded the User info. I think this is more of a network issue on my side because I tested on dev and had the same issues. When Warp was running really good then every function on this PR would work really well. I can create a ticket for the wiring pfp to the sidebar and other needed places if thats cool with you @sdwoodbury was waiting for the conclusion for the state discussion  I did mention to Stu that I have ticket #362 that I will be adding more info about the green indicator around the input boxes.

dariusc93:
> Approved on my side, I did have some issues here and there where a friend would go missing and not appear for about 3 mins then it seemed like Warp finally loaded the User info. I think this is more of a network issue on my side because I tested on dev and had the same issues. When Warp was running really good then every function on this PR would work really well. I can create a ticket for the wiring pfp to the sidebar and other needed places if thats cool with you @sdwoodbury was waiting for the conclusion for the state discussion I did mention to Stu that I have ticket #362 that I will be adding more info about the green indicator around the input boxes.

Hard to really say if its an issue with warp or if its related to uplink. How often does this happen? Is it when you send, receive, or after accept a request that it disappears or not show up? Does anything appear in the logs?

common/src/state/identity.rs
```diff
@@ -69,4 +72,8 @@ impl Identity {
     pub fn set_platform(&mut self, platform: Platform) {
         self.platform = platform;
     }
+
+    pub fn set_warp_identity(&mut self, ident: multipass::identity::Identity) {
```
dariusc93, line 76: `Identity::from` (and to that extent, `Identity::into`) could be used since it would being `multipass::identity::Identity` into this wrapper.

common/src/warp_runner/manager/commands/multipass_commands.rs
```diff
@@ -125,9 +165,90 @@ pub async fn handle_multipass_cmd(cmd: MultiPassCmd, warp: &mut super::super::Wa
...
     }
 }
 
+async fn multipass_refresh_friends(
+    account: &mut Account,
+) -> Result<HashMap<DID, multipass::identity::Identity>, Error> {
+    let ids = account.list_friends().await?;
+    let mut friends = HashMap::new();
+    for id in ids {
+        if let Ok(ident) = account.get_identity(id.clone().into()).await {
```
dariusc93: Could `Identifier::did_keys` be better here? That way you could pass `ids` into the identifier when calling `MultiPass::get_identity` to get the list itself then construct the hashmap from that (or just the list of `Identity`?)

eg something like this should suffice 

```rust
async fn multipass_refresh_friends(
    account: &mut Account,
) -> Result<HashMap<DID, multipass::identity::Identity>, Error> {
    let ids = account.list_friends().await?;

    let list = account.get_identity(Identifier::did_keys(ids.clone())).await?;
    let friends: HashMap<DID, identity::Identity> = list.iter().map(|ident| (ident.did_key(), ident.clone())).collect();
    if list.is_empty() {
        log::warn!("No identities found");
    }
    Ok(friends)
}

```

dariusc93: Note: Could also compare the `ids` against the map and log any identities not found too.

dariusc93 (dismissed):
LGTM. Left some comments and suggestions but non-blocking to this PR

common/src/warp_runner/manager/commands/multipass_commands.rs
sdwoodbury: I like this suggestion. I didn't realize I could do this. That's a nice feature of MultiPass.

tooshel (approved):

PR #379, opened 2023-03-03
feat(DragDrop): Drag drop upload files

### LINKED ISSUE #346: Files - Upload files with drag and drop
Upload files with drag and drop

lgmarchi:
### 1. Upload files using drag and drop 

https://user-images.githubusercontent.com/63157656/222801679-cb3eb664-2216-43dc-a766-4f9dc65aa3b1.mov

### 2. Upload feedback works with file explorer as well 

https://user-images.githubusercontent.com/63157656/222801945-b5b9f329-d57a-48f1-b7ce-09ed2807fc07.mov

### 3. Fix small bug if folder name is very big
image

- 

- Resolve #346


phillsatellite:
Hey Lucas! Worked perfect on MacOS! On Windows I had an issue 

When I drag and drop files in, you can see in the log that its registering me hovering/dropping the files but they will never actually appear in my uploaded files. Manually uploading still works fine though. I did try waiting a while for it to possibly update, closed and re opened the app and still nothing 

https://user-images.githubusercontent.com/93608357/222834483-cc066845-1161-4c1c-83d8-06b6b93ea2c2.mp4

common/src/warp_runner/manager/commands/constellation_commands.rs
```diff
@@ -441,20 +483,49 @@ async fn upload_files(
...
                 }
-                match set_thumbnail_if_file_is_image(warp_storage, filename.clone()).await {
-                    Ok(_) => log::info!("Image Thumbnail uploaded"),
-                    Err(error) => log::error!("Error on update thumbnail for image: {:?}", error),
-                }
-                match set_thumbnail_if_file_is_video(warp_storage, filename.clone(), file_path) {
-                    Ok(_) => log::info!("Video Thumbnail uploaded"),
-                    Err(error) => log::error!("Error on update thumbnail for video: {:?}", error),
+
+                let image_thumb_uploaded =
```
sdwoodbury: when I upload something that isn't a picture or video, Uplink gives warning messages. This isn't a blocker but one alternative is to check the extension type before calling `set_thumbnail_if_file_is_image` or  `set_thumbnail_if_file_is_video`. This way, if those functions return an error, we then know that we should pay attention to it.  

I notice that the image extensions can be checked via  `ImageFormat::from_extension()` and video extensions can be compared to `VIDEO_FILE_EXTENSIONS`.

again, not a blocker. I'm fine if this isn't done at all or is done in another PR. 

image

common/src/warp_runner/manager/commands/constellation_commands.rs
```diff
@@ -441,20 +483,49 @@ async fn upload_files(
...
+                            true
+                        }
+                        Err(error) => {
+                            log::warn!("Not possible to update thumbnail for image: {:?}", error);
+                            false
+                        }
+                    };
+
+                if !image_thumb_uploaded {
+                    match set_thumbnail_if_file_is_video(warp_storage, filename.clone(), file_path)
```
sdwoodbury: I tested the thumbnail functionality on Linux with an image and video. When I uploaded a `.mkv`, the thumbnail wasn't displayed. 

image

image

sdwoodbury:
review in progress

common/src/warp_runner/manager/commands/constellation_commands.rs
sdwoodbury: ah, because I don't have ffmpeg

image

Maybe this isn't a problem. But we should make note of having ffmpeg as a dependency for later. We may need that information when creating an installer for Uplink.

ui/src/layouts/storage.rs
```diff
@@ -604,3 +735,58 @@ fn update_items_with_mock_data(
...
+        file_name_formatted = match &file_name.get(0..15) {
+            Some(name_sliced) => format!(
+                "{}...",
+                name_sliced,
+            ),
+            None => file_name.clone(),
+        };
+    }
+    file_name_formatted
+}
```
sdwoodbury: not a blocker, but the built in functions for `Option`  could simplify the calculation of `file_name_formatted`  - mainly `unwrap_or_else`. 

```
fn format_item_name(file_name: String) -> String {
    let item = PathBuf::from(&file_name);

    let file_stem = item
        .file_stem()
        .and_then(OsStr::to_str)
        .map(str::to_string)
        .unwrap_or_default();

    file_name
        .get(0..15)
        .map(|x| x.to_string())
        .map(|x| {
            if file_stem.len() > MAX_LEN_TO_FORMAT_NAME {
                format!("{x}...")
            } else {
                x
            }
        })
        .unwrap_or_else(|| file_name.clone())
}
```

ui/src/layouts/storage.rs
```diff
@@ -300,20 +386,64 @@ pub fn FilesLayout(cx: Scope<Props>) -> Element {
...
+        if !STATIC_ARGS.use_mock {
+            *directories_list.write_silent() = storage.directories.clone();
+            *files_list.write_silent() = storage.files.clone();
+            *current_dir.write_silent() = storage.current_dir.clone();
+            *dirs_opened_ref.write_silent() = storage.directories_opened.clone();
+        };
+        state.write().storage = storage;
+        storage_state.set(None);
+    }
+    let window = use_window(cx);
```
sdwoodbury: this is redundant. the same `window` was define on line 113.

sdwoodbury (changes requested):
Nice work. I left some non-blocking suggestions. I know this isn't your fault but there's a ton of clippy warnings, all but one of which can be fixed automatically. I'll send over a patch. 

2 more non-blocking suggestions which might be better off in their own PRs
- the progress box might be more visually pleasing if in the center of the screen rather than the bottom left corner. 
- the file names on the storage page are hard to see because the font is so thin and so dark. a brighter and perhaps thicker font would help with this. 

edit: I think that `warp_runner` should use [async_process](https://docs.rs/async-process/latest/async_process/) when running ffmpeg.

common/src/warp_runner/manager/commands/constellation_commands.rs
lgmarchi: Opened a PR with instructions on readme!

https://github.com/Satellite-im/Uplink/pull/390

sdwoodbury (approved):

phillsatellite (approved):

PR #380, opened 2023-03-03
feat(messages): edit messages

sdwoodbury:
### What this PR does 📖

- adds the ability to edit messages
image
image

- deleting all text causes the message to be deleted rather than edited
- the edit can be cancelled via the context menu
image

- Resolve #

### Additional comments 🎤
- I had trouble focusing the cursor at the end of the textarea. it would be nice to have that but i'm stumped. maybe it can be done later.

### DIFF
--- a/.github/workflows/cargo.yml
+++ b/.github/workflows/cargo.yml
@@ -9,6 +9,8 @@ jobs:
     name: Cargo Checks
     runs-on: ubuntu-latest
     steps:
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v1
       - name: Checkout sources
         uses: actions/checkout@v2
       - name: Run cargo checks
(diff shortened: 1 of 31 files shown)


phillsatellite:
Really neat, the only things I found were very minor, 

1. We should probably have a way to exit editing a message in case the User would change their mind. 

https://user-images.githubusercontent.com/93608357/223184674-687cc7ee-1798-4598-8653-9772cd3f22c5.mov

2. This I could only replicate on MacOS, when there is a large message and you right+click to edit it will highlight 1/3 of the message

https://user-images.githubusercontent.com/93608357/223185233-446d9399-2968-4a1b-b199-1e0c14157f08.mov

dariusc93 (dismissed):

phillsatellite:
I talked with Stu, fixing the things I mentioned above seem to be a little more difficult than we thought it would be, going to write up a ticket for what Stu had mentioned in the Reviewing Notes (focusing the curser at the of the textare) and aother for Dismissing edit modal when user clicks away from textarea 👍

phillsatellite (approved):

PR #384, opened 2023-03-04
chore(cargo): Restructure workspace Cargo.toml 

dariusc93:
### What this PR does 📖

- Restructures Cargo.toml, moving dependencies into the workspace Cargo.toml

- N/A

There is no testing required since this is just restructuring the workspace configuration and not changing the underlining code itself

### Additional comments 🎤
This change would eventually allow the workspace be more organize and allow for a better path in updating the common dependencies used in the workspace. This would also prevent mismatch of dependencies where one crate is using a specific version than the other.

Note: Its best not to push every single dependency into the workspace Cargo.toml, but rather common dependencies that are used across other crates or members in the workspace.


InfamousVague:
This looks so much better than the current solution. Didn't know about the `*.workspace = true`

dariusc93:
> This looks so much better than the current solution. Didn't know about the `*.workspace = true`

It was stabilized last year in 1.64 or 1.65, but knew there was a bit more discussions on it so never dove into it at that time (nor had a need for it) but should help keep things clean

phillsatellite (approved):

InfamousVague (approved):

PR #394, opened 2023-03-06
feat(language): Make adding languages easier

Flemmli97:
### What this PR does 📖

The goal of this PR is to make adding more language easier. For the previous version to add another language one had to change things at 3 location:
1. Create a const
2. Add the const to the Hashmap of `change_language::change_language`
3. Add the const to `change_language::get_available_languages` 

With this new version the only thing thats needed is to add the new language to the map of `LANGUAGES`. One could still define the language as a constant in addition to that (if e.g. a reference of it is needed elsewhere like the default english lang) but that is not needed.

Something to note is that the LANGUAGES is now a static map so it looses advantages of constant lang fields (like compiler inlining) but i dont think this will be much an issue.

### DIFF
--- a/common/src/language/mod.rs
+++ b/common/src/language/mod.rs
@@ -1,25 +1,36 @@
 use std::collections::HashMap;
 
-use fluent_templates::{once_cell::sync::Lazy, LanguageIdentifier, Loader};
+use fluent_templates::{
+    lazy_static::lazy_static, once_cell::sync::Lazy, LanguageIdentifier, Loader,
+};
 use unic_langid::langid;
 use warp::sync::RwLock;
 
 use crate::LOCALES;
 
 pub const US_ENGLISH: (LanguageIdentifier, &str) = (langid!("en-US"), "English (USA)");
-const BR_PORTUGUESE: (LanguageIdentifier, &str) = (langid!("pt-BR"), "Português (Brasil)");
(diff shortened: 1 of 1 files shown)


common/src/language/mod.rs
```diff
@@ -1,25 +1,36 @@
...
 use warp::sync::RwLock;
 
 use crate::LOCALES;
 
 pub const US_ENGLISH: (LanguageIdentifier, &str) = (langid!("en-US"), "English (USA)");
-const BR_PORTUGUESE: (LanguageIdentifier, &str) = (langid!("pt-BR"), "Português (Brasil)");
-const PT_PORTUGUESE: (LanguageIdentifier, &str) = (langid!("pt-PT"), "Português (Portugal)");
-const MX_SPANISH: (LanguageIdentifier, &str) = (langid!("es-MX"), "Español (México)");
+
+lazy_static! {
```
dariusc93: It might be worth nothing that using `Lazy` from `once_cell` could be used over `lazy_static`, and in some cases is preferred due to the flexibility (and would help reduce the 2 different variants).

Flemmli97: changed it

dariusc93 (approved):

phillsatellite (approved):

sdwoodbury (approved):

PR #398, opened 2023-03-07
chore(state): unify data model concerning identities

sdwoodbury:
### What this PR does 📖

- Identities need to be updated with the user's username, status, profile picture, banner, and whether or not they are online. Identities were previously stored in both the friends list and with every chat, and the two lists didn't necessarily overlap. Obviously this is bad. 
- A `HashMap<DID, Identity>` was added to state. everywhere else in state, `Identity` was replaced with `DID`. 
- A few fields in state were made private because altering them may (either now or in the future) require updating `state.identities`. 
- Lots of utility functions were added to state in response to these fields becoming private. 
- If you read this far, good job! 
- There are also miscellaneous changes due to `cargo fmt` and `cargo clippy`

- Resolve #

### DIFF
--- a/common/src/state/action.rs
+++ b/common/src/state/action.rs
@@ -103,8 +103,8 @@ pub enum Action {
     CancelRequest(Identity),
 
     /// Handle a new incoming friend request
-    #[display(fmt = "IncomingRequest")]
-    IncomingRequest(Identity),
+    // #[display(fmt = "IncomingRequest")]
+    // IncomingRequest(Identity),
     /// Accept an incoming friend request
     #[display(fmt = "AcceptRequest")]
     AcceptRequest(Identity),
--- a/common/src/state/chats.rs
+++ b/common/src/state/chats.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::{HashMap, VecDeque},
+    collections::{HashMap, HashSet, VecDeque},
     time::Instant,
 };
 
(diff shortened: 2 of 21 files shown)


phillsatellite (approved):

common/src/state/action.rs
```diff
@@ -103,8 +103,8 @@ pub enum Action {
     CancelRequest(Identity),
 
     /// Handle a new incoming friend request
-    #[display(fmt = "IncomingRequest")]
-    IncomingRequest(Identity),
+    // #[display(fmt = "IncomingRequest")]
+    // IncomingRequest(Identity),
```
InfamousVague, line 107: why comment this out? Should we just remove it if it's dead code?

common/src/state/chats.rs
```diff
@@ -41,10 +39,10 @@ pub struct Chat {
 // warning: Chats implements Serialize
 #[derive(Clone, Debug, Default, Deserialize)]
 pub struct Chats {
-    #[serde(default)]
+    #[serde(skip)]
     pub initialized: bool,
     // All active chats from warp.
-    #[serde(default)]
+    #[serde(skip)]
```
InfamousVague, line 45: we would want to cache these in state so we don't have to call warp every time I would think, doesn't this break that flow?

InfamousVague (approved):
Some comments, looks good otherwise

common/src/state/chats.rs
sdwoodbury: warp doesn't have to be called every time, only once on startup...

common/src/state/action.rs
sdwoodbury: i didn't know if something was planned for later, but I'm fine with deleting it.

InfamousVague: Let's nuke it 👍

PR #403, opened 2023-03-07
feat(identity): show online indicator and platform

sdwoodbury:
### What this PR does 📖

- makes the remote user's profile picture light up green when they are online
image

- Resolve #


phillsatellite:
Hey Stu! Just some 2 minor things, 

So it worked exactly as it should I just noticed sometimes it would take a good bit of time to update/sometimes it would update instantly. Obviously not block worthy just wanted to mention it. 

The other thing I had on both MacOS and Windows, when I would change views from desktop to mobile it would never register the icon would remain the same as it was for desktop. In screenshot below the Mr.Windows is in mobile view while Macc is on Desktop

Screenshot 2023-03-07 at 4 11 48 PM

sdwoodbury:
@phillsatellite  I don't believe `Platform` changes. In Warp it's checked at compile time. Even if I lie to `State` about its own platform, that change won't get sent to the peer. 
image
image


Update: The user's platform can now be mocked locally
image

From this it seems that the platform is usually "Unknown" when it should be desktop. We might want to investigate that in a separate PR. I double checked to make sure the platform wasn't getting set to Default somewhere on the Uplink side

dariusc93:
> @phillsatellite I don't believe `Platform` changes. In Warp it's checked at compile time. Even if I lie to `State` about its own platform, that change won't get sent to the peer. 

Youre correct. As you saw, `Platform` itself doesnt change as its based on the platform its compiled for rather than introducing a interface to allow changing it.


> Hey Stu! Just some 2 minor things,
> 
> So it worked exactly as it should I just noticed sometimes it would take a good bit of time to update/sometimes it would update instantly. Obviously not block worthy just wanted to mention it.
> 
> The other thing I had on both MacOS and Windows, when I would change views from desktop to mobile it would never register the icon would remain the same as it was for desktop. In screenshot below the Mr.Windows is in mobile view while Macc is on Desktop
> 

From what I was told, the mobile view is just about the resizing of the window itself rather than trying to emulate a `Platform`, so what you see there is intended.

common/src/state/mod.rs
```diff
@@ -1071,16 +1075,37 @@ impl State {
...
+    pub fn mock_own_platform(&mut self, platform: Platform) {
+        if let Some(ident) = self.identities.get_mut(&self.did_key()) {
+            ident.set_platform(platform);
+        }
+    }
+
+    // identities are updated once a minute for friends. but if someone sends you a message, they should be seen as online.
+    // this function checks if the friend is offline and if so, sets them to online. This may be incorrect, but should
+    // be corrected when the identity list is periodically updated
+    pub fn update_identity_status_hack(&mut self, id: &DID) {
```
dariusc93, line 1129: Not sure the intended reason for this, but using `MultiPass::identity_status` can provide the actual identity status to show if they are online, offline as well as away or busy (which can be set through `MultiPass::set_identity_status`). If they are offline (eg closed down the client and no longer connected), it wont show away or busy since it is based on their actual connection status first before the metadata.

dariusc93: Actually, looking through the code, it makes a little bit more sense on the use of this function, though im wondering if there is a better way on handling this.

sdwoodbury: yea, this function avoids an extra call to warp_runner. I think that eventually, when events are sent as an identity is updated, this function would be removed.

phillsatellite:
Thanks for clarifying @dariusc93 and @sdwoodbury , since mobile is not a priority atm and its not back tracking mobile on dev branch I dont think its something thats worthy of blocking the PR right now. I just had a few questions

In the screen recording you can see when I first logged in all of the indicators are the hammer/screwdriver but then when I switch to mobile view and back to desktop my indicator changes to a green bubble. Is the green bubble supposed to be the default indicator for desktop instead of that hammer/screwdriver? 

Another thing i wanted to point out (this is more of something I imagine we would wanna do another ticket for lemme know what you guys prefer) Anyone added in your favorites will have the same indicator as you (this seems to be a windows only problem)

https://user-images.githubusercontent.com/93608357/223776295-0013654a-2434-4905-97ce-85a67dda6a11.mov

phillsatellite (dismissed):

dariusc93 (dismissed):

ui/src/utils/mod.rs
```diff
@@ -103,13 +95,12 @@ mod test {
...
-        let r = get_pretty_name("pretty/name1.scss");
-        assert_eq!(r, String::from("name1"));
-    }
-
-    #[test]
-    fn test_get_pretty_name_windows() {
-        let r = get_pretty_name("c:\\pretty\\name2.scss");
-        assert_eq!(r, String::from("name2"));
+        if cfg!(windows) {
+            let r = get_pretty_name("c:\\pretty\\name2.scss");
```
InfamousVague, line 99: looks good to me, but what is this doing?

sdwoodbury: conditionally executing if on windows

InfamousVague: Right but what's the get_pretty_name used for, I assume this is just a test but is the pretty name used for files or something?

InfamousVague (approved):

PR #414, opened 2023-03-08
feat(cache): export cache

### LINKED ISSUE #227: Settings - Wire up "Compress & Download Cache"
**Task:** Like Title says we want to wire up Compress and Download Cache in Settings/Developer. We should include a notification when User clicks button

**Screenshot:**
Screenshot 2023-02-01 at 6 26 52 PM

sdwoodbury:
### What this PR does 📖

- This PR is a first pass at exporting state. It could use a loading indicator but otherwise it seems to work. 

- Resolve #227 

### Additional comments 🎤
to test, create a folder like `uplink_cache`, then unzip `uplink.zip` into `uplink_cache/.user`. 

at some point we might want a download progress indicator 

at some point we might want to also zip themes and extensions, but zipping extensions can take a very very long time.


common/src/warp_runner/manager/commands/other_commands.rs
```diff
@@ -0,0 +1,107 @@
...
+            let r = compress_folder(src, dest).await;
+            let _ = rsp.send(r);
+        }
+    }
+}
+
+async fn compress_folder(src: PathBuf, dest: PathBuf) -> Result<(), error::Error> {
+    let (tx, rx) = oneshot::channel::<Result<(), ZipError>>();
+    // I know that warp_runner is basically single threaded but still...put the blocking operation in a separate thread and await it
+    thread::spawn(move || {
```
dariusc93: you could technically get away with using `tokio::spawn_blocking` since it would use tokio own threadpool to spawn a thread for blocking operations but would allow it to yield back to the executor so there technically wouldnt be any need for the oneshot channel here.

phillsatellite (approved):

PR #430, opened 2023-03-10

dariusc93:
### What this PR does 📖

- Updates dependency
- Added enum to IdentityUpdate to allow clearing images (picture and banner) as well as status
- Fix tesseract clearing out store at random or initialization
- Added `MultiPassEventKind::IdentityUpdate`
- Added `experimental` to static arguments. 

- N/A

- Its recommended to create a new account with this update
- Test photos and banners (future PR to uplink may change how it refreshes so it relies on the emitted event rather than a minute refresh)
- Test with and without `--experimental-node` flag. See `Additional comments` for more info

### Additional comments 🎤
- This will not be compatible with previous commit.
- `MultiPassEventKind::IdentityUpdate` is only emitted from updates with friends:
  - `UpdateKind::StatusMessage` doesnt emit at this time
  - `UpdateKind::Username` will show both old and new names
  - `UpdateKind::Picture` and `UpdateKind::Banner` is emitted without any variant to the enum (though this, along with `UpdateKind::StatusMessage` may get collapsed into a single enum variant)
  - `UpdateKind::Status` updates when there is a change to `IdentityStatus`. Note: This doesnt emit when a user goes online or offline at this time, but may in the future.
- Internal changes are done related to discovery and identity updates.
- Using `experimental-node` flag will use a separate server for discovery as well as a relay with a higher transmission rate. For most networks and systems, direct connection may be done through the relay (so its not dependent on the relay directly), but some who systems arent able to establish a direct connection would still be able to take advantage of the relay. Note that this may change in the future so its not to be relied on all that much at this time but can be used while testing. Also note that there may be a much lower resource usage while using this flag due to it not being apart of the public nodes that is used by default.


phillsatellite:
Wanted to leave some notes, was just a draft but Darius wanted me to check this out and see how it performed. Was going good but I did realize PFP would not load after 60 seconds or restarting the app so Darius is going to take a look into that

sdwoodbury (approved):

phillsatellite (approved):

PR #440, opened 2023-03-11
fix(messages): prevent messages jumping around on re-render

### LINKED ISSUE #343: Chat - Messages jump around when resizing window
**Issue:** When you manually resize the app window while on the chat screen the messages seem to jump around a little 

**OS:**
-Windows
-MacOS

**Screen recording:**

https://user-images.githubusercontent.com/93608357/221326169-8f1bb0c0-c641-455f-bdaf-2f962d6cde87.mov

ealmloff:
### What this PR does 📖

- Fixes messages jumping around when they re-render. This was caused by scripts running every render instead of just the first render.
- Moves all scripts from script elements to use_eval calls inside use_effect hooks. 
- Fixes some Uuid's regenerating ids every render causing poor performance.

Resizing the window now runs less scripts:
Old:
image
New:
image

- Resolve #343

### DIFF
--- a/kit/src/components/context_menu/mod.rs
+++ b/kit/src/components/context_menu/mod.rs
@@ -4,7 +4,7 @@ use dioxus::{
     events::{MouseData, MouseEvent},
     prelude::*,
 };
-use dioxus_desktop::use_window;
+use dioxus_desktop::{use_eval, use_window};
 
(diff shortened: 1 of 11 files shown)


InfamousVague (approved):
looks great to me

PR #445, opened 2023-03-13
feat(chat): Disable various chat feature if not friend anymore

### LINKED ISSUE #421: Chat - Indicate that someone is no longer friends with you
**Issue:** Right now when someone removes you, there are times when only the removers side is updated and the other party can continue to send you messages even though you are no longer friends. We do not want to send the user a notification saying "User A removed you as a friend" but more like grey out the Chat input for the other party so that when you remove them they cannot still try to send you messages and put something like "both parties must be friends to message"

### LINKED ISSUE #458: Chat - Chat bar size will persist after sending a large message
**Issue:** When you are sending a large message, after you hit send the input box in chat will remain the same size as it was when you had the message typed up 

**Screen recording:**

https://user-images.githubusercontent.com/93608357/225359653-a60719b7-4688-4a04-8c0b-d2f319e392c6.mov

Flemmli97:
### What this PR does 📖

- Disables various chatting feature when one isn't a friend anymore:
      Chatting will be grayed out with a text with a tooltip saying that you are not friends anymore
      Buttons will be disabled (for e.g. attachment, sending msg, replies)
- Note: Only applies to DM (participants of 2) and not group chats. One can always do things in group chats.
  Maybe someone else has something to add here
- Also fixed a thing in `get_chatbar` where extensions where assigned to the control field and not the extension field see [https://github.com/Satellite-im/Uplink/blob/friend_chat_disable/ui/src/components/chat/compose.rs#L856](https://github.com/Satellite-im/Uplink/blob/friend_chat_disable/ui/src/components/chat/compose.rs#L856)
- Also fixes chatbar resizing after sending message not working (#458)
see [https://github.com/Satellite-im/Uplink/pull/445/files#diff-6c3d4f34c9eca5781fecc62f1c2baf341f5a244271e0e1a291709c19cfb41973R85](https://github.com/Satellite-im/Uplink/pull/445/files#diff-6c3d4f34c9eca5781fecc62f1c2baf341f5a244271e0e1a291709c19cfb41973R85)

- Resolve #421
- Also resolve #458

### DIFF
--- a/kit/src/elements/textarea/mod.rs
+++ b/kit/src/elements/textarea/mod.rs
@@ -7,6 +7,8 @@ use std::{cell::RefCell, rc::Rc};
 use dioxus::prelude::*;
 use dioxus_html::input_data::keyboard_types::{Code, Modifiers};
 
+use crate::elements::tooltip::{ArrowPosition, Tooltip};
+
 #[derive(Clone, Copy)]
 pub enum Size {
     Small,
@@ -41,6 +43,10 @@ pub struct Props<'a> {
     onchange: EventHandler<'a, (String, bool)>,
     onreturn: EventHandler<'a, (String, bool, Code)>,
     value: String,
+    #[props(default = false)]
+    is_disabled: bool,
+    #[props(default = "".to_owned())]
+    tooltip: String,
 }
 
 #[allow(non_snake_case)]
@@ -56,6 +62,8 @@ pub fn Input<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
         onchange,
         onreturn,
         value,
+        is_disabled,
+        tooltip,
     } = &cx.props;
 
     let height_script = include_str!("./update_input_height.js");
@@ -64,9 +72,10 @@ pub fn Input<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
     dioxus_desktop::use_eval(cx)(focus_script.to_string());
 
     let script = include_str!("./script.js")
-        .replace("UUID", id)
+        .replace("$UUID", id)
         .replace("$MULTI_LINE", &format!("{}", true));
     let current_val = value.to_string();
+    let disabled = *loading || *is_disabled;
 
     let cv2 = current_val.clone();
 
@@ -76,9 +85,9 @@ pub fn Input<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
 
     cx.render(rsx! (
         div {
-            class: format_args!("input-group {}", if *loading { "disabled" } else { " " }),
+            class: "input-group",
             div {
-                class: "input",
+                class: format_args!("input {}", if disabled { "disabled" } else { " " }),
                 height: "{size.get_height()}",
                 script { "{script}" },
                 textarea {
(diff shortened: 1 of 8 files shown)


phillsatellite:
Maybe @InfamousVague can help me test this more later this evening, -I had Windows and _MacbookPro_ added as friends with a chat
-Removed _MacbookPro_ from the _Windows_ account  
-MacbookPro_ side Windows was still a friend and able to message them 


**MacbbokPro**

https://user-images.githubusercontent.com/93608357/224847015-abe91119-d30d-412e-9d72-f40216cc3ef3.mov

**Windows**

https://user-images.githubusercontent.com/93608357/224847040-d2b284b1-3ff7-4ec0-9cd1-56751413fa1d.mp4

Flemmli97:
if i have the app open and someone unfriends me it works. in fact during that it also removes the chat.

 but if i am not online, someone unfriends me during that and i open the app again it still shows as friends.

edit: 
the message seems to be delayed as when i have the app open and wait a bit the message for "friend removal" comes in and removes the false friend

kit/src/elements/textarea/mod.rs
```diff
@@ -42,6 +44,10 @@ pub struct Props<'a> {
     onreturn: EventHandler<'a, (String, bool, Code)>,
     #[props(!optional)]
     reset: Option<UseState<bool>>,
+    #[props(optional)]
+    is_disabled: Option<bool>,
+    #[props(optional)]
+    tooltip: Option<String>,
```
sdwoodbury: the `#[props(optional)]` has no effect on a `Option`. I recommend not using it for this purpose. 

By default, making any Dioxus prop an `Option` means that if the field is omitted, it defaults to `None`. 
If for some reason you need to pass an actual `Option` to a prop, that macro can be helpful if used like this: `#[props(!optional)]` .

phillsatellite:
Tested this with Hogan today, but couldn't get the chat to persist after removing each other. As soon as one of us did the chat would be removed from the sidebar so although thats good and the correct behavior we wanted originally it has been making this a bit difficult to test. Going to try one more time this afternoon with @InfamousVague since he has this problem occur a lot

Flemmli97:
it is because in [https://github.com/Satellite-im/Uplink/blob/dev/common/src/state/mod.rs#L874](https://github.com/Satellite-im/Uplink/blob/dev/common/src/state/mod.rs#L874) it also removed the chat. what feature should be kept then? the removal or the disabling of chat?

phillsatellite:
Hi @Flemmli97 I was able to test with Matt and we could also not replicate the bug still but I asked Matt what he wanted to do and he suggested this, 

If the other person in chat removes you, we should just grey out the call buttons and input box and not remove the chat for now. Matt made a mockup to reference

image

Flemmli97:
ok. the chat now doesnt get removed when someone unfriends you.
it should still get removed though if you unfriend them yourself (think that makes sense also).

the chat was getting disabled when the friend status is updated correctly. but the problem prior was that warp has some kind of delay updating things when you were offline and stuff happened.

kit/src/elements/textarea/mod.rs
Flemmli97: oh ok.
i made to field an Option so it doesnt require major refactoring on the things calling the textarea.
will remove the `#[props(optional)]` though

Flemmli97: replaced Option with props(default...)

phillsatellite (approved):

phillsatellite (approved):

sdwoodbury:
review in progress

sdwoodbury:
This is the real diff
[changes.diff.txt](https://github.com/Satellite-im/Uplink/files/11052052/changes.diff.txt)

sdwoodbury (approved):

sdwoodbury:
This is the real diff. GitHub was erroneously showing 62 files changed. 
[changes.diff.txt](https://github.com/Satellite-im/Uplink/files/11052052/changes.diff.txt)

PR #454, opened 2023-03-14
chore(ui): suggest changes to AddFilePreview branch 

sdwoodbury:
### What this PR does 📖

- Essentially an `Option<Vec<Option<WindowId>>>` was added to `UI` for handling file previews. This was changed to a `Vec<WindowId>`. 
- The `WindowDropHandler` was moved into the utils folder and changed so that it can be reusable
- The `WindowDropHandler` was removed from `FilePreviewProps` because there isn't a way to tell which window id was created in time. This would have to happen before the `WindowDropHandler` was created. 

- Resolve #

### DIFF
--- a/common/src/state/action.rs
+++ b/common/src/state/action.rs
@@ -67,15 +67,17 @@ pub enum Action {
     AddOverlay(Weak<WebView>),
     /// used for the popout player or media player
     #[display(fmt = "SetPopout")]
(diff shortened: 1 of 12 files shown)


sdwoodbury:
i'm gonna try to make a way to remove a WindowId for a FilePreview from State

PR #460, opened 2023-03-15
feat(chat): lazy load messages

sdwoodbury:
### What this PR does 📖

- Only loads 10 messages for the active chat initially. then loads 10 more every 1/2 second. This allows the user to switch between conversations with almost zero lag. 
- The compose component is refactored to actually allow code maintenance. It's a big change but without this, the `rsx!` macro to render messages consists of what seems like 1000 lines of code and is nearly impossible to modify.  
- Some actions in State were refactored to require less cloning. This was accomplished by using references and replacing structs such as `Chat`, `Message`, and `Identity` with their `Uuid` or `DID` where appropriate. 

- Resolve #

### Additional comments 🎤
Some variables in the friends pages were change unnecessarily. This is because I thought I could use a single reference instead of cloning the variable X times (for the `move` closures). The IDE didn't show the error at first and then I had to go change it back.


ui/src/components/chat/compose.rs
```diff
@@ -448,166 +481,244 @@ fn get_messages(cx: Scope<ComposeProps>) -> Element {
...
+        active_chat_id: _,
+    } = cx.props;
+
+    let messages = &group.messages;
+    let last_message = messages.last().unwrap().message;
+    let sender = state.read().get_identity(&group.sender);
+    let sender_name = sender.username();
+    let active_language = &state.read().settings.language;
+
+    // todo: investigate why updating one's profile picture affects their status.
```
dariusc93: Is this still the case since #430 was merged?

phillsatellite (dismissed):

ui/src/components/chat/compose.rs
sdwoodbury: it doesn't look like it. I removed the comment.

common/src/state/action.rs
```diff
@@ -135,25 +127,19 @@ pub enum Action {
     RemoveFromSidebar(Uuid),
     /// Adds or removes a chat from the favorites page
     #[display(fmt = "ToggleFavorite")]
-    ToggleFavorite(Chat),
+    ToggleFavorite(&'a Uuid),
```
dariusc93, line 130: `Uuid` implements `Copy` so you could easily dereference it without needing to borrow it

common/src/state/action.rs
```diff
@@ -135,25 +127,19 @@ pub enum Action {
...
     #[display(fmt = "RemoveReaction")]
     RemoveReaction(Uuid, Uuid, String),
-    /// Reply to a given message by ID
-    #[display(fmt = "Reply")]
-    Reply(Chat, Message),
-    /// Prep the UI for a message reply.
+    /// chat id, message id
     #[display(fmt = "StartReplying")]
-    StartReplying(Chat, Message),
+    StartReplying(&'a Uuid, &'a ui_adapter::Message),
```
dariusc93, line 142: ditto for `Uuid`

dariusc93 (approved):
Looks good overall. I did leave a comment but isnt blocking to this PR.

Jekrimo (approved):

common/src/state/action.rs
sdwoodbury: thanks, I'll make these changes in the next iteration

PR #462, opened 2023-03-15
feat(settings): Add did + copy field to profile setting

### LINKED ISSUE #450: Settings - Add end of the DID into Settings/Profile
**Task:** We want to add the last of the users did into the Settings Profile page so they have a way of finding it themselves other than Copying/pasting it on the friends page. 

- It should be to the right of the Username input box 
- It should not be editable but you should be able to copy from it

**Screenshot:**

Flemmli97:
### What this PR does 📖

- Adds a DID field to profile setting
- Shows the last 5 chars of the key
- Clicking on button copies the key to clipboard

- Resolve #450

### DIFF
--- a/ui/src/components/settings/sub_pages/profile/mod.rs
+++ b/ui/src/components/settings/sub_pages/profile/mod.rs
@@ -149,9 +149,10 @@ pub fn ProfileSettings(cx: Scope) -> Element {
         special_chars: None,
     };
 
-    let did_string = state.read().did_key().to_string();
+    let did_string = state.read().get_own_identity().did_key().to_string();
+
     let mut did_short = "#".to_string();
-    did_short.push_str(&did_string[did_string.chars().count() - 5..]);
+    did_short.push_str(&state.read().get_own_identity().short_id());
 
     let change_banner_text = get_local_text("settings-profile.change-banner");
     cx.render(rsx!(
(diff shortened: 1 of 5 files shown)


ui/src/components/settings/sub_pages/profile/mod.rs
```diff
@@ -146,6 +149,10 @@ pub fn ProfileSettings(cx: Scope) -> Element {
         special_chars: None,
     };
 
+    let did_string = state.read().did_key().to_string();
+    let mut did_short = "#".to_string();
+    did_short.push_str(&did_string[did_string.chars().count() - 5..]);
```
dariusc93: if the state is `Identity`, you can extract the short id from it instead by doing `state.read().short_id()`. Do note this would yield the last 8 characters of the did key if you decided to do that.

dariusc93 (dismissed):
It looks good to me from a glance. Did leave a comment though but not exactly blocking to this PR

ui/src/components/settings/sub_pages/profile/mod.rs
InfamousVague: Let's go this path just to keep things standardized.

InfamousVague (changes requested):
I think after the change Darius suggested this should be good to go

phillsatellite (dismissed):

phillsatellite:
Was good on Windows/Mac/Linux & desktop/mobile once the conflicts are fixed good to go on my side 😄

phillsatellite (approved):

dariusc93 (approved):

sdwoodbury:
Matt is gone for 2 weeks and is the only person who can unblock this PR. let's just re-do the PR.

PR #463, opened 2023-03-15
fix(input): add num pad enter to keyup

### LINKED ISSUE #453: UI - User input on keypad enter does not send on any inputs
**Issue:** Keypad enter does not send message when in a chat with another user as well as other inputs across the app

Should just be a matter of adding the key event into the submit on inputs

**OS:**
-Windows
-MacOS
-Linux

**Screenshot:**
image

Jekrimo:
### What this PR does 📖

- fix numpad enter, just had to account for the code as expected

- Resolve #453


phillsatellite (dismissed):

phillsatellite:
Good across all OS's 💯

sdwoodbury (changes requested):
we should do this for `textarea::Input` also

Jekrimo:
> we should do this for `textarea::Input` also

good looking out, totally forgot about the textarea! updated

sdwoodbury (approved):

Issue #152, opened 2023-01-18
bug(macOS): Click + Drag for chat scrollbar not working


phillsatellite:
**Issue:** Like title says, on Windows you are able to click + drag the scrollbar to move up or down in chat, on MacOS this is not working 

**OS:**
-MacOS
-Linux

**Screen recording:**

https://user-images.githubusercontent.com/93608357/213250064-85561b5c-3007-4c16-9feb-06501dabb50a.mov

phillsatellite:
Re-tested and replicated on latest dev but only on MacOS

phillsatellite:
Still an issue on MacOS, keeping open

sdwoodbury:
likely the same issue as #732

phillsatellite:
Replicated on Dev branch and on PR #732

stavares843:
https://github.com/Satellite-im/Uplink/assets/29093946/bfc8e183-4388-4be4-914b-f462ef78e097

seems fixed, closing 🔨

PR #468, opened 2023-03-16
fix(sidebar): Fix missing scrollbar in sidebar

### LINKED ISSUE #422: Files - Sidebar jumps when moving from Chat to Files page
**task:** Because the files sidebar does not have the scroll implemented in it, it causes the sidebar to jump around a little when going into the files page. Adding scroll should fix this 

**Screen recording:**

https://user-images.githubusercontent.com/93608357/223850632-14d8cf5b-3179-48e0-9f71-525078c6cfd2.mov

Flemmli97:
### What this PR does 📖

- Fixes the scrollbar in for sidebar in files missing causing it to jump around when switching tabs
- Also disable scrollbar onhover increasing width as on hover seems unreliable. Instead scrollbar width in general increased a bit:

Before when you hover over any scrollbar it gets bigger. At least that should be the case but that seems really unreliable. I managed to trigger it reliably by middle and right clicking while hovering over it. Else it also flickers if you move the mouse across it quickly. (for demonstration purposes i set the width increase in the vid to 14px).

I removed this feature and instead increased the base size of it from 6->8px.

https://user-images.githubusercontent.com/34157027/225683131-a92541b3-032f-461d-a41c-de472d2906d7.mp4

- Resolve #422

### DIFF
--- a/ui/src/layouts/style.scss
+++ b/ui/src/layouts/style.scss
@@ -103,11 +103,12 @@
 
     }
     // hide scrollbar to prevent issues with 6 pixel scrollbar
-    ::-webkit-scrollbar {
+    .files-list::-webkit-scrollbar {
         display: none;
+        scrollbar-width: none;  /* Firefox */
     }
+
     -ms-overflow-style: none;  /* IE and Edge */
-    scrollbar-width: none;  /* Firefox */
 
     .files-list {
         display: flex;
--- a/ui/src/style.scss
+++ b/ui/src/style.scss
@@ -103,13 +103,9 @@ html, body {
 
 /* width */
 ::-webkit-scrollbar {
-    width: 6px;
-    height: 6px;
+    width: 8px;
+    height: 8px;
     opacity: 1;
-
-    &:hover {
-        width: 10px;
-    }
 }
 
 /* Track */


sdwoodbury (commented):
This definitely makes the scroller visible when I mouse over it. 

Perhaps this is a different issue, but if I see the scroller, I expect to be able to drag it. I wonder if showing the users the scroller would cause them to expect the same thing. Maybe we shouldn't show the scroller unless it can also be dragged.

phillsatellite (approved):

phillsatellite:
> This definitely makes the scroller visible when I mouse over it.
> 
> Perhaps this is a different issue, but if I see the scroller, I expect to be able to drag it. I wonder if showing the users the scroller would cause them to expect the same thing. Maybe we shouldn't show the scroller unless it can also be dragged.

Just wanted to mention the ticket for Dragging scrollbar not working #152 this problem also seems to be only in chat and  only on Linux and MacOS on Windows user can click and drag all scrollbars

Flemmli97:
> This definitely makes the scroller visible when I mouse over it.
> 
> Perhaps this is a different issue, but if I see the scroller, I expect to be able to drag it. I wonder if showing the users the scroller would cause them to expect the same thing. Maybe we shouldn't show the scroller unless it can also be dragged.

do you mean the scroller for the sidebar? if so it doesnt show for me. only if the elements in the sidebar become big enough that scroll is needed:


https://user-images.githubusercontent.com/34157027/225717572-0a0851f1-2cbe-4c60-8193-ea45cb8b0565.mp4

PR #470, opened 2023-03-16
fix(config): Remove additional entry in Cargo.toml

dariusc93:
### What this PR does 📖

- Removes additional entry in `Cargo.toml`


dariusc93:
Will merge this in due to it only removing a duplicate entry. No testing is needed

PR #472, opened 2023-03-17
fix(chat): Save the last active chat when switching tabs

### LINKED ISSUE #459: Chat - User has to click back into chat
**Issue:** Brought this up in dev chat and Stu was able to replicate as well. When you leave chat page and come back, you will have to click back into the chat again

**Screen recording:**

https://user-images.githubusercontent.com/93608357/225362757-bd2ded2c-8704-4223-8db0-857c9f1503b5.mov

Flemmli97:
### What this PR does 📖

- Saves the uuid of the active chat when one switches to another tabs so when it switches back it still shows the chat
- The uuid is saved in the chats struct of `chats.rs` as `last_active` so it also persists over saves

- Resolve #459

### DIFF
--- a/common/src/state/chats.rs
+++ b/common/src/state/chats.rs
@@ -63,6 +63,8 @@ pub struct Chats {
     pub in_sidebar: VecDeque<Uuid>,
     // Favorite Chats
     pub favorites: Vec<Uuid>,
+    // The last active chat when one switches tab
+    pub last_active: Option<Uuid>,
 }
 
 impl Chats {
@@ -105,6 +107,7 @@ impl Serialize for Chats {
         state.skip_field("active_media")?;
         state.serialize_field("in_sidebar", &self.in_sidebar)?;
         state.serialize_field("favorites", &self.favorites)?;
+        state.serialize_field("last_active", &self.last_active)?;
 
         state.end()
     }
--- a/common/src/state/mod.rs
+++ b/common/src/state/mod.rs
@@ -639,6 +639,9 @@ impl State {
     }
     /// Clears the active chat in the `State` struct.
     fn clear_active_chat(&mut self) {
+        if let Some(c) = self.chats.active {
+            self.chats.last_active = Some(c);
+        }
         self.chats.active = None;
     }
     pub fn clear_typing_indicator(&mut self, instant: Instant) -> bool {
--- a/common/src/testing/mock.rs
+++ b/common/src/testing/mock.rs
@@ -80,6 +80,7 @@ pub fn generate_mock() -> State {
         active_media: None,
         in_sidebar,
         favorites: vec![],
+        last_active: None,
     };
     let friends = Friends {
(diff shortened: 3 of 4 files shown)


phillsatellite (approved):

sdwoodbury (changes requested):
Adding an extra variable to state is unnecessary. The problem is here
image

This was introduced by commit `0cc5161402543020d4d6833d5fbb1790fec66767`. All we have to do is revert that change. `Action::ClearChat` shouldn't be called when one leaves the Chats view.

Flemmli97:
> Adding an extra variable to state is unnecessary. The problem is here image
> 
> This was introduced by commit `0cc5161402543020d4d6833d5fbb1790fec66767`. All we have to do is revert that change. `Action::ClearChat` shouldn't be called when one leaves the Chats view.

i added an extra state since i wasnt sure if removing that line causes other issues as then the active chat will be always present. even if e.g. you are in another tab

edit: the linked commit does indeed depend on it.

sdwoodbury:
> > Adding an extra variable to state is unnecessary. The problem is here image
> > This was introduced by commit `0cc5161402543020d4d6833d5fbb1790fec66767`. All we have to do is revert that change. `Action::ClearChat` shouldn't be called when one leaves the Chats view.
> 
> i added an extra state since i wasnt sure if removing that line causes other issues as then the active chat will be always present. even if e.g. you are in another tab
> 
> edit: the linked commit does indeed depend on it.

[This](https://github.com/Satellite-im/Uplink/blob/dev/common/src/state/ui.rs#L58) can probably be used instead. If the current layout isn't `Compose` then a notification should be played.

Flemmli97:
> > > Adding an extra variable to state is unnecessary. The problem is here image
> > > This was introduced by commit `0cc5161402543020d4d6833d5fbb1790fec66767`. All we have to do is revert that change. `Action::ClearChat` shouldn't be called when one leaves the Chats view.
> > 
> > 
> > i added an extra state since i wasnt sure if removing that line causes other issues as then the active chat will be always present. even if e.g. you are in another tab
> > edit: the linked commit does indeed depend on it.
> 
> [This](https://github.com/Satellite-im/Uplink/blob/dev/common/src/state/ui.rs#L58) can probably be used instead. If the current layout isn't `Compose` then a notification should be played.

A lot of places seem to assume active chat = chat that is currently open. Some examples are e.g. 
in `common::src::state::chats.rs#active_chat_has_unreads` or `#get_replying_to` etc. The comment for the field itself says this too. So changing this implication seems to break quite a bit of stuff.

sdwoodbury:
I maintain that `State` doesn't need any new variables. I believe that fixing notifications could have been a 1 line change (before the bug was introduced). I submitted an alternative PR here: https://github.com/Satellite-im/Uplink/pull/481.

sdwoodbury:
closed in favor of #481

PR #484, opened 2023-03-21
feat(chat): File drag & drop onto chat

### LINKED ISSUE #424: Chat - Add drag and drop to messages
**Task: ** Here we want to implement the ability to drag and drop files into the active chat you are on. Similar to how Lucas added the ability to drag and drop onto the files page

### LINKED ISSUE #452: Files - Drag & drop not working on Linux
**Issue:** Was replicated on Linux Ubuntu, with the latest dev branch I wasnt able to get drag/drop files to work

**Screen recording:**

https://user-images.githubusercontent.com/93608357/224859836-474e9f4c-d97b-4c8c-ad02-099ae92fabca.mov

Flemmli97:
### What this PR does 📖

- Adds drag and dropping files onto chat to send them
- This also fixes an issue with file paths being html encoded and causing file not found errors on linux.
  -> Whitespaces are encoded as `%20`
  Other os seem to be fine regarding this

- Resolve #424
- Resolve #452


ui/src/components/chat/overlay.js
```diff
@@ -0,0 +1,13 @@
+var filesLayout = document.getElementById('$ELEMENTID')
```
lgmarchi: My friend, I think we can change this name to chatLayout

ui/src/layouts/storage.rs
```diff
@@ -844,3 +858,8 @@ async fn drag_and_drop_function(
         tokio::time::sleep(std::time::Duration::from_millis(100)).await;
     }
 }
+
+// Decodes the given path from an html encoded path
+pub fn decoded_path_string(path: &Path) -> String {
```
lgmarchi: Is it not better to add this code just for linux?

lgmarchi (approved):
For me this PR is very good. 

Just let a suggestion and a question, but non block

sdwoodbury:
Please make the Cargo Workflow pass before merging. It just wants you to satisfy `cargo clippy` and `cargo fmt`. 
image

phillsatellite:
Maybe this is an issue specifically with my VM?  Whenever I drag+drop a file in Linux it says 
`Drag Event: Cancelled`

Another thing I noticed that would be cool to get fixed in here as well, it looks like when you drop an attachment and do not type any text and try to use "Enter Key" to send, it will not. The user is forced to click the send button. This isnt caused by the PR it broke on dev as well but I think it fits the scope of this and Stu had mentioned its pretty easy to fix 

Could anyone else verify if their linux drag/drop works fine?

https://user-images.githubusercontent.com/93608357/226684754-a75420ef-723c-4477-98b7-1f25d9810784.mov

Flemmli97:
> Maybe this is an issue specifically with my VM? Whenever I drag+drop a file in Linux it says `Drag Event: Cancelled`
> 
> Another thing I noticed that would be cool to get fixed in here as well, it looks like when you drop an attachment and do not type any text and try to use "Enter Key" to send, it will not. The user is forced to click the send button. This isnt caused by the PR it broke on dev as well but I think it fits the scope of this and Stu had mentioned its pretty easy to fix
> 
> Could anyone else verify if their linux drag/drop works fine?

using linux it works on my machine.

the textarea decides what input is valid or not (which is currently that the input it not empty) would it make more sense instead if the component that handles the input (not the textarea) should decide whether something is valid or not?

ui/src/components/chat/overlay.js
Flemmli97: changed some names

sdwoodbury:
> > Maybe this is an issue specifically with my VM? Whenever I drag+drop a file in Linux it says `Drag Event: Cancelled`
> > Another thing I noticed that would be cool to get fixed in here as well, it looks like when you drop an attachment and do not type any text and try to use "Enter Key" to send, it will not. The user is forced to click the send button. This isnt caused by the PR it broke on dev as well but I think it fits the scope of this and Stu had mentioned its pretty easy to fix
> > Could anyone else verify if their linux drag/drop works fine?
> 
> using linux it works on my machine.
> 
> the textarea decides what input is valid or not (which is currently that the input it not empty) would it make more sense instead if the component that handles the input (not the textarea) should decide whether something is valid or not?

maybe we should just go back to having `textarea::Input` say that any call to `onreturn` has valid input, except when triggered by `onblur`. Then when sending a message, only let an empty message be sent if there's an attachment.

phillsatellite:
Had a small issue when pulling the latest changes, on all OS's 

Screenshot 2023-03-21 at 2 27 28 PM

ui/src/layouts/storage.rs
```diff
@@ -831,7 +832,20 @@ async fn drag_and_drop_function(
                 window.eval(&script);
             }
             FileDropEvent::Dropped(files_local_path) => {
+                #[cfg(not(target_os = "linux"))]
                 ch.send(ChanCmd::UploadFiles(files_local_path));
+                #[cfg(target_os = "linux")]
```
dariusc93: Instead of a double feature gate here, what could be done is is to use the `cfg!` macro inside `decoded_path_string` to either replace or passthrough the content on other platforms.

ui/src/layouts/storage.rs
```diff
@@ -844,3 +858,9 @@ async fn drag_and_drop_function(
         tokio::time::sleep(std::time::Duration::from_millis(100)).await;
     }
 }
+
+// Decodes the given path from an html encoded path
+#[cfg(target_os = "linux")]
+pub fn decoded_path_string(path: &Path) -> String {
```
dariusc93: Maybe it would be best to return a `PathBuf` here instead.

dariusc93 (commented):
LGTM but I did leave some suggestions to reduce dup code. Overall, it might be worth looking into why its escaping the spaces in the path too so we could make sure other areas would not be affected too.

ui/src/layouts/storage.rs
Flemmli97: now gets passed the vec<pathbuf> and returns that

ui/src/layouts/storage.rs
```diff
@@ -844,3 +846,20 @@ async fn drag_and_drop_function(
...
+
+pub fn decoded_pathbufs(paths: Vec<PathBuf>) -> Vec<PathBuf> {
+    let mut paths = paths;
+    #[cfg(target_os = "linux")]
+    {
+        let decode = |path: &Path| path.as_os_str().to_string_lossy().replace("%20", " ");
+        paths = paths
+            .iter()
+            .map(|p| {
+                let mut path = PathBuf::new();
```
dariusc93:
```suggestion
```

ui/src/layouts/storage.rs
```diff
@@ -844,3 +846,20 @@ async fn drag_and_drop_function(
...
+                path.push(decode(p));
```
dariusc93:
```suggestion
                PathBuf::from(decode(p))
```

ui/src/layouts/storage.rs
```diff
@@ -844,3 +846,20 @@ async fn drag_and_drop_function(
...
+                path
```
dariusc93:
```suggestion
```

dariusc93 (approved):
Left a quick suggestion (suggest confirming it in the IDE) but besides that LGTM.

phillsatellite (approved):

phillsatellite:
I couldnt get this to work with my parallels VM but I had Hogan verify for me that it was working with his Linux. The only issue he had was he couldnt use the enter key to send the message after attaching a file. Going to create ticket for it and approve this

PR #501, opened 2023-03-23
chore(misc): Refresh identity from request

### LINKED ISSUE #464: Ul - Usernames appear as DID key
**Issue:** When you add someone, sometimes in Pending/All Friends/or chat&sidebar they will appear as a blank profile picture with their DID appearing as they're name. This flow below will get the app to update the UI to the correct information 
 
1. User A adds User B 
2. User B appears as "z6M...Zcd" 
3. User B updates either their pfp, username, or status 
4. After this all of Users B information appears fine on Users A side

**OS**
-MacOS
-Windows
-Linux

**Screenshot:**
Screenshot 2023-03-15 at 5 35 32 PM

dariusc93:
### What this PR does 📖

- Refreshes identity from incoming and outgoing request

- Resolve #464 

### Additional comments 🎤
Due to uplink storing the identity internally, it does not call `MultiPass::get_identity` unless there is some trigger that warrant for calling it. As a result, it leaves some identities unresolved with the did-like name, specifically in the pending outgoing (or in some cases, incoming too). This PR will reuse the refresh command, which is repurposed to refresh identities of incoming and outgoing friend request every 5 seconds.


sdwoodbury (approved):

dariusc93:
With the changes done in the recent PR, this should no longer be needed.

PR #502, opened 2023-03-23
feat(friends): Add friend using name & hash

Flemmli97:
### What this PR does 📖

- Currently its only possible to add friends via their DID. This makes it possible to add friends also via their username + hash (which currently is the last 8 chars of their DID).
  With this change it first tries to parse the input as an DID and on success uses the current method. Else it will try to search for an identity using the input as username + hash.
  Some things to note:
    - If the other person is offline its not possible to use this method since there is no way currently to obtain an identity this way. This is more like the fault of warp though and darius mentioned that this might change in the future making it possible to query it this way. Assuming the API doesn't change then UI would require no further changes too.


phillsatellite (approved):

common/src/warp_runner/manager/commands/multipass_commands.rs
```diff
@@ -118,9 +121,30 @@ pub async fn handle_multipass_cmd(cmd: MultiPassCmd, warp: &mut super::super::Wa
...
-            let _ = rsp.send(r);
+        MultiPassCmd::RequestFriend { id, rsp } => {
+            // First attempt using a did
+            if let Ok(some) = DID::from_str(id.as_str()) {
+                let r = warp.multipass.send_request(&some).await;
+                let _ = rsp.send(r);
+                return;
+            }
+            // Invalid did. Try using it as a username (+ short did)
+            let r = warp.multipass.get_identity(Identifier::Username(id)).await;
```
dariusc93: You may want to add something to confirm that there is a short id attached because using `Identifier::Username` will accept with or without but if its with the short id, it would single out to one item in the array rather than multiple if you were to use it without and there is no short id.

common/src/warp_runner/manager/commands/multipass_commands.rs
```diff
@@ -118,9 +121,30 @@ pub async fn handle_multipass_cmd(cmd: MultiPassCmd, warp: &mut super::super::Wa
         MultiPassCmd::CreateIdentity { .. } | MultiPassCmd::TryLogIn { .. } => {
             // do nothing and drop the rsp channel
         }
-        MultiPassCmd::RequestFriend { did, rsp } => {
-            let r = warp.multipass.send_request(&did).await;
-            let _ = rsp.send(r);
+        MultiPassCmd::RequestFriend { id, rsp } => {
+            // First attempt using a did
+            if let Ok(some) = DID::from_str(id.as_str()) {
```
dariusc93: You could do a match condition to try for the did and if it errors to then proceed to attempt to fetch the identity by the username, grab the did get and then pass it along to `MultiPass::send_request`

eg

```rust
let did = match DID::from_str(&id) {
  Ok(did) => did,
  _ => {
    match warp.multipass.get_identity(Identifier::Username(id)).await.ok().and_then(list| list.first().cloned().map(|i|  i.did_key()) {
      Some(did) => did,
      _ => { 
          let _ = rsp.send(Err(Error::IdentityInvalid));
          return;
      }
    }
  }
}

let r = warp.multipass.send_request(&did).await;
let _ = rsp.send(r);
return;
```

dariusc93 (commented):
i did leave some comments but overall its a good start, but the problem is that if there is a mistake in the placement of the did key, it would automatically attempt to use it as a username, which may not be the correct way to go about it. It might be best to either use a separate command to possibly get the identity and pass the did along to send the request, use something like `Either` from the `either` crate and use some logic before hand to identify if its a possible username or did key (eg it may be a did if it begins with `did:key`, its a specific/fixed length, while username in uplink shouldnt exceed 32 characters, etc).

common/src/warp_runner/manager/commands/multipass_commands.rs
Flemmli97: > i did leave some comments but overall its a good start, but the problem is that if there is a mistake in the placement of the did key, it would automatically attempt to use it as a username, which may not be the correct way to go about it. It might be best to either use a separate command to possibly get the identity and pass the did along to send the request, use something like `Either` from the `either` crate and use some logic before hand to identify if its a possible username or did key (eg it may be a did if it begins with `did:key`, its a specific/fixed length, while username in uplink shouldnt exceed 32 characters, etc).

from what im seeing username can go anywhere from 4-64 chars so the only way would be to check if it starts with `did:key`

Flemmli97:
> i did leave some comments but overall its a good start, but the problem is that if there is a mistake in the placement of the did key, it would automatically attempt to use it as a username, which may not be the correct way to go about it. It might be best to either use a separate command to possibly get the identity and pass the did along to send the request, use something like `Either` from the `either` crate and use some logic before hand to identify if its a possible username or did key (eg it may be a did if it begins with `did:key`, its a specific/fixed length, while username in uplink shouldnt exceed 32 characters, etc).

from what im seeing is that username can go anywhere from 4 to 64 chars though. the only way then would be to check if the input starts with `did:key`

common/src/warp_runner/manager/commands/multipass_commands.rs
Flemmli97: added something that 
1. checks if the input contains a hash symbol
2. if the second part matches the length of a short id

dariusc93:
> from what im seeing is that username can go anywhere from 4 to 64 chars though. the only way then would be to check if the input starts with `did:key`

While its true that warp supports up to 64 characters (which is left that way there as a buffer in warp), uplink supports up to 32 characters. https://github.com/Satellite-im/Uplink/blob/99e84804f3b35aeb468d99dee68077428c471879/ui/src/components/settings/sub_pages/profile/mod.rs#L120-L134

common/locales/en-US/main.ftl
```diff
@@ -60,7 +60,8 @@ friends = Friends
     .requested = Requested 
     .sent = Sent
     .requests = Requests
-    .add-failed = You are already friends!
+    .add-existing-friend = You are already friends!
+    .add-failed = No such user exist!
```
sdwoodbury: I think this message (on line 64) should indicate that the user wasn't found, not that they don't exist.
If the user is offline, this message may appear for a user which does exist.

sdwoodbury: Also, the `Cargo Workflow` failed

Flemmli97: > > from what im seeing is that username can go anywhere from 4 to 64 chars though. the only way then would be to check if the input starts with `did:key`
> 
> While its true that warp supports up to 64 characters (which is left that way there as a buffer in warp), uplink supports up to 32 characters.
> 
> https://github.com/Satellite-im/Uplink/blob/99e84804f3b35aeb468d99dee68077428c471879/ui/src/components/settings/sub_pages/profile/mod.rs#L120-L134

done. i wouldve made the length check use constants but i didn't find an appropriate place to put those constant that made sense for both warp runner and ui.
changed ui values to use constants though

Flemmli97:
> > from what im seeing is that username can go anywhere from 4 to 64 chars though. the only way then would be to check if the input starts with `did:key`
> 
> While its true that warp supports up to 64 characters (which is left that way there as a buffer in warp), uplink supports up to 32 characters.
> 
> https://github.com/Satellite-im/Uplink/blob/99e84804f3b35aeb468d99dee68077428c471879/ui/src/components/settings/sub_pages/profile/mod.rs#L120-L134

done. i wouldve made the length check use constants but i didn't find an appropriate place to put those constant that made sense for both warp runner and ui.
changed ui values to use constants though

sdwoodbury (approved):

PR #526, opened 2023-03-27
feat(OpenDyslexic): Adds an option to toggle support for open dyslexic font faces

InfamousVague:
### What this PR does 📖

image

image

image


InfamousVague:
I also included a few CSS changes to unify things a bit.

sdwoodbury:
review in progress

common/src/state/action.rs
```diff
@@ -29,6 +29,9 @@ pub enum Action<'a> {
     // UI
     #[display(fmt = "WindowMeta")]
     SetMeta(WindowMeta),
+    #[display(fmt = "SetDyslexicEnabled")]
+    SetDyslexicEnabled(bool),
+
```
sdwoodbury: This should be deleted. Everywhere `ConfigAction::SetDyslexicEnabled` is used. There is no need to also have an `Action::SetDyslexicEnabled`

common/src/state/mod.rs
```diff
@@ -229,6 +229,9 @@ impl State {
 
             // ===== Configuration =====
             Action::Config(action) => self.configuration.mutate(action),
+            Action::SetDyslexicEnabled(flag) => self
```
sdwoodbury: This should be deleted and instead, `Action::Config` should be used.

sdwoodbury: you can see an example on any of the settings pages. ex: `state.write().mutate(Action::Config(ConfigAction::SetInterfaceSoundsEnabled(e)));`

ui/src/main.rs
```diff
@@ -65,6 +65,8 @@ mod overlay;
 mod utils;
 mod window_manager;
 
+pub static OPEN_DYSLEXIC: &'static str = include_str!("./open-dyslexic.css");
```
sdwoodbury: please fix all clippy warnings so that the `Cargo Workflow` passes. 
image

sdwoodbury (changes requested):

InfamousVague:
Also went ahead and removed the un-used splash screen setting. Additionally an accessibility page was created in settings for this feature as i'm sure we'll have other accessibility settings in the future.

phillsatellite:
So I was able to build fine, and I noticed the text glitch a little when toggling on and off but font never actually changed 

https://user-images.githubusercontent.com/93608357/228009370-105fa3ff-7ebd-409e-85eb-46e5c3222cc6.mov

native_extensions/emoji_selector/src/lib.rs
```diff
@@ -125,7 +125,7 @@ impl EmojiSelector {
             var emoji_selector = document.getElementById('emoji_selector');
             emoji_selector.focus();
         "#;
-        let eval = use_eval(cx);
+        let _eval = use_eval(cx);
```
XileHorizon: The _ in "_eval" breaks it. (Darius figured that out)
image

XileHorizon (commented):
The font did not change when toggled.

phillsatellite:
Seems good on my end, I noticed the font on the Accessibility page does not change (should it?), but to not block this I can create another ticket 

https://user-images.githubusercontent.com/93608357/228093967-abe276dc-b965-4a53-a5ea-6c7a5f40611e.mov

phillsatellite (approved):

common/locales/en-US/main.ftl
```diff
@@ -120,15 +121,15 @@ settings-profile = Profile Settings
...
-    .splash-screen-description = Disabling the splash screen could speed up load times.
     .app-language = App Language
     .change-language = Change Language
     .theme = Theme
     .theme-description = Change the theme of the app.
     .theme-reset = Reset Theme
     .theme-reset-description = Reset the theme to the default.
     .theme-reset-cta = Clear Theme
+    .dyslexia = Open Dyslexic
+    .dyslexia-description = Open Dyslexic may help some users who suffer from dyslexia, it's a custom font you can enable.
```
lgmarchi, line 133: Did you add translates for other languages we have in Uplink?

InfamousVague: I did not since I don't know the translations, I think we should instead of adding them to every file, we should fall back to the english version of the translation (I think fluent already supports this)

Issue #521, opened 2023-03-24
add ability to submit bug report


sdwoodbury:
needs a ui to do the following:
- attach a screenshot and description 
- attach a crash report

needs an API to send the report to

InfamousVague:
one thing we need to be very careful of here is to not expose any important personal information (or personal information about the users account at all, really) when capturing the reports. This will make things a lot safer while we're still pre code audit as well.

sdwoodbury:
sounds like we could have a link to an [upvoty](https://www.upvoty.com/) page where users could provide feedback

PR #534, opened 2023-03-27
feat(about): add a simple about page and add crash reporting

### LINKED ISSUE #520: add an about page

sdwoodbury:
### What this PR does 📖

- Add a simple about page, so users can know what version of the app they are running
- Give users the option to, in the event of a crash, save some logs on their machine, which they could then choose to send to us. 

- Resolve #520 

### Additional comments 🎤
related to #521 but isn't enough to close that ticket.


phillsatellite:
This is supppppperrrr minor but just to keep all the Settings pages matching is there anyway we could move the content to sit underneath of "About"

Appears correctly on Mobile view 🙂

Screenshot 2023-03-27 at 2 59 08 PM
Screenshot 2023-03-27 at 2 59 03 PM

stavares843:
Agree with Phil as well 😂

dariusc93:
Have you simulated a crash (eg include a panic, possibly in the main thread) to see what the report would look like?

sdwoodbury:
> Have you simulated a crash (eg include a panic, possibly in the main thread) to see what the report would look like?

yes. I could attach a sample when I get around to working on this ticket again. I just put a `todo!()` in one of the init functions in `main.rs`

dariusc93 (approved):

ui/src/main.rs
```diff
@@ -144,6 +147,40 @@ fn main() {
...
+            None => "panic occurred".into(),
+        };
+        let location = match panic_info.location() {
+            Some(loc) => format!(" at file {}, line {}", loc.file(), loc.line()),
+            None => "".into(),
+        };
+
+        let logs = logger::dump_logs();
+        let crash_report = format!("{intro}{location}\n{logs}\n");
+        println!("{crash_report}");
```
lgmarchi, line 169: Should this print be removed and changed to log?

lgmarchi (approved):

ui/src/components/settings/styles.scss
```diff
@@ -51,6 +52,13 @@
         flex-direction: column;
         gap: var(--gap-less);
     }
+    .settings-container {
+        flex: 1;
+        min-width: 0;
+        display: inline-flex;
+        flex-direction: row;
+        gap: var(--gap-less);
+    }
```
InfamousVague: instead of creating this, we should reuse the existing styles for the settings stuff if possible

ui/src/components/settings/sub_pages/about.rs
```diff
@@ -0,0 +1,48 @@
...
+
+#[allow(non_snake_case)]
+pub fn AboutPage(cx: Scope) -> Element {
+    let version = env!("CARGO_PKG_VERSION");
+    // todo: give the executable a better name than "ui"
+    let app_name = env!("CARGO_PKG_NAME");
+    cx.render(rsx!(
+        div {
+            id: "settings-about",
+            SettingContainer {
```
InfamousVague: maybe we can use settings section here instead

InfamousVague (changes requested):

ui/src/main.rs
sdwoodbury: no. the crash report should print to the terminal regardless of logging.

ui/src/components/settings/styles.scss
sdwoodbury: it would be easier if there were CSS classes for flexboxes so that I could make something display as flex-column without having to make a custom class for the element. Should I make that? Besides that, I don't see a better option than what's currently there.

sdwoodbury: done

ui/src/components/settings/sub_pages/about.rs
sdwoodbury: i did that where possible and changed SettingsContainer to re-use all the styling from SettingsSection

InfamousVague:
Made a few small changes in the incoming commit, switched to using the SettingSection component to avoid needing a new component, also moved the open codebase button from developer settings into the new about page and updated existing translation locations to be under the about page as well. Lastly I switched to using the "open" crate to open the url in the default browser without needing to style up an anchor tag.

InfamousVague:
image

InfamousVague:
> Give users the option to, in the event of a crash, save some logs on their machine, which they could then choose to send to us. 

I don't think this should be a problem, but it's worth making sure the crash reports can't leak things like state or other secure data. And if it can, warn the user about this possibility. 

We'll want to avoid things like https://abawazeeer.medium.com/using-mimikatz-to-get-cleartext-password-from-offline-memory-dump-76ed09fd3330 being used in conjunction with potential exploits that purposefully crash the application at a given moment in order to plaintext things on disk in the crash dump. 

For now though until our code is audited, it's probably enough to just keep this in the back of our minds.

sdwoodbury:
> > Give users the option to, in the event of a crash, save some logs on their machine, which they could then choose to send to us.
> 
> I don't think this should be a problem, but it's worth making sure the crash reports can't leak things like state or other secure data. And if it can, warn the user about this possibility.
> 
> We'll want to avoid things like https://abawazeeer.medium.com/using-mimikatz-to-get-cleartext-password-from-offline-memory-dump-76ed09fd3330 being used in conjunction with potential exploits that purposefully crash the application at a given moment in order to plaintext things on disk in the crash dump.
> 
> For now though until our code is audited, it's probably enough to just keep this in the back of our minds.

that's why I don't log messages or passwords. we might want to not log the DID or chat UUID either

phillsatellite:
I think this is good to go, links work on Windows/Mac/Linux but my other one suggestion would be to capitalize Uplink

Screenshot 2023-03-31 at 12 37 05 PM

InfamousVague (approved):

PR #540, opened 2023-03-28
fix(username): profile username number types mixed up

### LINKED ISSUE #536: Settings - PIN rules being applied to Username
**Issue:** When you edit your username, the rules for PIN are being applied to username so User is unable to create a username that is longer than 4 chars 

**OS:**
Mac
Windows
Linux

**Screen recording:**

https://user-images.githubusercontent.com/93608357/228024623-01a8f08f-3b11-4807-8a53-e3ae5661a214.mov

Jekrimo:
### What this PR does 📖

- import of number limit wasn't working, switched to basic number

- Resolve #536

### DIFF
--- a/ui/src/components/settings/sub_pages/profile/mod.rs
+++ b/ui/src/components/settings/sub_pages/profile/mod.rs
@@ -19,8 +19,6 @@ use rfd::FileDialog;
 use warp::multipass;
 use warp::{error::Error, logging::tracing::log};
 
-use crate::layouts::create_account::{MAX_USERNAME_LEN, MIN_USERNAME_LEN};
-
 #[derive(Clone)]
 enum ChanCmd {
     Profile(String),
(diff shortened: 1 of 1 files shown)


phillsatellite (approved):

ui/src/components/settings/sub_pages/profile/mod.rs
```diff
@@ -121,9 +119,9 @@ pub fn ProfileSettings(cx: Scope) -> Element {
     // Set up validation options for the input field
     let username_validation_options = Validation {
         // The input should have a maximum length of 32
-        max_length: Some(MIN_USERNAME_LEN),
+        max_length: Some(32),
```
lgmarchi, line 122: Here is not better to use variables as before, but just changing them? Because they are inverted

dariusc93 (approved):

PR #564, opened 2023-03-30

dariusc93:
### What this PR does 📖

- Resolved duplication error due to changes made upstream, which caused issue with sending/receiving request.
- Add Cargo.lock

- N/A

### Additional comments 🎤

Due to `Cargo.lock` being added, it would be suggested to reduce the need of doing `cargo update` whenever possible, and only update it in a PR if there is a change to any dependency (eg adding, removing, or updating a crate).


phillsatellite (approved):

phillsatellite:
When I first logged in I had x2 themes in the dropdown and crashed when I tried to select one. After I cleared my cache I couldnt replicate that again and it only happened on MacOS. Other than that I could add/remove/message friends, update names and pro pic again 🌟

InfamousVague (approved):

PR #565, opened 2023-03-31
fix(compose): fix compose view

sdwoodbury:
### What this PR does 📖

- the "chat draft" thing in State caused lots of re-renders and we finally decided to do it better. 
- now typing in the chatbar only re-renders the chatbar instead of the entire app. 
- I tried Fermi but Fermi doesn't support `make_mut` yet and I didn't want to clone a `HashMap` every time a chat draft was updated. 
- I was able to make a second `UseSharedState` just for chat drafts. I tested it with the emoji extension too, and it worked. 

- Resolve #


InfamousVague:
I think we might want to go about it another way but want to make sure I understand everything before going ahead and closing it; waiting to chat with Stuart tomorrow before going one way or the other, though.

PR #570, opened 2023-03-31
feat(installer): start on a windows installer

sdwoodbury:
### What this PR does 📖

- creates a windows installer using a GitHub action. (releasing it to GitHub doesn't quite work yet)
- bundles assets with the installer
    - creates a zip file using a combination `build.rs` 
    - `ui/wix/main.wxs` expects the zip file to be at a certain location
    - when `uplink` runs, it unpacks the assets into `.uplink`
    - `.uplink` has a separate directory for user created themes and fonts
- if compiled in release mode, unless an environment variable named `DEBUG` is found, uplink will not use the assets from the source code but instead search for the zip file. 

- Resolve #


phillsatellite (approved):

ui/src/utils/mod.rs
```diff
@@ -16,26 +16,31 @@ pub mod lifecycle;
...
-                styles,
-            };
-
-            themes.push(theme);
+    let mut add_to_themes = |themes_path| {
+        for file in WalkDir::new(themes_path)
+            .into_iter()
+            .filter_map(|file| file.ok())
+        {
+            if file.metadata().unwrap().is_file() {
```
dariusc93, line 24: Even though this been here prior to the PR, best to be cautious about unwrapping

dariusc93 (approved):
Look good to me from a glance.

WanderingHogan (approved):

PR #584, opened 2023-04-04
update(StateManagement): Update just chatbar when select emoji

lgmarchi:
### 1. Update just chatbar instead of entire compose when select an emoji 

https://user-images.githubusercontent.com/63157656/229807756-cb00d428-e154-4038-85fb-ef5494d7bd9c.mov

### 2. Old code as reference 

https://user-images.githubusercontent.com/63157656/229807831-fd848939-a328-401c-84c5-74002eaad9d7.mov

- 

- Resolve #

### DIFF
--- a/common/src/state/scope_ids.rs
+++ b/common/src/state/scope_ids.rs
@@ -4,7 +4,7 @@ use serde::{Deserialize, Serialize};
 #[derive(Clone, Default, Debug, Deserialize, Serialize)]
 pub struct ScopeIds {
     #[serde(skip)]
-    pub chatbar: usize,
+    pub chatbar: Option<usize>,
 }
 
 impl ScopeIds {
--- a/common/src/state/scope_ids.rs
+++ b/common/src/state/scope_ids.rs
@@ -1,0 +1,14 @@
+use dioxus::prelude::ScopeId;
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Default, Debug, Deserialize, Serialize)]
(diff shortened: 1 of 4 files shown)


common/src/state/scope_ids.rs
```diff
@@ -0,0 +1,14 @@
+use dioxus::prelude::ScopeId;
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Default, Debug, Deserialize, Serialize)]
+pub struct ScopeIds {
+    #[serde(skip)]
+    pub chatbar: usize,
```
sdwoodbury: It might be safer to make this an `Option<usize>` to handle the case where the scope Id hasn't been initialized yet.

lgmarchi: Good point my friend, I was thinking on this as well! Updated!

sdwoodbury (approved):

dariusc93 (approved):
Looks good! 

Honestly, I thought the chatbar was rendered by itself when it updates and not with the whole compose component but looking at the changes it does make more sense.

phillsatellite:
Hello my friend! When tested on Linux/Mac/Windows when user selects emoji app will panic and crash 


I verified on all OS's on dev branch that this does not happen

https://user-images.githubusercontent.com/93608357/229850268-c7fbe68f-eb12-4880-9b29-25e46af20b4f.mov

phillsatellite (approved):

PR #595, opened 2023-04-05
feat(updates): add workflow for linux and code to check GitHub API

sdwoodbury:
### What this PR does 📖

- adds GitHub workflow to build and release .deb files
- releases binaries for all 3 platforms, in addition to the installers (this may not be needed)
- adds code to check for updates
- adds code to download the update to the location specified by the user
- adds similar functionality to the About page

- Resolve #

### Additional comments 🎤
- The styling may need changing. Wrapping the notification message in a context menu lowered the text a few pixels. 
- the context menu doesn't get highlighted
- we may want to change the notification text & color, and implement a progress bar. 

image
image
image
image


ui/src/utils/auto_updater.rs
```diff
@@ -0,0 +1,137 @@
...
+        bail!("unknown OS type. failed to find binary");
+    };
+
+    let binary_dest = match FileDialog::new()
+        .set_directory(dirs::home_dir().unwrap_or(".".into()))
+        .pick_folder()
+    {
+        Some(x) => x,
+        None => {
+            log::debug!("update download cancelled by user");
```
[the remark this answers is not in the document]
sdwoodbury: no it isn't

ui/src/utils/auto_updater.rs
```diff
@@ -0,0 +1,206 @@
...
+
+    Ok(latest_release.tag_name)
+}
+
+fn get_client() -> Result<Client, reqwest::Error> {
+    let mut headers = header::HeaderMap::new();
+    headers.insert(
+        "User-Agent",
+        header::HeaderValue::from_static(
+            "Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/111.0",
```
dariusc93, line 141: Instead of a useragent of a browser, we should use a custom one. Maybe the name and version of the application itself?

ui/src/components/settings/sub_pages/about.rs
```diff
@@ -1,15 +1,138 @@
...
+                    icon: Icon::ArrowDown,
+                    onpress: move |_| {
+                        // be sure to update this before closing the app
+                        state.write().mutate(Action::DismissUpdate);
+                        if let Some(dest) = download_state.read().destination.clone() {
+                            std::thread::spawn(move || {
+                                let cmd = if cfg!(target_os = "windows") {
+                                    "explorer"
+                                } else if cfg!(target_os = "linux") {
+                                    "xdg-open"
```
dariusc93, line 115: Just as a note, this might not work on wayland (not blocking right now but would likely need testing to be sure it does)

dariusc93 (approved):

ui/src/utils/auto_updater.rs
sdwoodbury: I'm willing to do that, but I worry that the GitHub api wants to only be accessed by a browser. When I didn't set any user agent it rejected my request. I can test it out and see what happens.

dariusc93: The last I checked, their API required a user agent otherwise it would get rejected. The agent name can be anything (such as the name of the project).

dariusc93: https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#user-agent-required

phillsatellite (approved):

PR #597, opened 2023-04-05
update(Select): Never more crash with duplicate values on select

lgmarchi:
### 1. Never more crash with duplicate values on select 
a. Remove duplicate values inside component, to avoid any kind of crash because of it 
image

- 

- Resolve #

### DIFF
--- a/kit/src/elements/select/mod.rs
+++ b/kit/src/elements/select/mod.rs
@@ -1,3 +1,5 @@
+use std::collections::HashSet;
+
 use dioxus::prelude::*;
 
 #[derive(Props)]
@@ -18,10 +20,17 @@ pub fn emit(cx: &Scope<Props>, s: String) {
     }
 }
 
+fn remove_duplicates(values: Vec<String>) -> Vec<String> {
+    let unique_values: HashSet<_> = values.iter().cloned().collect();
+    let mut unique_values_vec: Vec<String> = Vec::new();
+    unique_values_vec.extend(unique_values);
+    unique_values_vec
+}
+
 #[allow(non_snake_case)]
 pub fn Select<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
     let initial_value = cx.props.initial_value.clone();
-    let mut options = cx.props.options.clone();
+    let mut options = remove_duplicates(cx.props.options.clone());
     options.retain(|value| value != &initial_value);
     options.insert(0, initial_value.clone());
     let iter = IntoIterator::into_iter(options.clone());


kit/src/elements/select/mod.rs
```diff
@@ -18,10 +20,17 @@ pub fn emit(cx: &Scope<Props>, s: String) {
     }
 }
 
+fn remove_duplicates(values: Vec<String>) -> Vec<String> {
```
dariusc93, line 23: You could do this

```rust
fn remove_duplicates(values: Vec<String>) -> Vec<String> {
    let unique_values = HashSet::<_>::from_iter(values);
    Vec::from_iter(unique_values)
}

```

Or if you prefer, could do

```rust
fn remove_duplicates<I>(values: I) -> Vec<String>
where
    I: IntoIterator<Item = String>,
{
    let unique_values = HashSet::<_>::from_iter(values);
    Vec::from_iter(unique_values)
}
```

Though the first one might be more preferable but either way it would reduce the need of `cloned` since you can pass the iterator (or values) directly into `from_iter` and convert it back into a `Vec` from there, which should remove any duplicate values. 

Note: Using this method will change the order of the array since `HashSet` is ordered based on hash of the value, so keep that in mind if the order of the array matters. If it doesnt, then either or of these functions will due.

dariusc93: I would have provided an example using the reference instead, but that would have also gotten cloned internally either way so I left it like this

dariusc93: Just as another precaution, it might be best to check for any values before pushing into a vector a head of time

lgmarchi: > Just as another precaution, it might be best to check for any values before pushing into a vector a head of time

What do you mean exactly here @dariusc93 ?

dariusc93: > > Just as another precaution, it might be best to check for any values before pushing into a vector a head of time
> 
> What do you mean exactly here @dariusc93 ?

Where ever `Select` is being used at, it would be best to check the supplied vector there before passing it to `Select`. As an example, in `GeneralSettings`, themes are being supplied to `option`. Moving up into `get_available_themes`, it would be best to check the vector for any themes that may be there. Eg 

```rust
if !themes.contains(&theme) {
   themes.push(theme);
}
```
Although, because names are used in `GeneralSettings`, it would be best to manually implement `PartialEq` and checking to see if the name field equal each other. eg

```rust
#[derive(Clone, Debug, Default, Deserialize, Serialize)]
pub struct Theme {
    pub filename: String,
    pub name: String,
    pub styles: String,
}

impl PartialEq for Theme {
    fn eq(&self, other: &Self) -> bool {
        self.name.eq(&other.name)
    }
}
```

That way, when it checks to see if another `Theme` equals another, it would compare the name instead of the whole struct. 

Same with font too since the name is used there apart of the settings.

lgmarchi: Opened a PR for it, @dariusc93, thanks for tips.

https://github.com/Satellite-im/Uplink/pull/602

dariusc93: Does that PR supersede this one? If so, we could close this PR

phillsatellite (approved):

InfamousVague (approved):

PR #602, opened 2023-04-06
update(Themes): Never more duplicated themes

lgmarchi:
### 1. Never more duplicated themes
a. Just add if not contain same name 
image

b. Redundant code, but impl function to remove duplicates and sort Vec<Themes> by name 
image

image

- 

- Resolve #

### DIFF
--- a/common/src/state/ui.rs
+++ b/common/src/state/ui.rs
@@ -323,11 +323,26 @@ impl ToastNotification {
     }
 }
 
-#[derive(PartialEq, Clone, Debug, Default, Deserialize, Serialize)]
+#[derive(Eq, Clone, Debug, Default, Deserialize, Serialize)]
 pub struct Theme {
     pub filename: String,
     pub name: String,
     pub styles: String,
+}
+
+impl Theme {
+    pub fn remove_duplicates_and_sort_by_name(themes: &mut Vec<Theme>) -> Vec<Theme> {
+        themes.sort_by_key(|theme| theme.name.clone());
+        themes.dedup();
+        themes.sort_by_key(|theme| theme.name.clone());
+        themes.clone()
+    }
+}
+
+impl PartialEq for Theme {
+    fn eq(&self, other: &Self) -> bool {
+        self.name.eq(&other.name)
+    }
 }
 
 #[derive(PartialEq, Clone, Debug, Default, Deserialize, Serialize)]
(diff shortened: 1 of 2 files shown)


common/src/state/ui.rs
```diff
@@ -323,13 +323,27 @@ impl ToastNotification {
...
-#[derive(PartialEq, Clone, Debug, Default, Deserialize, Serialize)]
+#[derive(Eq, Clone, Debug, Default, Deserialize, Serialize)]
 pub struct Theme {
     pub filename: String,
     pub name: String,
     pub styles: String,
 }
 
+impl Theme {
+    pub fn remove_duplicates_and_sort_by_name(themes: &mut Vec<Theme>) -> Vec<Theme> {
```
dariusc93: Would be best to let this function own `themes` than to borrow it

```suggestion
    pub fn remove_duplicates_and_sort_by_name(mut themes: Vec<Theme>) -> Vec<Theme> {
```

ui/src/utils/mod.rs
```diff
@@ -33,15 +33,16 @@ pub fn get_available_themes() -> Vec<Theme> {
...
+                    themes.push(theme);
+                }
             }
         }
     };
     add_to_themes(&STATIC_ARGS.themes_path);
     add_to_themes(&STATIC_ARGS.extras_path.join("themes"));
 
-    themes.sort_by_key(|theme| theme.name.clone());
+    let themes = Theme::remove_duplicates_and_sort_by_name(themes.as_mut());
```
dariusc93:
```suggestion
    let themes = Theme::remove_duplicates_and_sort_by_name(themes);
```

common/src/state/ui.rs
dariusc93: Although maybe we should move the logic here into `get_available_themes` since this is the only purpose and may not be used in other parts of the UI?

dariusc93 (commented):
Looks good to me. Did leave a couple of suggestions but can we also do the same here for `Font` just to be on the safe side? Though, Im not sure if it was fonts or themes that lead to #597 (I assume themes), but it might also to be safe there too.

InfamousVague (approved):
this is a good addition, this will prevent issues if a theme does become duplicated somehow

phillsatellite (approved):

PR #604, opened 2023-04-06
update(FilePreview): Fix title bar in file preview

### LINKED ISSUE #522: Files - Copy title bar styling from main window for popups in the file viewer popup
Image

Image

We should copy the same styling of the main window for the file preview window, so that we keep a consistent style.

lgmarchi:
### 1. Fix title bar in file preview
image

### 2. Define min size and hide title if it has small width 

https://user-images.githubusercontent.com/63157656/230480918-69240459-5c08-408a-aec1-b09fe7ce98fa.mov

### 3. Erase title in app as well when width is small 
image

- 

- Resolve #522


phillsatellite:
For small images this can start to look kinda messy but I understand that this is what was requested, @InfamousVague any input? I also noticed as seen in picture you are unable to exit the preview on linux 

Screenshot 2023-04-06 at 1 17 18 PM

sdwoodbury:
if a thumbnail can't be opened, it's probably better to display a notification instead of opening a window
image

ui/src/main.rs
```diff
@@ -992,7 +1006,7 @@ fn app(cx: Scope) -> Element {
     cx.render(main_element)
 }
 
-fn get_pre_release_message(cx: Scope) -> Element {
+fn get_pre_release_message(cx: &ScopeState) -> Element {
```
sdwoodbury: changing `get_pre_release_message` to take a `ScopeState` will conflict with a branch I'm working on. Long story short is in order to use `use_shared_state` in this Element, I had to create it like this: `get_pre_release_message{}` instead of like this: `get_pre_release_message(cx)`.

phillsatellite:
Everything is all good just one small issue Lucas is going to try to solve, on Linux and Windows the X in the toolbar is missing so user is unable to close out of the file preview

sdwoodbury (approved):

phillsatellite (approved):

WanderingHogan (approved):

PR #630, opened 2023-04-11
feat(extensions): release extensions alongside uplink

sdwoodbury:
### What this PR does 📖

- include the emoji and clear-all extensions in the installer
- tell the workflows to use a set version of rustc and refuse to build uplink if the version doesn't match

- Resolve #

### Additional comments 🎤
if after this PR is merged, your `rustup` toolchain gets updated and uplink refuses to build, you can fix it as follows: `rustup install 1.68.2`. you can verify your currently installed version with `rustup show`


phillsatellite:
Hey Stu, I had Sheldon and Hogan both try as well and we keep running into this error, is there a step we might be missing? 

image

sdwoodbury:
> Hey Stu, I had Sheldon and Hogan both try as well and we keep running into this error, is there a step we might be missing?
> 
> image

make sure the wix toolset is installed and the windows feature for ASP.NET is enabled. I'd argue that if the github workflow works, which I believe it does, then there aren't changes to request in this case :grin:

sdwoodbury:
as long as on windows, uplink is launched from the searchbar or the desktop shortcut, the assets and extensions are found correctly. for some reason, clicking on the .exe in C:\uplink\bin makes it not find the icons. I believe there's a problem with the way Dioxus looks up paths for images and that if one could pass an absolute path to an  `img` tag, this problem could be fixed.

ui/build.rs
```diff
@@ -10,6 +10,10 @@ use walkdir::WalkDir;
 
 fn main() -> Result<(), Box<dyn Error>> {
     let version = rustc_version::version().unwrap();
+    // rustup install 1.68.2 will ensure that the compiler matches
+    if version.major != 1 || version.minor != 68 || version.patch != 2 {
```
dariusc93: Hmm as I begin to think about it, hard coding it at this level might not be preferable in some instances especially when building it manually on a newer rust version and extensions arent exactly a priority. Might want to add a feature to `Cargo.toml` to bypass this check.

common/src/lib.rs
```diff
@@ -195,3 +196,27 @@ fn get_assets_dir() -> anyhow::Result<PathBuf> {
...
+pub fn get_extensions_dir() -> anyhow::Result<PathBuf> {
+    let extensions_path = if cfg!(target_os = "windows") {
+        let exe_path = std::env::current_exe()?;
+        exe_path
+            .parent()
+            .and_then(|x| x.parent())
+            .map(|x| x.join("extensions"))
+            .ok_or(anyhow::format_err!("failed to get Windows extensions dir"))?
+    } else if cfg!(target_os = "linux") {
+        PathBuf::from("/opt/im.satellite/extensions")
```
dariusc93, line 220: Not sure how I feel about the path being specifically hard coded but should be fine for now

tooshel (approved):
Except for the clippy warnings and the fact that the windows tests are not passing it looks good! The mascot images work, the themes work, the extensions are there. And of course the graphics being there is good too! Should go over this with Phill and also Luis to see what's up with the tests but it's close.

PR #634, opened 2023-04-12
feat(chat): Link preview

### LINKED ISSUE #525: Chat - Posting a url should load the preview of the url
This code can be ported from older versions. When pasting a link we should get info about the link and display it inline

Flemmli97:
### What this PR does 📖

- Adds ability to click on links
- The first link in a message gets a preview in chat:
  - Displays title, icon and a description
- Adds "--text-color-links" style
- Adds libraries: `linkify` (link parsing), `reqwest` (http requests) and `select` (http parsing)

- Resolve #525


phillsatellite:
Just a few things:

On Windows and MacOS a preview will appear for the link sent, but on Linux I could not get that to appear 

Screenshot 2023-04-12 at 12 39 26 PM
Screenshot 2023-04-12 at 12 41 24 PM

And on Windows the link color doesn't change from blue so if User has the default theme set then it makes it kinda challenging to read 
image

phillsatellite:
@Flemmli97 you are correct about changing themes fixing the blue text on links, is there anyway we can have it so the user doesnt have to change themes to fix that bug?

Flemmli97:
> @Flemmli97 you are correct about changing themes fixing the blue text on links, is there anyway we can have it so the user doesnt have to change themes to fix that bug?

~~i think that might be better as a separate issue. the problem is that when you run the app for the first time all themes get copied to the `./uplink/themes` folder. and any theme in that folder overwrites the default themes.
As such if we change anything in the `.scss` files it doesnt get reflected in the app.~~

~~we would need to find a way to~~
~~1. update the themes of that folder when we e.g. add new configurable fields to default themes
2. also make sure that user changed themes dont get replaced this way.~~

edit: seems i was a bit mistaken about the theme copy on first run. 

it might still be better though as a new issue

edit 2: issue created #639

InfamousVague (approved):

InfamousVague:
I would like to clean up the styling a bit in a future PR, looks good to go to me though!

lgmarchi (approved):

phillsatellite:
Re-tested and only thing different this time was the text in chat will all appear as black (with or without links) once this is fixed I think it should be good to go I can create a ticket to improve the UI a little and show link as Orange without being forced to switch themes in a future PR 

Replicated on all OS's

Screenshot 2023-04-20 at 12 06 34 PM

stavares843:
Captura de ecrã 2023-04-25, às 18 21 07

phillsatellite (approved):

dariusc93 (approved):

PR #648, opened 2023-04-13
chore(UI): add PrismJs, CSS updates and remove dev toolbar

InfamousVague:
### What this PR does 📖
- adds PrismJs for markdown highlighting 
- Lots of little tweaks to fix up UI issues.
- removes the toolbar for developer mode


ui/src/main.rs
```diff
@@ -1199,71 +1198,13 @@ fn get_titlebar(cx: Scope) -> Element {
         ))
     }
 
-    cx.render(rsx!(
```
sdwoodbury: was removing the toolbar for developer mode a mistake?

InfamousVague: Nah, it wasn't really useful and was more headache than it was worth just manually resizing the window.

sdwoodbury (approved):

sdwoodbury:
PR names like this will probably make it harder to tell which PRs changed what. Seems that the whole point of PRs is to aid in change tracking.

phillsatellite (approved):

sdwoodbury (changes requested):
image
image

Placing the prism scripts in `extra/assets` causes them to be included in the `.deb` and `.dpkg` files. Windows requires explicitly naming every file to be copied over so this doesn't affect Windows. 

If we want to not do this, we could do one of the following:
- change the installers to only copy over `extra/assets/img` which I believe is all that was used before
-  move `extra/assets/img`  to `extra/images`
- move `extra/assets/prism_langs`, `scripts`, and `styles` one directory up

sdwoodbury:
also if we're going to use this, I hope the spelling bot doesn't keep posting 73 comments every PR. Maybe we can tell it to allow the British spelling.

InfamousVague:
> image image
> 
> Placing the prism scripts in `extra/assets` causes them to be included in the `.deb` and `.dpkg` files. Windows requires explicitly naming every file to be copied over so this doesn't affect Windows.
> 
> If we want to not do this, we could do one of the following:
> 
>     * change the installers to only copy over `extra/assets/img` which I believe is all that was used before
> 
>     * move `extra/assets/img`  to `extra/images`
> 
>     * move `extra/assets/prism_langs`, `scripts`, and `styles` one directory up

Why not copy the entire extra folder and strip the stuff we don't want?

sdwoodbury (approved):
I'd  like to hold off on adding the prism_lang.zip folder to the wix installer until the other PR goes in. if this PR goes in first than I can update the other PR before merging it.

sdwoodbury:
main changes I made:
use prism autoloader to load the js files differently per installation method
image

image

for windows, unzip the folder during loading screen
image

PR #659, opened 2023-04-17
feat(EditGroup): Add and Remove participants in group

lgmarchi:
### 1. Add participants in a group 

https://user-images.githubusercontent.com/63157656/232572772-a8d54412-bd80-4537-83d0-96f9e561dbf8.mov

### 2. Remove participants from a group

https://user-images.githubusercontent.com/63157656/232572899-aea38733-f088-4160-8d6a-35e0de8da416.mov

- 

### Which issue(s) this PR fixes 🔨

- Resolve #545

PR #661, opened 2023-04-17

dariusc93:
### What this PR does 📖

- Update dependencies
- Change how messages are sent and distributed

- N/A

- This change does come with changes towards direct and group chat, though key focus would be mostly towards group chat due to how its now works compared to before. 

### Additional comments 🎤
- Messages prior to this PR wont be compatible with it (you can wipe the `messages` directory in the warp directory since this does not affect identity)
- When a user leaves (or deletes) the group conversation, it will attempt an update to the list for all connected members of the group, however the application may crash due to events not being implemented (will be resolved with #659)


sdwoodbury (approved):

lgmarchi (approved):

phillsatellite (approved):

phillsatellite:
@dariusc93  this can go in but cargo fmt --check is complaining about a lot of whitespace problems. So let me know if you want to fix or ignore them. it's good to go if you are okay with it

PR #662, opened 2023-04-17
chore(titlebar): Titlebar updates

WanderingHogan:
### What this PR does 📖
- We were re-using the same windows-controls component in three places. I moved it to kit.
- Moved the pre-release component to it's own component in ui/src (since it is application specific) and loaded it back in main, and loaded it into the titlebar instead of, hierarchically the same level as titlebar, and removed fixed positioning
- I cleaned up the CSS a little; mac override for a css class in titlebar (to accommodate for the mac windows-controls) and removed things we didnt need.
- removed the 'Uplink" title text. It just got in the way and made that section of the title hard to grab/move

- Resolve #

### DIFF
--- a/kit/src/components/mod.rs
+++ b/kit/src/components/mod.rs
@@ -19,3 +19,5 @@ pub mod message_typing;
 pub mod file_embed;
 
 pub mod context_menu;
+
+pub mod topbar_controls;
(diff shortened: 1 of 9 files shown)


dariusc93 (commented):
Could you run this on windows and linux to confirm that this atleast builds on other platforms? This is just to ensure imports arent missing in some way

PR #680, opened 2023-04-19
fix(themes): undo regression in themes

sdwoodbury:
### What this PR does 📖

- undo regression caused by https://github.com/Satellite-im/Uplink/pull/630
- themes should be reloaded from disk on startup
- not sure if the theme should be cleared from State if the theme file is deleted, but I added that too. 

image

- Resolve #


sdwoodbury:
turns out Flemli is working on the same thing

PR #707, opened 2023-04-24
fix(friends): Blocked text fix

### LINKED ISSUE #698: UI - Blocked Users status appears as REQUESTED NOW
**Issue:** 
Like the title says, when you block a user their status will change to REQUESTED NOW

**OS:**
All

**STR:**
Block a user

**Screenshot:**
Screenshot 2023-04-20 at 8 42 05 PM

Flemmli97:
### What this PR does 📖

- Fixes blocked use showing "Requested" text instead of "Blocked"

- Resolve #698


common/locales/en-US/main.ftl
```diff
@@ -76,8 +76,9 @@ friends = Friends
     .incoming_requests = Incoming Requests
     .outgoing_requests = Outgoing Requests
     .cancel = Cancel Request
-    .requested = Requested 
-    .sent = Sent
+    .requested = Requested { $time }
```
dariusc93: I wonder if we should remove the time out of this since in warp, it never shows the time when a user was blocked, and friend request no longer does (since I changed the corresponding functions to show DID rather than a struct of the request due to it being a bit redundant and causing more confusion).

dariusc93 (approved):
LGTM. Left a quick comment related to the use of time in the ftl.

phillsatellite:
I think my change is related to what Darius mentioned above. The NOW seems a little confusing maybe just have "Blocked"

Screenshot 2023-04-24 at 3 25 58 PM

phillsatellite (approved):

PR #709, opened 2023-04-24
update(Group): Fix crash when remove users from group

### LINKED ISSUE #674: Chat - Crash when removing all members in chat
On https://github.com/Satellite-im/Uplink/blob/dev/ui/src/components/chat/compose/mod.rs#L217-L220 we are enforcing a check where it requires at least 2 members to be in the chat, however in group chat, when all members are removed from the chat, it causes a crash. We should perform a check to allow no requirements for group chat while enforcing it for direct messages.

Relates to #612

lgmarchi:
### 2. Possible to edit group name after create a group

https://user-images.githubusercontent.com/63157656/234028109-12185988-b1f7-49c5-8130-76043f19842f.mov

- 

- Resolve #674


common/src/warp_runner/manager/commands/raygun_commands.rs
```diff
@@ -187,6 +193,20 @@ pub async fn handle_raygun_cmd(
             let r = raygun_remove_recipients_from_a_group(conv_id, recipients, messaging).await;
             let _ = rsp.send(r);
         }
+        RayGunCmd::UpdateConversationName {
+            conv_id,
+            new_conversation_name,
+            rsp,
+        } => {
+            let r = match messaging
```
dariusc93: this could be changed to something like

```rust
let r = messaging.update_conversation_name(conv_id, &new_conversation_name).await.map(|_| conv_id);
```

and you can pass `r` into `res.send` without the match condition

dariusc93 (approved):

phillsatellite (approved):

PR #727, opened 2023-04-25
chore(appium): delete chat accounts only if jobs succeeded

luisecm:
### What this PR does 📖

- Changing appium macos workflow to avoid deleting chat accounts if jobs failed
- Updating same workflow to fail execution if the prior job fails

- Resolve #


stavares843 (approved):

stavares843:
🎉

PR #736, opened 2023-04-27
update(ChatBar): Add message when user reach maximum of 1024 chars

### LINKED ISSUE #591: UI - Chats Input Bar - No error message displayed when user has reached the max number of chars
- After typing 1024 characters, user does not get any error message mentioning that maximum number of chars has been reached

Screenshot:
image

lgmarchi:
### 1. Add message when user reach maximum of 1024 chars

https://user-images.githubusercontent.com/63157656/234915855-2b7d4ac6-bdfb-486b-b63f-ba7f438ffa66.mov

- 

- Resolve #591


stavares843 (commented):
seems we can still add 2 more chars when the error appears


https://user-images.githubusercontent.com/29093946/234921332-b6586433-aeba-424f-aadc-bc120fced61d.mp4



macOS

lgmarchi:
> seems we can still add 2 more chars when the error appears
> 
>  Screen.Recording.2023-04-27.at.5.05.54.PM.mp4 
> macOS

@stavares843 about this, I made purpose to allow until 1025 chars, because sometimes happen a bug and you can not type last char. 

For example if I let max as 1024, sometimes will stop on 1023 and will not show message to the user. 

I did this way, because I think it is better to has a feedback. What is your opinion?

phillsatellite:
Hey @lgmarchi just a small request from my side, after the User sends the message the error message will persist after message is sent. It will also persist when switching through chats 

https://user-images.githubusercontent.com/93608357/234972356-8a9710a5-9438-4cb1-a77c-cd243ae301aa.mov

lgmarchi:
> Hey @lgmarchi just a small request from my side, after the User sends the message the error message will persist after message is sent. It will also persist when switching through chats
> 
>  Screen.Recording.2023-04-27.at.3.23.05.PM.mov

Fixed man!

phillsatellite (approved):

ui/src/components/chat/compose/chatbar.rs
```diff
@@ -274,117 +274,161 @@ pub fn get_chatbar<'a>(cx: &'a Scoped<'a, super::ComposeProps>) -> Element<'a> {
...
-        ignore_focus: cx.props.ignore_focus,
-        onchange: move |v: String| {
-            if let Some(id) = &active_chat_id {
-                state.write_silent().mutate(Action::SetChatDraft(*id, v));
-                update_send();
-                local_typing_ch.send(TypingIndicator::Typing(*id));
-            }
-        },
-        value: state
+    if value_chatbar.len() >= 1024 && !error.0 {
```
tooshel: Why don't we use that default prop above (which is in another file but still) so if we want to change it later we can change it in just one place?

lgmarchi: This is a particular case for chatBar, it does not receive Validation, and error is just maximum chars to 1024. I didn't see needs to do it in other file and be reusable. 

And particular for chatbar, it is not using dioxus states inside, so avoided to use it as well.

dariusc93 (approved):

sdwoodbury (changes requested):
i noticed that `error.1` isn't evaluated before displaying the error message. In other words, it looks like an error message for the wrong chat could be displayed.

sdwoodbury (approved):
looks like I was wrong about that bug, though I don't know why. I'll approve to unblock the PR but I still think that part could use improvement.

PR #738, opened 2023-04-27
feat(groups): Add blocked message functionality

### LINKED ISSUE #664: Groups - Add alert that message is from blocked User
**Issue:**
User A, B, and C are all in a group chat
User C blocks User B
User B can still message that group chat since User C 

We should let the message send through but have it collapsed with an alert message like _from blocked user_ with the option to view the message

Flemmli97:
### What this PR does 📖

- When user blocks someone messages will now appear as blocked with option to click on "Show Messages" to show them still and "Hide Messages" to hide them again

- Resolve #664


phillsatellite (approved):

InfamousVague (approved):
I think we don't use the `text-decoration: underline;` anywhere else but i don't think it's worth blocking this PR over. LGTM.

PR #749, opened 2023-04-28
fix(rendering): don't re-render every 5 seconds

sdwoodbury:
### What this PR does 📖

- the function to clear notifications returns a `bool` so you can use `write_silent()` and only update state if the typing indicator actually needs to be cleared. Not sure when `main.rs` stopped using that functionality, but this PR changes it back. 
- also add some more trace logs. 

- Resolve #

### DIFF
--- a/ui/src/main.rs
+++ b/ui/src/main.rs
@@ -642,6 +642,7 @@ fn app(cx: Scope) -> Element {
                 if !state.read().has_toasts() {
                     continue;
                 }
+                log::trace!("decrement toasts");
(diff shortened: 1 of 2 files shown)


tooshel (approved):
More logging to make debugging easier can only be a good thing right?

InfamousVague (approved):

dariusc93 (approved):

phillsatellite (approved):

phillsatellite:
waiting for tests to finish to put QA tested

PR #755, opened 2023-04-29
chore(autoupdate): remove autoupdate new action

luisecm:
### What this PR does 📖

- I just noticed that after merging the autoupdate workflow, the jobs that should be triggered again on pull requests updates are no longer running. Maybe it is just a setup on the action that would need to be done, but in the meantime we do some more research, I think its better to rever it
- Required checks that should run again on PR updates are not running:

image

- Checks on PR appears as zero, after the action updated to latest dev:
image

- Resolve #

This was going to be hard to catch before merging to dev


dariusc93 (approved):

dariusc93:
Guess that is something that should be looked more into

PR #782, opened 2023-05-04
feat(update): display toast after checking for updates

sdwoodbury:
### What this PR does 📖

- on the about page, after checking for updates, if no update is available, communicate this to the user with a toast

- Resolve #

### DIFF
--- a/common/locales/en-US/main.ftl
+++ b/common/locales/en-US/main.ftl
@@ -250,6 +250,7 @@ settings-about = About Settings
     .open-website-description = Opens our website in your default web browser.
     .open-codebase = Open Source Code
     .open-codebase-description = Opens the codebase in your default web browser.
+    .no-update-available = Your software is up to date!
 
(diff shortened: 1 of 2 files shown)


dariusc93 (approved):

common/locales/en-US/main.ftl
```diff
@@ -250,7 +250,9 @@ settings-about = About Settings
     .open-website-description = Opens our website in your default web browser.
     .open-codebase = Open Source Code
     .open-codebase-description = Opens the codebase in your default web browser.
-
+    .no-update-available = Your software is up to date!
+    .update-check-error = There was an error checking for the update. Please report this!
```
stavares843: this will also appear if user doesn't have internet, so maybe in the future we could add a error just for the internet connection

stavares843 (approved):

stavares843 (approved):

PR #783, opened 2023-05-04
fix(pkg): fix the .dpkg file not found

tooshel:
### What this PR does 📖

- fix this error
image

ERROR | failed to check for release: .dpkg file not found in software release

- Resolve #

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ opt-level = 3
 codegen-units = 1
 
 [workspace.package]
-version = "0.1.3"
+version = "0.0.1"
 rust-version = "1.68"
 
 [workspace.dependencies]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1339,7 +1339,7 @@ dependencies = [
 
 [[package]]
 name = "common"
-version = "0.1.3"
+version = "0.0.1"
 dependencies = [
  "anyhow",
  "base64 0.20.0",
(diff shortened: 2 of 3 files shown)


stavares843 (approved):

Cargo.toml
```diff
@@ -23,7 +23,7 @@ opt-level = 3
 codegen-units = 1
 
 [workspace.package]
-version = "0.1.3"
+version = "0.0.1"
```
dariusc93, line 26: Any reason to change the version of the package to 0.0.1?

dariusc93 (changes requested):

Cargo.toml
dariusc93: Actually as I check, I do see why, though this might be best to get bumped to 0.0.2

dariusc93 (approved):
Imma approve it since the version can be bumped later

Cargo.toml
tooshel: we should bump it to 0.0.2 when we want to do a release and have a matching tag. The tag and the release should be in sync. Also, for this to work, we needed it to be 0.0.1 so we can test another PR that tells you everything is up to date.

phillsatellite (approved):

Cargo.toml
stavares843: marking as resolved because darius approved and without marking as resolved we can't merge unless we force merge


Captura de ecrã 2023-05-04, às 23 27 53

PR #784, opened 2023-05-04
chore(locales): add more pt-PT locales

stavares843:
### What this PR does 📖

- chore(locales): add more pt-PT locales


tooshel (approved):
How do you compare to see what we are missing? We need this on every language file! :)

stavares843:
so the other day i used the diff checker and now i have been updating every time we add a new locale 😂

PR #786, opened 2023-05-05
feat(notification): Clickable notifications

Flemmli97:
### What this PR does 📖

- Makes it so notifications can be used to run actions inside the app:
  -  When you tap out of the app receiving new messages or friend requests makes a notification pop up
  - If that notification is a new message clicking on it will bring you to the chat where the message came from
  - If that notification is a friend request clicking on it will bring you to the friend request page
- On linux the actions are a dropdown
Screenshot from 2023-05-05 14-05-10
- ~~Needs testing on macos~~
- ~~Windows is missing atm~~
- Windows:
  - Windows Toast need to have an AppID assigned for the toasts. Currently it uses Powershell here so the toast shows that it comes from powershell. To be a valid AppID it needs to be also a shortcut in the startmenu. Im not sure what our plan regarding that will be but a possible way of doing is outlined here:
  [https://smsagent.blog/2020/10/20/adding-your-own-caller-app-for-custom-windows-10-toast-notifications/](https://smsagent.blog/2020/10/20/adding-your-own-caller-app-for-custom-windows-10-toast-notifications/)
  Changing the app id would then be done here:
  [https://github.com/Satellite-im/Uplink/pull/786/files#diff-3a3b0114c99d74ea1fc510bd6ed952782740c102efea7fbe4471830949776b7dR129](https://github.com/Satellite-im/Uplink/pull/786/files#diff-3a3b0114c99d74ea1fc510bd6ed952782740c102efea7fbe4471830949776b7dR129)

### Which issue(s) this PR fixes 🔨

- Resolve #694

PR #787, opened 2023-05-05
fix(notification): Move notification into separate thread

dariusc93:
### What this PR does 📖

- Move notification into separate thread

- N/A

### Additional comments 🎤
During testing for a separate PR, I begun to notice that messaging and other task performed were being blocked for a short duration. After investigating, it was determined that `Notification` was blocking, which would only be noticeable if there is an issue with the system notification handler, or if it has trouble communicating to one. Even without any issues, this function should be move into a separate thread due to its blocking nature to the thread.

### DIFF
--- a/common/src/notifications.rs
+++ b/common/src/notifications.rs
@@ -16,7 +16,7 @@ pub fn push_notification(
             .body(&content)
             .timeout(timeout)
             .show();
-        
+
         if let Some(sound) = notification_sound {
             Play(sound);
         }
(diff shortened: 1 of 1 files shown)


Flemmli97:
this is kinda addressed already in my draft pr #786.
the notification there is moved to another thread

sdwoodbury (approved):

phillsatellite (approved):

PR #808, opened 2023-05-09

dariusc93:
### What this PR does 📖

- Add limits on names for files and directories (limit is set to 256)
- Adds `Constellation::max_size` and `Constellation_current_size` to give the amount of data stored and the max allowed
- Switch from using `Constellation::put_stream` to `Constellation::put`
- Minor Optimizations

- N/A

Should not have any major changes that require testing

### Additional comments 🎤
- The limit for the names of the files or directories are a hard cap (keeping in compatibility with most major OS max name limits). Uplink still has it capped off at 64 characters, which is fine. If there are any interactions with `Directory` or `File` when setting or renaming and the name hits the limit, it will only use the first 256 characters. 
- Sending/receiving friend request with this PR to/from a prior commit will not work, however this does not require a new account.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -302,7 +302,7 @@ dependencies = [
  "num-traits",
  "rusticata-macros",
  "thiserror",
- "time 0.3.20",
+ "time 0.3.21",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


stavares843:
linter is failing 🔨

lgmarchi (approved):

stavares843 (approved):

PR #837, opened 2023-05-22
feat(chat): Pending msg rewrite

### LINKED ISSUE #631: task(ui): Improve file attachment UI by adding a loading indicator
**Task:**

- When sending a file, we should display a progress bar under the input bar representing each file that is being uploaded. (put progress bar with name on the bottom left under the bar and the progress percent / elapsed time on the right under the bar) We should move the preview of the file to above the chatbar like discord and other apps do. This isn't loading stuff but this is something we should (doesn't have to be you) work on. I can help with styling this. I just am worried i'll hit a bunch of issues trying to get the functionality, though it should be quite simple.

**Screenshot:**
Screenshot 2023-04-11 at 6 09 48 PM

### LINKED ISSUE #798: chat(ui): show messages with half opacity while waiting for loading for sending messages
**Issue:**
When we are waiting for messages to send, instead of showing the current loading indicator we want to show the message but with only half the opacity 

**OS:**
All

**Screenshot:**

https://user-images.githubusercontent.com/93608357/236917229-636ae583-3998-4f55-adf4-11146d7cf9c7.mov

Flemmli97:
### What this PR does 📖

- Rewrites the current system for tracking pending messages. Currently that is a simple int and as such also does not track what message was sent. This PR makes it so a dummy message containing the sent message content and their attachment is saved.
- Pending messages are now displayed with a lower opacity instead of previously as a loading icon
- Pending attachments show a loading bar displaying the progress of uploading that file
- Attachment for the chatbar are now above it instead of below
~~- TODO: As it stands now  when uploading attachments the state will be constantly updated causing the app to freeze up till uploading is done. see [https://github.com/Satellite-im/Uplink/commit/6f856a960062dad190144913f619292f4c019cc1#diff-e108e69cf0f5933cb78e6fed28b1c742b6ec02401d348c79d689db091b3aefcbR634](https://github.com/Satellite-im/Uplink/commit/6f856a960062dad190144913f619292f4c019cc1#diff-e108e69cf0f5933cb78e6fed28b1c742b6ec02401d348c79d689db091b3aefcbR634). __Add system to only update file component instead.__~~

- Resolve #798
- Resolve #631


phillsatellite (approved):

phillsatellite (approved):

common/src/state/pending_message.rs
```diff
@@ -0,0 +1,182 @@
...
+pub static MESSAGE_CHANNEL: Lazy<PendingMessageChannels> = Lazy::new(|| {
+    let (tx, rx) = tokio::sync::mpsc::unbounded_channel();
+    PendingMessageChannels {
+        tx,
+        rx: Arc::new(Mutex::new(rx)),
+    }
+});
+
+#[derive(Clone, Debug, PartialEq, Eq, Default)]
+pub struct PendingSentMessages {
```
sdwoodbury: instead of having a `PendingSentMessages` in state, can we just use a Vec? 
Also, instead of a `PendingSentMessage`, can we call it a `PendingMessage` (since there's no such thing as a pending incoming message) and give it a message Uuid and list of pending attachments (in the form of `warp::constellation::Progression`?

Also, in `state > chats.rs > Chat`, `pending_outgoing_messages` shouldn't be `skip`. Though within the `PendingMessage` struct, the list of pending attachments should be `skip`. I suggest this because if Uplink is closed and re-opened, although we would lose the channel that sends progress updates for the attachments, Uplink will still get the `Sent` event for the pending message when it is eventually sent.

common/src/state/pending_message.rs
```diff
@@ -0,0 +1,182 @@
...
+use tokio::sync::{
+    mpsc::{UnboundedReceiver, UnboundedSender},
+    Mutex,
+};
+use uuid::Uuid;
+use warp::{constellation::Progression, crypto::DID};
+
+use crate::warp_runner::ui_adapter::Message;
+
+pub struct AttachmentProgress {
```
sdwoodbury: I don't think this struct is needed. I think that updating the `pending_outgoing_messages` field of `state > chats.rs > Chat` should be sufficient. See the comment for the `PendingSentMessages` struct.

common/src/state/pending_message.rs
```diff
@@ -0,0 +1,182 @@
...
+    pub conversation_id: Uuid,
+    pub msg: PendingSentMessage,
+}
+
+pub struct PendingMessageChannels {
+    pub tx: UnboundedSender<AttachmentProgress>,
+    pub rx: Arc<Mutex<UnboundedReceiver<AttachmentProgress>>>,
+}
+
+pub static MESSAGE_CHANNEL: Lazy<PendingMessageChannels> = Lazy::new(|| {
```
sdwoodbury: I think `MESSAGE_CHANNEL` should be removed and that instead, a `WarpEvent::RayGunEvent` should be emitted when an update is received for a pending message. This suggestion fits within the current design for warp_runner and reduces the number of channels that `main.rs` needs to keep track of, with less being better.

kit/src/components/message/mod.rs
```diff
@@ -73,6 +76,10 @@ pub struct Props<'a> {
     parse_markdown: bool,
     // called when a reaction is clicked
     on_click_reaction: EventHandler<'a, String>,
+
+    pending: bool,
+
+    pending_attachments: Option<Vec<(String, Progression)>>,
```
sdwoodbury: I think `pending_attachments` could be a `Vec<Progression>` or an `Option<Vec<Progression>>`. The tuple I believe is unnecessary because `Progression` contains the file name. Also, `pending_attachments` should probably be renamed to `attachments_pending_upload`,  because there's already a property named `attachments_pending_download`. 

also, the `pending` field at least needs a comment. But I think it should be removed, and that the presence of a non-empty vector being passed to `pending_attachments`, with at least one file that hasn't completed uploading, should be sufficient to mark a message as "pending".

ui/src/components/chat/compose/messages.rs
```diff
@@ -356,29 +381,93 @@ struct AllMessageGroupsProps<'a> {
...
+}
+
+#[derive(Props)]
+struct PendingMessagesListenerProps<'a> {
+    data: &'a Rc<super::ComposeData>,
+    on_context_menu_action: EventHandler<'a, (Event<MouseData>, Identity)>,
+}
+
+//The component that listens for upload events
+fn render_pending_messages_listener<'a>(
```
sdwoodbury, line 410: I think it would be better if  `render_pending_messages_listener` wasn't used, and instead if a message was rendered as "pending" or not based on the information in `state > chats.rs > Chat`. 

In the `render_message` function, one could check within state to determine if the message is "pending" and if so, call/return a different render function if needed.  I think this way avoids making the code to display messages even more complicated.

sdwoodbury (changes requested):

common/src/state/pending_message.rs
Flemmli97: it is a struct cause it has functions for it that do things instead of directly operating on the vec. can rewrite it to be not struct bound if you like but i just thought this way its more clearer for organization

common/src/state/pending_message.rs
Flemmli97: see [https://github.com/Satellite-im/Uplink/pull/837/commits/bb3b02a64e31a6497f671da694a42f4c014ad899#diff-e108e69cf0f5933cb78e6fed28b1c742b6ec02401d348c79d689db091b3aefcbR619](https://github.com/Satellite-im/Uplink/pull/837/commits/bb3b02a64e31a6497f671da694a42f4c014ad899#diff-e108e69cf0f5933cb78e6fed28b1c742b6ec02401d348c79d689db091b3aefcbR619) where i merged them with the warp event channel and see if that is fine. we need to handle it separately because else it would constantly update `State` which will hang the app till upload is done

kit/src/components/message/mod.rs
Flemmli97: the pending field is necessary because even with an empty `pending_attachments` messages can still be pending.

ui/src/components/chat/compose/messages.rs
Flemmli97: `render_pending_messages_listener` was created so we can only update that component when progress events come in. it is the component that only contains all pending messages.

common/src/state/pending_message.rs
Flemmli97: related to message channel below

kit/src/components/message/mod.rs
sdwoodbury: I think you're right about `pending` but perhaps for a different reason - I realize now that `kit` can't access state.  I still hope that `pending` can be removed from things like `chat > compose > messages.rs > MessageProps` since the message_uuid and chat_uuid could be used to retrieve that information from state.

common/src/state/pending_message.rs
sdwoodbury: this looks good. I think the two `if let`'s could be combined into a `if let WarpEvent::Message(MessageEvent::AttachmentProgress))` and that the ` state.write().process_warp_event(evt);` could be wrapped in an `else{}` but yea, I like this approach :+1:

common/src/state/pending_message.rs
sdwoodbury: I appreciate that. I think that in this case, if we need to do that, the functions could be placed within `chats.rs`. But for now, I'd prefer removing the struct. 

It feels like `PendingSentMessages` simplified some things with `PendingSentMessage` which shouldn't have been there to begin with, and that we  won't need `PendingSentMessages`  if `PendingSentMessage` is removed. More about that - I think the `text` and `attachments` fields are redundant and that they are both part of `raygun::Message`. I also think it's incorrect to compare `PendingSentMessage` using the `text` instead of the message uuid (because two messages could have the same text). It seems that replacing `PendingSentMessage` with a new struct containing the message Uuid and a list of `warp::constellation::Progression` could be a lot simpler.

common/src/state/pending_message.rs
Flemmli97: resolved in [b403140](https://github.com/Satellite-im/Uplink/pull/837/commits/b40314027f7462cd3318a0705eb0e52caa106092)

kit/src/components/message/mod.rs
Flemmli97: wouldnt that cause unnecessary lookups instead? as if we already know at message creation that its pending we can just save it there already

common/src/state/pending_message.rs
Flemmli97: the uuid problem is though that pending messages are actually dummy messages. the uuid of the actual message that will come from warp will be different than the dummy one. 
we can use the uuid in cases where we update attachment progress (as at that point we will have a ref to the dummy msg uuid) but not in cases where warp will send a "Message Sent Successfully" event to ui.

also Warp `Message` have a vec of files as attachment but when sending msg with attachments we only have a vec of string so while we prob can convert the string into files it doesn't make much sense imo to do that. 
so i think `PendingSentMessage` is still useful

sdwoodbury: about the Uuid - excellent point. I bet we can request that `Raygun::send` be modified to return the `Message` which was created.  @dariusc93  is this doable? 

I agree that something is needed to contain the `Message` and the `attachment_progress` (not to be confused with `AttachmentProgress` ). 

As for the `Message` containing a Vec of `File`, each `File` has a `Name`. I could be wrong but I thought this would suffice and that an additional `Vec<String>` is unneeded.

stavares843:
removing changes requested label as review was addressed

Flemmli97:
> removing changes requested label as review was addressed

some things still missing

common/src/state/pending_message.rs
Flemmli97: when we send a message the list of attachment is passed as a string vec. warp then converts that into a vec of file (not uplink). also `File` here is a struct from warp and not the one from std lib in rust. it doesnt make much sense to also create dummy files when only thing relevant is the string here.

sdwoodbury (approved):
approving to unblock...

phillsatellite (approved):

stavares843:
@sdwoodbury requesting your review due to latest code

sdwoodbury (approved):

PR #839, opened 2023-05-24
fix(unreads): displaying wrong unread messages indicator

### LINKED ISSUE #830: bug(chat): displaying wrong unread messages indicator
Luis noticed the following bug

1. User A sends message to User B
2. User B does not have the convo open but see the sidebar with the unread messages counter (2) and time ago (now)
3. User A deletes the messages
4. User B still without the conversation open, sees again in sidebar "No messages sent yet", but the unread messages indicator is still there

image

macOS, m1

Jekrimo:
### What this PR does 📖

- de-increments unreads when a user deletes a message

- Resolve #830


phillsatellite:
This PR works really well! I did notice that when the unread message is deleted the notification bubble will persist on the app icon until the User clicks. I assumed we would want the notification bubble to also clear out. This happens on dev as well and a little out of the scope of this PR so to keep it simple I will create a ticket for it 😄 

https://github.com/Satellite-im/Uplink/assets/93608357/d33b307b-881c-4a2b-a163-a0eff8c8e595

phillsatellite (approved):

stavares843 (approved):

stavares843:
**AI CODE REVIEW**


The first patch fixes the issue with displaying the wrong unread messages indicator.
In the impl State block, the chats field is changed from private to public by modifying the visibility specifier from chats: chats::Chats to pub chats: chats::Chats.
In the MessageEvent::MessageDeleted match arm, the chat.unreads counter is decremented by 1 if it is greater than 0. This change ensures that the unread count is updated correctly when a message is deleted.
In the second patch, there is a minor change where the visibility specifier of the chats field is reverted back to private by removing the pub keyword.
The third patch removes the get_unread_count method from the impl State block, which seems unnecessary given that the chat.unreads field is already accessible publicly.
Overall, the code changes appear to address the issue mentioned in the commit message and make the necessary adjustments to handle unread message counts correctly.

sdwoodbury (changes requested):
I believe this won't work under the following circumstances:

there exist messages A, B, C, and D
messages A and B have been read
messages C and D are unread
Message A is deleted -> unreads would become 1 but should be 2.

Jekrimo:
> 

@sdwoodbury  I missed this sorry! I think currently there is no way to to only read two messages in a chat and still have unreads. I think it's either unread or read. C and D get read when A and B do no matter what. Unless I'm misunderstanding?

sdwoodbury:
> > 
> 
> @sdwoodbury I missed this sorry! I think currently there is no way to to only read two messages in a chat and still have unreads. I think it's either unread or read. C and D get read when A and B do no matter what. Unless I'm misunderstanding?

the sender can delete a message before the receiver opens the chat.

Jekrimo:
> 

that's what this is handling in this pr right? We currently don't have anything in place in which the user with unreads can only 'read' a certain number of unreads, it's all or nothing.

sdwoodbury:
> > 
> 
> that's what this is handling in this pr right? We currently don't have anything in place in which the user with unreads can only 'read' a certain number of unreads, it's all or nothing.

That's not what I said. if you read all the messages from a conversation and then switch conversations, and someone sends you new messages and also deletes old messages, think about that....

Jekrimo:
> 



> 

do you mean if someone is deleting messages from lets say a week ago even though the new message they just sent is not deleted?

sdwoodbury:
yes. observe

https://github.com/Satellite-im/Uplink/assets/8636602/c07a1385-99a3-4cf1-93ca-b7f3ac108621

sdwoodbury:
abandoning this in favor of #916

PR #845, opened 2023-05-26
fix(notification): new message notification sound after turning off all notifications

### LINKED ISSUE #836: bug(notification): new message notification sound after turning off all notifications
- Still receiving new message notification sound after turning off all notifications in settings tab
[]Windows
MacOS

Jekrimo:
### What this PR does 📖

- Add check for all enabled

- Resolve #836
-


phillsatellite (approved):

sdwoodbury (commented):
do we need to be able to configure a way for sounds to play when a message is received while notifications are disabled?

also, this seems like this is a good time to address the `todo` on line 370 - about only dispatching notifications when the app isn't in focus.

Jekrimo:
> do we need to be able to configure a way for sounds to play when a message is received while notifications are disabled?
> 
> also, this seems like this is a good time to address the `todo` on line 370 - about only dispatching notifications when the app isn't in focus.

EDIT: what do you mean by `way for sounds to play when a message is received` would these not still be a notification? I'm trying to think of an exception past like audio messages but that dif

I don't see us turning off notifications sounds as the same thing as turning off all sounds for the app. We definitely need a sound / audio manager in settings but I don't think they would be connected, right? @sdwoodbury

sdwoodbury:
> > do we need to be able to configure a way for sounds to play when a message is received while notifications are disabled?
> > also, this seems like this is a good time to address the `todo` on line 370 - about only dispatching notifications when the app isn't in focus.
> 
> EDIT: what do you mean by `way for sounds to play when a message is received` would these not still be a notification? I'm trying to think of an exception past like audio messages but that dif
> 
> I don't see us turning off notifications sounds as the same thing as turning off all sounds for the app. We definitely need a sound / audio manager in settings but I don't think they would be connected, right? @sdwoodbury

My understanding is that what you call a notification has 2 parts - a toast (what i'm calling a Notification) and a sound.

stavares843:
as discussed [here](https://discord.com/channels/997540176488300544/1023763769081475072/1115692644887179285) that needs a bigger refactor in the settings, merging to improve that later

PR #866, opened 2023-06-07
feat(blink): add blink to uplink

sdwoodbury:
### What this PR does 📖

Lets you call people (audio only). The UI has been simplified - the media component was removed from the compose page for now. 

image

image

I noticed the following at the top of `main.rs`
```
#![allow(dead_code)]
#![allow(unused_variables)]
```

IDK why we need that for uplink. Unless someone strongly objects, I propose removing this. 

- Resolve #

### DIFF
--- a/ui/src/components/media/player.rs
+++ b/ui/src/components/media/player.rs
@@ -153,7 +153,7 @@ pub fn MediaPlayer(cx: Scope<Props>) -> Element {
                 text: cx.props.end_text.clone(),
                 onpress: move |_| {
                     state.write().mutate(Action::ClearCallPopout(window.clone()));
-                    state.write().mutate(Action::DisableMedia);
+                   // state.write().mutate(Action::DisableMedia);
                 }
             },
             Button {
--- a/ui/src/components/media/player.rs
+++ b/ui/src/components/media/player.rs
@@ -42,9 +42,9 @@ pub fn MediaPlayer(cx: Scope<Props>) -> Element {
     let silenced = state
         .read()
         .ui
-        .current_call
-        .clone()
-        .map(|x| x.silenced)
+        .call_info
+        .active_call()
+        .map(|x| x.call_silenced)
         .unwrap_or(false);
 
     let silenced_str = silenced.to_string();
--- a/ui/src/components/media/player.rs
+++ b/ui/src/components/media/player.rs
@@ -29,7 +29,7 @@ pub struct Props {
 }
 
 #[allow(non_snake_case)]
-pub fn MediaPlayer(cx: Scope<Props>) -> Element {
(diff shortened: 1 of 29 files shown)


phillsatellite:
I noticed in 1on1 calling, if User A & B are talking and User B hangs up, User A is still left in the call alone. When User B tries to call them back the call will not go through until User A ends the original call they remained in. I was talking with Matt about that and we are thinking if its a 1on1 call, and one of the users hangs up it should end the call for both users

https://github.com/Satellite-im/Uplink/assets/93608357/475801a3-1301-4771-97ac-f810d886480b


I don't think its THAT major that we would need to hold up the PR since this is such an awesome feature to be added. If we want to merge I will create a ticket

ui/src/components/media/player.rs
```diff
@@ -36,20 +29,25 @@ pub struct Props {
 }
 
 #[allow(non_snake_case)]
-pub fn MediaPlayer(cx: Scope<Props>) -> Element {
+pub fn _MediaPlayer(cx: Scope<Props>) -> Element {
     let state = use_shared_state::<State>(cx)?;
 
     let window = use_window(cx);
 
+    //let ch = use_coroutine(cx, |mut rx: UnboundedReceiver<_>| {
```
dariusc93: Probably want to remove this if it will not be used.

dariusc93 (approved):
Everything looks good overall.

dariusc93:
> I noticed in 1on1 calling, if User A & B are talking and User B hangs up, User A is still left in the call alone. When User B tries to call them back the call will not go through until User A ends the original call they remained in. I was talking with Matt about that and we are thinking if its a 1on1 call, and one of the users hangs up it should end the call for both users
> Screen_Recording_2023-06-08_at_10.36.58_AM.1.mov
> 
> I don't think its THAT major that we would need to hold up the PR since this is such an awesome feature to be added. If we want to merge I will create a ticket

In direct calls, it might be best to have it hang up on both ends unless one side is awaiting for a timeout in case of a temporarily disconnection. In group chat, i think its acceptable to have the call like that but should probably look into hooking back into the existing call for said group. For now, maybe we can look into figuring out about hooking back into the call if User A is still on the call and User B tries calling back. Thoughts on this @sdwoodbury ?

sdwoodbury:
@dariusc93  & @phillsatellite,

 I like the idea of hanging up automatically if the other participant leaves a 1-1 call. As for hooking back  - sounds like automatically connecting the participants? - I think it would help if the UI showed which conversations had an ongoing call. We probably want to do that at some point, but for now it seems easiest to just automatically disconnect 1-1 calls.

phillsatellite (approved):

PR #869, opened 2023-06-08
fix(chat): Fix wrong timestamp

### LINKED ISSUE #790: bug(ui): timestamps are appearing as seconds ago
**Issue:**
In the sidebar when you get a message in and the timestamp updates it will only ever say seconds ago. I sent a message and 5 minutes later the timestamp appeared as '4 seconds ago'

**OS:**
All

**Screenshot:**
Screenshot 2023-05-05 at 3 44 54 PM

Flemmli97:
### What this PR does 📖

- Fixes timestamp for last message calculated wrongly

- Resolve #790

### DIFF
--- a/kit/src/components/user/mod.rs
+++ b/kit/src/components/user/mod.rs
@@ -34,7 +34,7 @@ pub fn get_time_ago(cx: &Scope<Props>) -> String {
     let duration: std::time::Duration = match c.to_std() {
         // for the sidebar, don't want the timestamp to increment a few seconds every time the typing indicator comes over.
         // prevent this by rounding down and giving a duration in minutes only.
-        Ok(d) => std::time::Duration::from_secs(d.as_secs() / 60),
+        Ok(d) => std::time::Duration::from_secs(d.as_secs() / 60 * 60),
         Err(_e) => std::time::Duration::ZERO,
     };
 


kit/src/components/user/mod.rs
```diff
@@ -34,7 +34,7 @@ pub fn get_time_ago(cx: &Scope<Props>) -> String {
     let duration: std::time::Duration = match c.to_std() {
         // for the sidebar, don't want the timestamp to increment a few seconds every time the typing indicator comes over.
         // prevent this by rounding down and giving a duration in minutes only.
-        Ok(d) => std::time::Duration::from_secs(d.as_secs() / 60),
+        Ok(d) => std::time::Duration::from_secs(d.as_secs() / 60 * 60),
```
WanderingHogan, line 37: could you just remove the divide by 60 here?

phillsatellite (approved):

kit/src/components/user/mod.rs
Flemmli97: i thought the point was to not have it as seconds? this way it only displays at most as minutes.
since e.g. 40 sec -> 0 sec -> "now"
or 65 -> 60 -> 1 min

WanderingHogan: I was just trying to point out

`std::time::Duration::from_secs(d.as_secs()),`

is always going to be the same as 

`std::time::Duration::from_secs(d.as_secs() / 60 * 60),`, 

right?

Flemmli97: no actually. as its an int division. so if e.g. `d.as_secs()` is 40 then `40 / 60 * 60` would be 0

WanderingHogan: oh, cool. I did not know that

WanderingHogan: thank you :)

WanderingHogan (approved):

PR #878, opened 2023-06-09
chore(copy): improve copy

stavares843:
### What this PR does 📖
- chore(copy): improve copy


WanderingHogan:
i think the current copy is better, thanks for the PR, though

PR #899, opened 2023-06-14
feat(User): Add default profile picture when user does not have a picture

### LINKED ISSUE #881: chore(chat): when user has no avatar just appears a white ball
STR:

- settings
- themes
- lunar eclipse theme
- chat
- when users have no avatar just appears a white ball 

Captura de ecrã 2023-06-09, às 19 35 39

macOS, m1

lgmarchi:
### 1. Create a random identicon when user create an account, and use it if user does not have a profile picture 
image
image
image

### 2. Split functions with warnings in 2 functions, to not happen problem anymore 
image
image

- Resolve #881


phillsatellite:
This is working really well for me on all OS's. There is one thing I think we could add depending on everyones thoughts,

If a User already has a profile picture and they go to Settings/Profile and right click _clear avatar_ ,instead of the old picture being replaced with a default avatar, it will be blank. 

https://github.com/Satellite-im/Uplink/assets/93608357/e34d25df-b2cb-4cc2-9462-ad25f147a473

If a new User with a default avatar goes to Settings/Profile and right clicks _clear avatar_ it will also leave them with a blank profile picture. (For New Users maybe we can disable the _clear avatar_ option)

https://github.com/Satellite-im/Uplink/assets/93608357/a538f7d7-7ece-4830-81e4-bbb2a5a51d1b

lgmarchi:
> This is working really well for me on all OS's. There is one thing I think we could add depending on everyones thoughts,
> 
> If a User already has a profile picture, when they go to Settings/Profile and right click _clear avatar_ instead of being old picture being replaced with a default avatar, it will be blank.
> 
> If a new User with a default avatar goes to Settings/Profile and right clicks _clear avatar_ it will also leave them with a blank profile picture. (For New Users maybe we can disable the _clear avatar_ option)
> 
>  Screen.Recording.2023-06-14.at.11.36.40.AM.mov

Ooh, I didn't know about this option, let me fix it right now my friend, you are totally correct

phillsatellite:
> > This is working really well for me on all OS's. There is one thing I think we could add depending on everyones thoughts,
> > If a User already has a profile picture, when they go to Settings/Profile and right click _clear avatar_ instead of being old picture being replaced with a default avatar, it will be blank.
> > If a new User with a default avatar goes to Settings/Profile and right clicks _clear avatar_ it will also leave them with a blank profile picture. (For New Users maybe we can disable the _clear avatar_ option)
> > Screen.Recording.2023-06-14.at.11.36.40.AM.mov
> 
> Ooh, I didn't know about this option, let me fix it right now my friend, you are totally correct

Sounds good bro, thank you!

lgmarchi:
> This is working really well for me on all OS's. There is one thing I think we could add depending on everyones thoughts,
> 
> If a User already has a profile picture, when they go to Settings/Profile and right click _clear avatar_ instead of being old picture being replaced with a default avatar, it will be blank.
> 
>  Screen.Recording.2023-06-14.at.11.48.14.AM.mov 
> If a new User with a default avatar goes to Settings/Profile and right clicks _clear avatar_ it will also leave them with a blank profile picture. (For New Users maybe we can disable the _clear avatar_ option)
> 
>  Screen.Recording.2023-06-14.at.11.36.40.AM.mov

Fixed!

stavares843:
Captura de ecrã 2023-06-14, às 18 33 36

missing linter 🔨

stavares843:
the comment from phil still happens with latest commit 🔨

phillsatellite:
@lgmarchi so I tried with the most recent commits on Windows/Linux/MacOS and I also had Sara try as well, and for both of us when User with a profile picture clears it it will remain blank instead of showing the default profile picture 

https://github.com/Satellite-im/Uplink/assets/93608357/292dc68c-73c1-4114-a52c-cc3187e88b0a

lgmarchi:
> @lgmarchi so I tried with the most recent commits on Windows/Linux/MacOS and I also had Sara try as well, and for both of us when User with a profile picture clears it it will remain blank instead of showing the default profile picture
> 
>  Screen.Recording.2023-06-14.at.1.30.48.PM.mov

I think there is no problem anymore

phillsatellite:
@lgmarchi So I tried with the latest commits across all OS's and had same behavior for each 

-User who already has a profile picture set goes to Settings and clicks _Clear avatar_
-Instead of changing to default Uplink photo, your profile picture will just remain the same 

Notes from trace2:
Screenshot 2023-06-14 at 4 50 55 PM

https://github.com/Satellite-im/Uplink/assets/93608357/88707095-1f98-4c7b-a6c4-3bed5c29b55b

lgmarchi:
> @lgmarchi So I tried with the latest commits across all OS's and had same behavior for each
> 
> -User who already has a profile picture set goes to Settings and clicks _Clear avatar_ -Instead of changing to default Uplink photo, your profile picture will just remain the same
> 
> Notes from trace2: Screenshot 2023-06-14 at 4 50 55 PM
> 
>  Screen.Recording.2023-06-14.at.4.50.42.PM.mov

Did you test with new accounts and had same behavior?

phillsatellite:
> > @lgmarchi So I tried with the latest commits across all OS's and had same behavior for each
> > -User who already has a profile picture set goes to Settings and clicks _Clear avatar_ -Instead of changing to default Uplink photo, your profile picture will just remain the same
> > Notes from trace2: Screenshot 2023-06-14 at 4 50 55 PM
> > Screen.Recording.2023-06-14.at.4.50.42.PM.mov
> 
> Did you test with new accounts and had same behavior?

When I tested with new account, was same behavior but without error in trace (I assumed it should not clear the default profile pic on new account thats why I didnt mention haha)

https://github.com/Satellite-im/Uplink/assets/93608357/9923582c-fb08-4ed5-b425-34230d3832c1

stavares843:
same here 🔨

common/src/warp_runner/mod.rs
```diff
@@ -185,6 +187,11 @@ async fn handle_login(notify: Arc<Notify>) {
                             Ok(_id) =>  match wait_for_multipass(&mut warp, notify.clone()).await {
                                 Ok(ident) => match save_tesseract(&warp.tesseract) {
                                     Ok(_) => {
+                                        let pfp = create_user_default_profile_picture(ident.did_key()).unwrap();
```
sdwoodbury: please don't use `unwrap()` unless it's in a unit test or the previous line ensures unwrap will never fail :pleading_face:

ui/src/layouts/storage/functions.rs
```diff
@@ -48,8 +48,17 @@ pub fn run_verifications_and_update_storage(
     }
 }
 
-// #[cfg(not(target_os = "macos"))]
+pub fn get_items_from_current_directory(cx: &Scoped<Props>, ch: &Coroutine<ChanCmd>) {
+    use_future(cx, (), |_| {
+        to_owned![ch];
+        async move {
+            sleep(Duration::from_millis(300)).await;
```
sdwoodbury: 300ms seems like a very short time. We're not processing hardware interrupts here. it's probably fine if it takes a second or two for uplink to detect a new file.

sdwoodbury (commented):
saving the default profile picture seems like a lot of work. It looks like if the same id is passed to [plot_icon::generate_png](https://docs.rs/plot_icon/0.3.0/plot_icon/fn.generate_png.html), the same image gets created. I think this would be easier. 

We could replace the trash icon on the profile picture (on the setting page) with a reset icon. That way it's clear that the profile picture would be reset to the default one. 

We could store the default profile picture in memory too, since we need it for comparisons in the profile settings page.

lgmarchi:
> saving the default profile picture seems like a lot of work. It looks like if the same id is passed to [plot_icon::generate_png](https://docs.rs/plot_icon/0.3.0/plot_icon/fn.generate_png.html), the same image gets created. I think this would be easier.
> 
> We could replace the trash icon on the profile picture (on the setting page) with a reset icon. That way it's clear that the profile picture would be reset to the default one.
> 
> We could store the default profile picture in memory too, since we need it for comparisons in the profile settings page.

First Paragraph, I thought the same, but or I change everywhere I am calling profile_picture, or change profile_picture() in warp to always get default polkadot when pfp is empty, updating default work as normal profile picture. So in this case way I did is less work or change warp as well. What do you think is better? And that tool from Rust, Deref would work here?

Second one, it can changed or not, other apps work with delete avatar or clear it, and back to default one as well. 

Third Paragraph can be done, but it will be used just few times and if user is in profile page.

sdwoodbury:
@lgmarchi 

> First Paragraph, I thought the same, but or I change everywhere I am calling profile_picture, or change profile_picture() in warp to always get default polkadot when pfp is empty, updating default work as normal profile picture. So in this case way I did is less work or change warp as well. What do you think is better? And that tool from Rust, Deref would work here?

I think that even if the change to `profile_picture()` is done in `Uplink` instead of `Warp` (as you have done in this PR), the user's DID could probably be used to generate the same picture, making it unnecessary to save the picture and save the path to it. That's all. Then again, if we ever want to use a different default profile picture, we may want to save the file. Your way may be better. 

I withdraw my comment about storing the picture in memory :grin: 

I still like the idea of a reset button.

lgmarchi:
> @lgmarchi
> 
> > First Paragraph, I thought the same, but or I change everywhere I am calling profile_picture, or change profile_picture() in warp to always get default polkadot when pfp is empty, updating default work as normal profile picture. So in this case way I did is less work or change warp as well. What do you think is better? And that tool from Rust, Deref would work here?
> 
> I think that even if the change to `profile_picture()` is done in `Uplink` instead of `Warp` (as you have done in this PR), the user's DID could probably be used to generate the same picture, making it unnecessary to save the picture and save the path to it. That's all. Then again, if we ever want to use a different default profile picture, we may want to save the file. Your way may be better.
> 
> I withdraw my comment about storing the picture in memory 😁
> 
> I still like the idea of a reset button.

I didn't change Warp, so your suggestion, is to change everywhere in Uplink where is calling profile_picture()? and always call a function passing this parameter to decide if use it or default profile picture? 

In memory, do you say in profile page memory as use_state, or in State? because has a base64 string inside State, and save it in Json, I think it is not good, and this string is huge

stavares843:
@sdwoodbury requesting your review due to your changes requested

phillsatellite:
Functionality wise this is good across all OS's, if another dev wants to check this out we can hopefully merge this in today. Awesome job @lgmarchi 🔥

phillsatellite (approved):

common/src/warp_runner/mod.rs
```diff
@@ -185,6 +187,11 @@ async fn handle_login(notify: Arc<Notify>) {
                             Ok(_id) =>  match wait_for_multipass(&mut warp, notify.clone()).await {
                                 Ok(ident) => match save_tesseract(&warp.tesseract) {
                                     Ok(_) => {
+                                        let pfp = create_user_default_profile_picture(ident.did_key()).unwrap_or_default();
+                                        match warp.multipass.update_identity(IdentityUpdate::Picture(pfp)).await {
+                                            Ok(_) => log::info!("Identity updated successfully with default polkadot placeholder"),
+                                            Err(e) => log::warn!("Failed to update identity with default polkadot placeholder: {}", e),
```
dariusc93: It might be best to perform a match on some errors to determine if its safe to actually proceed. Not a blocker though but just a thought on if an error should be returned to the oneshot on say an IO error, etc. Then again, this is for the account creation so this might not be worth returning an error.

dariusc93 (approved):
For the reset functionality, warp does have a option to use `IdentityUpdate::ClearPicture`, which would remove it from the identity. 

As for the default pfp, everything does look good, but I do wonder if it should be stored in warp or if it should be generated uplink if there is no picture available, assuming that it generates the same picture if the id is the same. Regardless though, this PR does look good from my pov with some non-blocker comments below

common/src/warp_runner/mod.rs
lgmarchi: You are right, I improved error handler here!

lgmarchi:
> For the reset functionality, warp does have a option to use `IdentityUpdate::ClearPicture`, which would remove it from the identity.
> 
> As for the default pfp, everything does look good, but I do wonder if it should be stored in warp or if it should be generated uplink if there is no picture available, assuming that it generates the same picture if the id is the same. Regardless though, this PR does look good from my pov with some non-blocker comments below

I agree with you, I just do not know if worth more to change method directly into Warp, let way it is now or change all calls from profile picture in Uplink. 

My main thought was to ensure that developers do not inadvertently call the wrong method and receive an empty string instead of the default profile picture, so I have decided to upload default profile picture to Warp. This is centred around a single location that handles the upload of the default profile picture. This strategy guarantees that the profile_picture() method, when invoked from the entity, will always return a valid result or default profile picture.

dariusc93:
> I agree with you, I just do not know if worth more to change method directly into Warp, let way it is now or change all calls from profile picture in Uplink.
> 
> My main thought was to ensure that developers do not inadvertently call the wrong method and receive an empty string instead of the default profile picture, so I have decided to upload default profile picture to Warp. This is centred around a single location that handles the upload of the default profile picture. This strategy guarantees that the profile_picture() method, when invoked from the entity, will always return a valid result or default profile picture.

As discussed, what could be done would be to use the `Identity` wrapper in https://github.com/Satellite-im/Uplink/blob/af4b0dc3b237acf7d24e924e1c6b3f6a568eeaab/common/src/state/identity.rs to create a `Identity::profile_picture` that would perform a check to determine if the picture is set within the inner `Identity` from warp and if not to fallback to using the default picture provided in this pr. 

I do have one possible idea on how i could have you do this in warp indirectly but would have to experiment a little first.

phillsatellite (approved):

dariusc93 (approved):

PR #911, opened 2023-06-15
update(StorageArch): New arch for storage feature

### LINKED ISSUE #858: chore(arch): new arch for rust app
About new arch for the app, ideia is to have files organized like this example: 

image

image

lgmarchi:

- Resolve #858 

Everyone, feel free to add your comments and thoughts, and suggestions to improve it.

@phillsatellite and @stavares843 to test this PR, just need to test Storage functions again. But this one I think will be better to test after devs review.


sdwoodbury (commented):
I think the folder `layouts/storage/data` is redundant. Nearly always, a data source / database has associated CRUD functions, which are exposed as a library. It appears to me that the view-specific `data` layer only performs CRUD functions. I believe that this functionality should be shared among views rather than re-created in each view. 

I think it's fine if the presentation layer uses the domain layer for data access. But I believe that the domain layer should rely on common modules to access data sources (ex: warp_runner). Also, it is my opinion that unless the domain layer is transforming data from one model to another, it would theoretically be OK if the domain layer exposed its data source to the controller.

I think that unless data is being transformed from one model to another, it's a waste to replicate every data access function (from the data source) inside the domain layer. I understand the value of transforming data into a view-specific model but don't think that this is always needed. For example, I think it would theoretically be OK if the files_page used a model which contained some types which were defined within the `warp` library. 

I brought up the previous paragraph because, from past discussions, it sounds like your preference is to perfectly separate everything through many layers of abstraction and data transformations, even if the net result is merely using a model which is defined within the View rather than using an equivalent model which is returned by some data source.

lgmarchi:
> I think the folder `layouts/storage/data` is redundant. Nearly always, a data source / database has associated CRUD functions, which are exposed as a library. It appears to me that the view-specific `data` layer only performs CRUD functions. I believe that this functionality should be shared among views rather than re-created in each view.
> 
> I think it's fine if the presentation layer uses the domain layer for data access. But I believe that the domain layer should rely on common modules to access data sources (ex: warp_runner). Also, it is my opinion that unless the domain layer is transforming data from one model to another, it would theoretically be OK if the domain layer exposed its data source to the controller.
> 
> I think that unless data is being transformed from one model to another, it's a waste to replicate every data access function (from the data source) inside the domain layer. I understand the value of transforming data into a view-specific model but don't think that this is always needed. For example, I think it would theoretically be OK if the files_page used a model which contained some types which were defined within the `warp` library.
> 
> I brought up the previous paragraph because, from past discussions, it sounds like your preference is to perfectly separate everything through many layers of abstraction and data transformations, even if the net result is merely using a model which is defined within the View rather than using an equivalent model which is returned by some data source.

If we were to summarize your points,
Would it be, not having the 'domain' and 'data' layers, and the call to WARP_CMD_CH, which happens in the 'datasource', would happen directly in the coroutine, as it was before, but now in a different file and better separated from the UI, is this correct?

Like just having this: 

image

stavares843:
missing linter 🔨

stavares843:
it doesn't build in windows

image



luis verified in his local machine

sdwoodbury:
> > I think the folder `layouts/storage/data` is redundant. Nearly always, a data source / database has associated CRUD functions, which are exposed as a library. It appears to me that the view-specific `data` layer only performs CRUD functions. I believe that this functionality should be shared among views rather than re-created in each view.
> > I think it's fine if the presentation layer uses the domain layer for data access. But I believe that the domain layer should rely on common modules to access data sources (ex: warp_runner). Also, it is my opinion that unless the domain layer is transforming data from one model to another, it would theoretically be OK if the domain layer exposed its data source to the controller.
> > I think that unless data is being transformed from one model to another, it's a waste to replicate every data access function (from the data source) inside the domain layer. I understand the value of transforming data into a view-specific model but don't think that this is always needed. For example, I think it would theoretically be OK if the files_page used a model which contained some types which were defined within the `warp` library.
> > I brought up the previous paragraph because, from past discussions, it sounds like your preference is to perfectly separate everything through many layers of abstraction and data transformations, even if the net result is merely using a model which is defined within the View rather than using an equivalent model which is returned by some data source.
> 
> If we were to summarize your points, Would it be, not having the 'domain' and 'data' layers, and the call to WARP_CMD_CH, which happens in the 'datasource', would happen directly in the coroutine, as it was before, but now in a different file and better separated from the UI, is this correct?
> 
> Like just having this:
> 
> image

yes

lgmarchi:
It was decided to not follow with this arch.

PR #929, opened 2023-06-22
fix(login): Login pin fix

### LINKED ISSUE #920: bug(login): no PIN accepted after clearing
**Issue:**
Log in like normal to the app and logout

On login type in an incorrect PIN, do not hit enter, right+click reset account 

After that it wont accept any PIN

You will likely need to sync the folders/re-trigger the .uplink folder creation when you wipe out that folder.

In `ui/src/layouts/unlock.rs`, when the fs::remove_dir_all, (~ line 273) is where you would likely want to do this (with a slight delay or some visual indicator that the folder is being deleted, a new one is created that is empty/ready for user to create a new account.

Please also fix the translation "Reset Account (Irreversable!)" -> "Reset Account (Irreversible!)"

Side note: Darius and I were discussing this and wanted to jot down some of the findings 
"the button shows up when there never been an account, which probably is the reason for the issue youre having. I think kev have a small fix for it if it doesnt exist, but the other problem might be related to the fact that its deleting the directory without reinitializing (or recreating) them again after redirecting its not recreating the stuff for uplink which is why it doesnt work after running it. Imo the button should only show up when an account does exist and it should also reinitialize uplink. "

**OS:**
All

**Screen recording:**

https://github.com/Satellite-im/Uplink/assets/93608357/146cf0d6-2a55-4dc8-a759-65baef521ba9

Flemmli97:
### What this PR does 📖

- Fixes account reset needing a restart
- Also fixes a spelling error

- Resolve #920


dariusc93:
After digging into it, it looks like it uplink needs to recreate the directory structure in order for it to properly proceed. This would mean needing to decouple the directory creation from `main.rs` into their own functions that can be call on at a later point. This PR should resolve one part of the issue though, but imo the button should probably but hidden if there is no account available while if there is one, it should show up to give the option of wiping out the directory. It might also be best to do something like

```rust
match fs::remove_dir_all(&STATIC_ARGS.dot_uplink) {
  Ok(()) => page.set(AuthPages::Unlock),
  Err(e) => {
       log::error!("Error deleting uplink directory: {e}");
       page.set(AuthPages::Unlock);
   }
};
```
(though to probably give an error notification if it fails to wipe it out; expanding out the error kind could probably help with that) 

Thoughts on this?

Flemmli97:
> After digging into it, it looks like it uplink needs to recreate the directory structure in order for it to properly proceed. This would mean needing to decouple the directory creation from `main.rs` into their own functions that can be call on at a later point. This PR should resolve one part of the issue though, but imo the button should probably but hidden if there is no account available while if there is one, it should show up to give the option of wiping out the directory. It might also be best to do something like
> 
> ```rust
> match fs::remove_dir_all(&STATIC_ARGS.dot_uplink) {
>   Ok(()) => page.set(AuthPages::Unlock),
>   Err(e) => {
>        log::error!("Error deleting uplink directory: {e}");
>        page.set(AuthPages::Unlock);
>    }
> };
> ```
> 
> (though to probably give an error notification if it fails to wipe it out; expanding out the error kind could probably help with that)
> 
> Thoughts on this?

for me it seems that calling requesting warp data automatically creates the needed files.
e.g. after i click on reset account the `.uplink` folder is deleted.
i then try to type in a new password and `.uplink` + `.uplink/.user/warp` gets recreated too since its trying a login attempt via `warp_cmd_tx.send(WarpCmd::MultiPass(MultiPassCmd::TryLogIn[...]`

phillsatellite (approved):

dariusc93:
> for me it seems that calling requesting warp data automatically creates the needed files. e.g. after i click on reset account the `.uplink` folder is deleted. i then try to type in a new password and `.uplink` + `.uplink/.user/warp` gets recreated too since its trying a login attempt via `warp_cmd_tx.send(WarpCmd::MultiPass(MultiPassCmd::TryLogIn[...]`

Warp does create the directory it will use for its data upon initialization or when it is storing data, but the issue lies in uplink and how it initializes the directories. Many of the paths are created in `main.rs`. 

https://github.com/Satellite-im/Uplink/blob/b74479b8a4e1b7bd5a7c2c9c379c63fa72b812d2/ui/src/main.rs#L177-L180

When the `dot_uplink` directory is deleted, and you proceed to create an account, warp is called and initializes the directory it will use, but uplink does not since it is created prior to initializing warp. 

Note: The lines above are only one of many where directory is created. Moving them into their own functions for initialization would be better in this case, otherwise it might be better to have the user close uplink and relaunch it prior to any attempts to creating a new account.

Flemmli97:
> > for me it seems that calling requesting warp data automatically creates the needed files. e.g. after i click on reset account the `.uplink` folder is deleted. i then try to type in a new password and `.uplink` + `.uplink/.user/warp` gets recreated too since its trying a login attempt via `warp_cmd_tx.send(WarpCmd::MultiPass(MultiPassCmd::TryLogIn[...]`
> 
> Warp does create the directory it will use for its data upon initialization or when it is storing data, but the issue lies in uplink and how it initializes the directories. Many of the paths are created in `main.rs`.
> 
> https://github.com/Satellite-im/Uplink/blob/b74479b8a4e1b7bd5a7c2c9c379c63fa72b812d2/ui/src/main.rs#L177-L180
> 
> When the `dot_uplink` directory is deleted, and you proceed to create an account, warp is called and initializes the directory it will use, but uplink does not since it is created prior to initializing warp.
> 
> Note: The lines above are only one of many where directory is created. Moving them into their own functions for initialization would be better in this case, otherwise it might be better to have the user close uplink and relaunch it prior to any attempts to creating a new account.

hmm ok. but from what im seeing is that only the themes and fonts folder (which are by default empty anyways) are missing which shouldn't really impact further processing in the app

Flemmli97:
i now did make it so the directories are recreated upon reset

lgmarchi (approved):

sdwoodbury:
this looks good. Unrelated to this PR, I found that uplink no longer loads fonts and themes when not in production mode. IDK when that change was made but it used to be that developer builds would load themes from `ui/extra/themes`

dariusc93 (approved):

PR #932, opened 2023-06-23

dariusc93:
### What this PR does 📖

- Implement a function field in the configuration to define a default profile picture.
- Added additional checks to current queue prior to sending out any request
- Reduced dependencies and minor cleanup
- Added additional logging
- Bump rust version to 1.70.0

- #936 

- ~~When #931 is merged, you will be able to properly clear the images out and be able to test it, otherwise when you clear the image it will not truly be cleared due to a terminator being supplied.~~
- This should fix some of the errors regarding sending messages out to users who are not online but unexpectedly not subscribed to their respected topic.
- Make sure to do `rustup update` to update rust to 1.70

### Additional comments 🎤
The configuration `default_profile_picture` should try to avoid any blocking code as much as possible. This function should probably be used for anything generated in memory or static images. If this isnt used and a user did not supply a profile picture, then no images will be apart of their `Identity` context.


sdwoodbury:
This branch doesn't compile when using rustc 1.68.2. It makes no sense why it doesn't compile, but I believe that the compiler is complaining about the reference captured by `DefaultPfpFn` within `IdentityDocument::resolve` in `warp-mp-ipfs/stc/store/document/identity.rs`. 

we could update the version of `rustc` but that would break our extensions. This might not be a problem since so far I think users would only have extensions released by us and software updates contain re-compiled extensions.

dariusc93:
> This branch doesn't compile when using rustc 1.68.2. It makes no sense why it doesn't compile, but I believe that the compiler is complaining about the reference captured by `DefaultPfpFn` within `IdentityDocument::resolve` in `warp-mp-ipfs/stc/store/document/identity.rs`.
> 
> we could update the version of `rustc` but that would break our extensions. This might not be a problem since so far I think users would only have extensions released by us and software updates contain re-compiled extensions.

I can check to see why it doesnt here shortly, although i dont recall making any changes to anything newer, however it probably wont hurt to bump the version up either. 

EDIT: Building on 1.68 does not cause any issues, though the error seems to be related to the async function itself so I may have a solution around it if it actually do fails to build

EDIT 2: Instead of the workaround, I went on and bump the version up to 1.70

phillsatellite:
When Sara and I both tested today seems that this helps the speed of the app dramatically 👍

sdwoodbury (approved):

phillsatellite (approved):

PR #934, opened 2023-06-26
fix(settings): Remove reset button when banner or profile picture is empty

### LINKED ISSUE #933: bug(profile): User can continue to reset Banner photo
**Issue:**
Within Settings -> Profile the user is able to right+click to reset the banner photo even after there is no photo set 

**OS"**
All

**STR:**
Clear banner photo 
After photo is cleared, clear it again 

**Screen recording:**

https://github.com/Satellite-im/Uplink/assets/93608357/95a4aed2-1ab9-415d-9463-bccd86efb772

Flemmli97:
### What this PR does 📖

- Removes the reset button for profile pictures or banner when the user does not have them

- Resolve #933


Flemmli97:
what I've seen now is that with some recent changes there is a default prof pic. and its made so when you have default pic the reset button is simply disabled instead of removed.
what is more preferred then here?
1. Simply disable the reset button when user has no banner/profile pic 
or 
2. Remove the button completely (which is what this PR currently does)

lgmarchi:
> what I've seen now is that with some recent changes there is a default prof pic. and its made so when you have default pic the reset button is simply disabled instead of removed. what is more preferred then here?
> 
> 1. Simply disable the reset button when user has no banner/profile pic
>    or
> 2. Remove the button completely (which is what this PR currently does)

@Flemmli97 I prefer first option for both items. Because I think is always good for user has a feedback of what is happening than just doesn't appear nothing. 

What do you think @XileHorizon ?

phillsatellite:
Just a minor little change
 
We should keep the same behavior for Banners that Profile Pic has. After you clear your banner, the option to clear it again should still be there but it should be greyed out. 


https://github.com/Satellite-im/Uplink/assets/93608357/ac76a516-3f50-4a6d-964d-3f83487b80c4

Screenshot 2023-06-26 at 12 56 05 PM

ui/src/components/settings/sub_pages/profile/mod.rs
```diff
@@ -43,6 +43,10 @@ pub fn ProfileSettings(cx: Scope) -> Element {
     let image = state.read().profile_picture();
     let image2 = image.clone();
     let banner = state.read().profile_banner();
+    let no_profile = image.eq("\0")
+        || get_user_default_profile_picture(state.read().did_key()) == image2.clone()
```
lgmarchi: Here, you just need comparison with image2.clone(), being different, it is already a reason to use default polkadot. 

And I would recommend to use name of the variable as "no_profile_picture"

ui/src/components/settings/sub_pages/profile/mod.rs
```diff
@@ -43,6 +43,10 @@ pub fn ProfileSettings(cx: Scope) -> Element {
...
+        || image.is_empty();
+    let no_banner = banner.eq("\0") || banner.is_empty();
```
lgmarchi: Here, would be nice to be "no_banner_picture" as well

lgmarchi (approved):
PR is approved, but left some suggestions to do and avoid unnecessary code.

phillsatellite:
New changes seem to be working well. I added _changes requested_ for the changes @lgmarchi asked for above. After that PR should be good to go

phillsatellite (approved):

stavares843:
linter is failing 🔨

PR #935, opened 2023-06-26
feat(Themes): Added accent color picker to themes.

InfamousVague:
### What this PR does 📖

image

image

image


phillsatellite:
Looks good! I found a few minor things here and there but nothing that is major enough to break anything. Only few things I found was when you are using Chat Search the highlight in unable to be seen when using Light mode. And it is very difficult to read the Pre-release | Issues/Feedback in the top bar. These are both things that I will create tickets for. Just need some dev reviews and this is good to go. Awesome PR 🔥 

Captura_de_ecra_2023-06-26_as_22 22 02

https://github.com/Satellite-im/Uplink/assets/93608357/e568d0f7-d4fc-4f69-976b-dcb4ce6f0996

phillsatellite (approved):

common/src/state/utils.rs
```diff
@@ -44,7 +44,16 @@ pub fn get_available_themes() -> Vec<Theme> {
...
+
+    let default = Theme {
+        filename: "".into(),
+        name: "Default".into(),
+        styles: "".into(),
+    };
+
+    themes.push(default);
+
+    themes.dedup(); // Why are we deduping here? We check above to only add if the theme doesn't already exist
```
dariusc93, line 56: Iirc this is done to remove any duplication (but do recall this being fixed else where but may be best to leave it in for now)

ui/src/main.rs
```diff
@@ -484,8 +484,20 @@ fn app(cx: Scope) -> Element {
             .map(|theme| theme.styles.clone())
             .unwrap_or_default();
 
+        let accent_color = state.read().ui.accent_color;
+        let accent_color = if let Some(color) = accent_color {
```
dariusc93, line 488: This could be simplified by doing something like the following but isnt a blocker

```rust
let accent_color = accent_color.map(|color| format!(":root {{ --primary: rgb({},{},{}); }}", color.0, color.1, color.2, )).unwrap_or_default();
```

dariusc93 (approved):

ui/src/main.rs
InfamousVague: Hopefully we'll be replacing that section of code when the todo item in the file is done so if it's not a blocker for now i'll merge it in since it'll be replaced anyhow.

PR #940, opened 2023-06-27
feat(Settings): Add more audio settings & volume slider.

InfamousVague:
### What this PR does 📖

This PR does not add actual functionality, simply adds the UI for more audio settings. Because there is no way to test the audio settings there is really no way to tell if it works or not, though. 

Working on an audio meter which will come in a future PR (as well as a single-choice component for echo cancellation).

There is a small bug where the microphone icons are incorrect which will also be resolved once the icon lib is updated.


phillsatellite:
Yo! Nothing major from me, I did notice the Linter was failing but I also wanted to get some other opinions about this, I noticed in light mode its a little hard to distinguish the Volume bars, I like to ask just in case its my eyes messing with me 


Screenshot_2023-06-28_at_10 58 16_AM

If everyone sides on its just my eyes this PR is good for me to get merged in after it gets reviewed by a dev Awesome feature 🔥

InfamousVague:
> Yo! Nothing major from me, I did notice the Linter was failing but I also wanted to get some other opinions about this, I noticed in light mode its a little hard to distinguish the Volume bars, I like to ask just in case its my eyes messing with me
> Screenshot_2023-06-28_at_10 58 16_AM
> 
> If everyone sides on its just my eyes this PR is good for me to get merged in after it gets reviewed by a dev Awesome feature 🔥

@phillsatellite 
I agree it's kinda hard to see but since it would involve changing around the light theme let's address tweaks to the light theme in another ticket if that's cool.

ui/src/components/settings/sub_pages/audio.rs
```diff
@@ -16,6 +21,59 @@ pub fn AudioSettings(cx: Scope) -> Element {
...
+                    onchange: move |_| {}
+                }
+            }
+
+            SettingSection {
+                section_label: get_local_text("settings-audio.sample-rate"),
+                section_description: get_local_text("settings-audio.sample-rate-description"),
+                Select {
+                    initial_value: "48000 Hz".into(),
+                    options: vec!["24000 Hz".into(), "48000 Hz".into(), "96000 Hz".into()],
```
sdwoodbury, line 71: the codec doesn't support 96000Hz

ui/src/components/settings/sub_pages/audio.rs
```diff
@@ -16,6 +21,59 @@ pub fn AudioSettings(cx: Scope) -> Element {
         div {
             id: "settings-audio",
             aria_label: "settings-audio",
+            SettingSection {
+                section_label: get_local_text("settings-audio.input-device"),
+                section_description: get_local_text("settings-audio.input-device-description"),
+                no_border: true,
+                Select {
+                    initial_value: "Default".into(),
```
sdwoodbury, line 29: "Default" should probably be replaced by a call to `get_local_text()`

Flemmli97:
I think it also would be better to update the text values of the sliders in real time (during dragging) instead of only after releasing it.

ui/src/components/settings/sub_pages/audio.rs
InfamousVague: These are just placeholders for when you wanted to hook things up. I just put them there so things were not empty.

InfamousVague:
> I think it also would be better to update the text values of the sliders in real time (during dragging) instead of only after releasing it.

Perhaps, but this might not be possible without Javascript which is un-desirable.

sdwoodbury (approved):

phillsatellite (approved):

PR #949, opened 2023-07-03
feat(UploadProgressBar): Create loading progress bar

### LINKED ISSUE #923: feat(files): new file upload bar
**Description:** Make a reusable component with this upload bar that can be used anywhere. Please plug this component into the files page in place of the current upload modal, which for the time being, will stay in chat.
 ----------- 
**Requirements:**
 ----------- 
- [ ] Upload bar should be above the breadcrumbs

| View      | Resource |
| ----------- | ----------- |
| File to upload      | Image |
| File is Uploading      | Image |
| Rename or Overwrite     | Image |

**Figma Mock:**
https://www.figma.com/file/EixccLouZCTmlkcy54RcEC/Files?type=design&node-id=0-1&t=x9RYHVzmFXO9tS0b-0

lgmarchi:
#### Create Upload Progress Bar reusable component
-  Reusable component upload progress bar 
-  Let user to add more files to upload even if there are files been uploaded 
-  Avoid folder navigation when files are been uploaded (Because file to be uploaded need to keep in that directory)
-  Add possibility to cancel an upload 
-  Show how many files are in queue to be uploaded 
-  Show how many files user are hovering to drop even if there are files been uploaded 
- Keep state showing upload progress even if user go out and back to storage page, possible to add more files in queue or cancel current file been uploaded 
- Organize files and variables, to decrease lines of code, and not hurt clippy rule passing many parameters to a function

##### Videos showing functionality

https://github.com/Satellite-im/Uplink/assets/63157656/7a1e8be8-1763-4030-b61e-98c507ef7ec8

https://github.com/Satellite-im/Uplink/assets/63157656/fdba3de6-1a54-4559-b6cc-2a98ea68a1d5

- Resolve Satellite-im/Uplink#923


common/src/warp_runner/manager/commands/constellation_commands.rs
```diff
@@ -373,31 +355,29 @@ fn go_back_to_previous_directory(
...
-    for file_path in files_path {
+    for file_path in files_path.clone() {
         let mut filename = match file_path
             .file_name()
             .map(|file| file.to_string_lossy().to_string())
         {
             Some(file) => file,
             None => continue,
         };
+        let _ = tx_upload_file.send(UploadFileAction::Starting(filename.clone()));
```
dariusc93: This should probably be moved until after the metadata function since this is indicating that a file is being uploaded but could silently fail. Moving it after `tokio::fs::metadata` or within the match condition might be the safer option in this case.

ui/src/layouts/storage/mod.rs
```diff
@@ -254,23 +235,30 @@ pub fn FilesLayout(cx: Scope<Props>) -> Element {
...
+                UploadProgressBar {
+                    are_files_hovering_app: upload_file_controller.are_files_hovering_app,
+                    files_been_uploaded: upload_file_controller.files_been_uploaded,
+                    disable_cancel_upload_button: upload_file_controller.disable_cancel_upload_button,
+                    on_update: move |files_to_upload: Vec<PathBuf>|  {
+                        add_files_in_queue_to_upload(upload_file_controller.files_in_queue_to_upload, files_to_upload, window);
+                    },
+                    on_cancel: move |_| {
+                        let _ = tx_cancel_file_upload.send(true);
+                        let _ = tx_cancel_file_upload.send(false);
```
dariusc93, line 254: Why is a second boolean being sent to the channel?

ui/src/layouts/storage/mod.rs
```diff
@@ -254,23 +235,30 @@ pub fn FilesLayout(cx: Scope<Props>) -> Element {
...
                 }
+                UploadProgressBar {
+                    are_files_hovering_app: upload_file_controller.are_files_hovering_app,
+                    files_been_uploaded: upload_file_controller.files_been_uploaded,
+                    disable_cancel_upload_button: upload_file_controller.disable_cancel_upload_button,
+                    on_update: move |files_to_upload: Vec<PathBuf>|  {
+                        add_files_in_queue_to_upload(upload_file_controller.files_in_queue_to_upload, files_to_upload, window);
+                    },
+                    on_cancel: move |_| {
+                        let _ = tx_cancel_file_upload.send(true);
```
dariusc93, line 253: ~~You should use `unbounded_send` as `send` returns a future that needs to await otherwise it might cause cause issues or wouldnt send since the future isnt progressing (or isnt being polled)~~

EDIT: I see its using `tokio` and not `futures` for the channel

dariusc93:
Do note that pushing each file upload into its own task while polling the stream would allow for each one to be uploaded possibly at the same time rather than being one-by-one. Long term, it might be best to assign each stream with `Notify` so when its being cancelled, it would signal that `Notify` for that specific file to allow the stream to be broken and the process to be cancelled for said file than utilizing a channel to break the stream.

ui/src/layouts/storage/mod.rs
lgmarchi: To reset value of cancel in receiver, for next files.

lgmarchi:
> Do note that pushing each file upload into its own task while polling the stream would allow for each one to be uploaded possibly at the same time rather than being one-by-one. Long term, it might be best to assign each stream with `Notify` so when its being cancelled, it would signal that `Notify` for that specific file to allow the stream to be broken and the process to be cancelled for said file than utilizing a channel to break the stream.

I understand what you are saying, I have some ideas to cancel specific files in the future, but for now, I did a lot of tests, and second file always wait first one be finished to start process to upload, so it had nice behavior on this scenario.

phillsatellite:
Screenshot 2023-07-05 at 12 17 49 PM

small change from me, I was talking with Sara and thought it looked a little odd that the progress box area stretches to very edge of the app window

lgmarchi:
> small change from me, I was talking with Sara and thought it looked a little odd that the progress box area stretches to very edge of the app window

It is following [Figma](https://www.figma.com/file/EixccLouZCTmlkcy54RcEC/Files?node-id=0%3A1&mode=dev). 

@XileHorizon What do you think?

stavares843:
linter is failing 🔨

ui/src/layouts/storage/mod.rs
dariusc93: In this case wouldnt it be better to move the cancel field into the inner portion so it would reset on the next iteration? Otherwise you could probably use `AtomicBool` instead to signal to the loop that its being canceled and then reset it in the next iteration.

XileHorizon:
UI-wise it looks good to me(tested on windows). The top bar stretches to the edge of the app window. The progress bar is above the file and folder area, not a part of it, so it wouldn't make sense to have it stop before the scroll bar. If you have suggestions on how to change it I would be open to discussing it though.

ui/src/components/files/upload_progress_bar/mod.rs
```diff
@@ -205,17 +205,17 @@ pub fn UploadProgressBar<'a>(cx: Scope<'a, Props>) -> Element<'a> {
     }
 
     if *cx.props.are_files_hovering_app.read() {
```
dariusc93: To make it cleaner, you could do the following

```rust
if !*cx.props.are_files_hovering_app.read() {
  return None;
}
  
cx.render(rsx!(
            div {
                class: "upload-progress-bar-container-file-count",
                p {
                    id: "upload-file-count",
                    class: "upload-file-count",
                }
            },
))
  
```

Not a blocker though for this

phillsatellite:
PR has been working well for me, @lgmarchi lemme know if you'd like to make the changes Darius's had suggested and I can retest again after

dariusc93:
> PR has been working well for me, @lgmarchi lemme know if you'd like to make the changes Darius's had suggested and I can retest again after

The suggestion isnt a blocker to the PR. Its just to make it a bit cleaner.

common/src/warp_runner/manager/commands/constellation_commands.rs
sdwoodbury, line 1: I was thinking about what happens when the `Progression::ProgressComplete` event is detected. It seems redundant to emit an event in the `match` statement and then again after the loop (the one on line 419). When I say the code emits an event, i'm referring to `UploadFileAction`. 

I noticed that `UploadFileAction::Uploading` takes a string and a file name but `UploadFileAction::Finished` just takes a string. It may be more consistent to pass a file name to `UploadFileAction::Finished`  as well. 

When `upload_progress` signals `Progression::ProgressComplete`, you emit this event: `UploadFileAction::Uploading` and give it the text "100%". This happens again in the `if` statements on lines `506` and `528`. That logic may as well be used when `Progression::ProgressComplete` is detected.  

the `if` statements on lines `506` and `528` emit `UploadFileAction::Uploading`  events but the text sent over probably gets immediately discarded because immediately afterward, `UploadFileAction::Finishing` is emitted.  This event should probably be the one emitted upon `Progression::ProgressComplete`.

if `Progression::ProgressFailed` is received, events are then emitted incorrectly indicating that the upload was successful.

common/src/warp_runner/manager/commands/constellation_commands.rs
```diff
@@ -449,6 +423,20 @@ async fn upload_files(
...
+                            log::trace!("starting upload file action listener");
+                            if let Ok(received_tx) = CANCEL_FILE_UPLOADLISTENER
+                                .rx
+                                .clone()
+                                .lock()
+                                .await
+                                .try_recv()
+                            {
+                                if received_tx {
+                                    upload_cancelled = true;
```
sdwoodbury: instead of using `upload_cancelled` and a `break` statement, you could use a [loop label](https://doc.rust-lang.org/reference/expressions/loop-expr.html) and then do this: `continue <loop label>`. This would help with handling the `ProgressFailed` event too.

common/src/warp_runner/manager/commands/constellation_commands.rs
```diff
@@ -519,15 +513,14 @@ async fn upload_files(
...
-                                FileTransferStep::Thumbnail(Some(())),
-                            ));
+                            let _ = tx_upload_file.send(UploadFileAction::Uploading((
+                                "100%".into(),
+                                get_local_text("files.thumbnail-uploaded"),
+                                filename.clone(),
+                            )));
                         }
                         Err(error) => {
                             log::error!("Not possible to update thumbnail for video: {:?}", error);
```
sdwoodbury, line 522: the error messages on lines 523 and 545 are inaccurate. I believe the intention is that this code execute when the upload is successful. The error occurs when the channel is closed.

sdwoodbury:
I got stuck on this page when there wasn't space available to upload a file 
image

sdwoodbury (dismissed):
requesting changes because if you try to upload a file that's too big, the UI gets stuck.

common/src/warp_runner/manager/commands/constellation_commands.rs
lgmarchi: `UploadFileAction::Finished` is when all process finished, all files from loop, and all thumbnails checked as well.

`UploadFileAction::Finishing` is when file was uploaded, but is missing maybe more files from loop, or thumbnail check is pendent yet, so it is correct your comment, I fix it.

`Progression::ProgressFailed` there was not send on this option, I added one with error now.

common/src/warp_runner/manager/commands/constellation_commands.rs
lgmarchi: I didn't get exactly what you mean, because this error is from function that tries to get thumbnail from a video file 
image

common/src/warp_runner/manager/commands/constellation_commands.rs
lgmarchi: Done. Much better this way.

lgmarchi:
> requesting changes because if you try to upload a file that's too big, the UI gets stuck.

Fixed it.


https://github.com/Satellite-im/Uplink/assets/63157656/40efaa80-ff1b-417d-9dbd-ddb07182390d

ui/src/components/files/upload_progress_bar/mod.rs
lgmarchi: Done.

common/src/warp_runner/manager/commands/constellation_commands.rs
sdwoodbury: it looks to me like the operation completed after the while loop.

sdwoodbury: oh, never mind. I was thinking the error message was the result of `tx_upload_file.send()` failing.

phillsatellite:
Along with the bug Sara found as well, I noticed when I tried to upload "spotify" it would get stuck at 0% and persist until I close down the app and re open 

image

lgmarchi:
> Along with the bug Sara found as well, I noticed when I tried to upload "spotify" it would get stuck at 0% and persist until I close down the app and re open
> 
> image

what is the size of spotify?

stavares843:
still got the same error as stu


image

stavares843:
the flow was adding a bunch of files and then the ones that he couldn't upload he returned error due to lack of space but got stuck in the loading bar

dariusc93 (approved):

phillsatellite (approved):

stavares843:
image


was still able to reproduce with the latest commits but fine with merging and improving later

stavares843:
https://github.com/Satellite-im/Uplink-Backlog/issues/47

PR #957, opened 2023-07-07
chore(Licenses): Added a place for us to display licenses

InfamousVague:
### What this PR does 📖
Adds a simple licenses page


tooshel (approved):

ui/src/components/settings/sub_pages/licenses.rs
```diff
@@ -0,0 +1,31 @@
...
+pub fn Licenses(cx: Scope) -> Element {
+    log::trace!("Licenses settings page rendered.");
+
+    cx.render(rsx!(
+        div {
+            id: "settings-licenses",
+            aria_label: "settings-licenses",
+            SettingSection {
+                section_label: "heroicons".into(),
+                section_description: "We have expanded upon the heroicons library we offer any additional icons under the same license as the original author.".into(),
```
dariusc93, line 19: This should probably get moved into translation so we can use something like `get_local_text("settings-licenses.description")`

ui/src/components/settings/sub_pages/licenses.rs
```diff
@@ -0,0 +1,31 @@
...
+                Button {
+                    text: "MIT".into(),
+                    appearance: Appearance::Secondary,
+                    icon: Icon::DocumentText,
+                    onpress: move |_| {
+                        let _ = open::that("https://github.com/tailwindlabs/heroicons/blob/master/LICENSE");
```
dariusc93, line 25: I thought we used `dioxus-heroicons`? If so or if its a forked of, it shouldnt point to tailwing but instead to dioxus-heroicons licenses instead, which would be a dual MIT and Apache 2.0 (The text could be `MIT/Apache 2.0` to represent dual license)

dariusc93 (commented):
Looks good so far. Made a comment below and if it is the case that we are using a fork, we should also copy over their license from the repo and into the `icon` crate in this PR.

ui/src/components/settings/sub_pages/licenses.rs
InfamousVague: See comment in chat, I don't think it is worth translating this since the end license they will be clicking is in English only anyways.

ui/src/components/settings/sub_pages/licenses.rs
InfamousVague: We do, but the icons generated are not hosted within the repo usually, they are built from directories containing svg files, which we have moved into our repo but are originally from tailwindlabs project. No matter what bundling tool we're using I figure it boils down to being their icons.

dariusc93 (approved):

phillsatellite (approved):

PR #984, opened 2023-07-14
update(Link): Link just with www, will work as text and not as link

### LINKED ISSUE #979: bug(chats): Message with a text starting with "www." is sent as link but the link is not working
1. Go to a chat
2. Type in the chat input bar a link starting with www. (like www.apple.com) and send message
3. Message sent is formatted as link without a preview
4. Click on the link just sent and we will see that the link is not working

Note: This is only happening with links sent starting with www.
If you send the link starting with https:// or http:// the link redirects to the URL

image

lgmarchi:
### 1. let "link" just with www work as text, and not as link
image

- Resolve #979

### DIFF
--- a/kit/src/components/message/mod.rs
+++ b/kit/src/components/message/mod.rs
@@ -88,7 +88,7 @@ pub struct Props<'a> {
 }
 
 fn wrap_links_with_a_tags(text: &str) -> String {
-    let re = Regex::new(r#"(?i)\b((?:https?://|www\d{0,3}[.]|[a-z0-9.\-]+[.][a-z]{2,4}/)(?:[^\s()<>]+|\(([^\s()<>]+|(\([^\s()<>]+\)))*\))+(?:\(([^\s()<>]+|(\([^\s()<>]+\)))*\)|[^\s`!()\[\]{};:'".,<>?«»“”‘’]))"#).unwrap();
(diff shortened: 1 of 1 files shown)


stavares843 (commented):
still doesn't work when we send www.google.pt is still not with the hyperlink and isn't clickable


i sent a PR here to address that

https://github.com/Satellite-im/Uplink/pull/991

stavares843:
closing due to https://github.com/Satellite-im/Uplink/pull/991 let me know your thoughts

lgmarchi:
> #991

I copied same behavior from other apps, when link does not have http or https, they send it as text.

stavares843:
yes but when luis added the ticket he meant like it should be clickable 🔨 

also other apps do that

example github website

www.google.pt

discord doesn't but its easier when sending url's to someone if we allow the www to be clickable

lgmarchi:
> yes but when luis added the ticket he meant like it should be clickable 🔨
> 
> also other apps do that
> 
> example github website
> 
> [www.google.pt](http://www.google.pt)
> 
> discord doesn't but its easier when sending url's to someone if we allow the www to be clickable

Actually Luis didn't say that, just that links with just www was not working

stavares843:
yes but I think he meant the expected behavior would be like being clickable the www that is

cc @luisecm if you want to enlighten us 😂

luisecm:
@lgmarchi @stavares843 yes, the issue was that www. links were "partially" working because the preview was there, but without icon and without link working. Looks like both behaviors are around there, example discord doest not convert www links but whatsapp and github does

PR #987, opened 2023-07-14
fix(File): Doesn't show warning when rename file name to same name

### LINKED ISSUE #983: bug(files): if you try to rename the same file the same name returns error
STR:
- go to files
- add a file
- rename to ''test''
- save
- works fine
- rename again the same file to ''test''
- says there is already a file with that name

https://github.com/Satellite-im/Uplink/assets/29093946/a7b57494-aaaa-4316-86e0-65c284a83d81

macOS

lgmarchi:
#### 1. Doesn't show warning when rename file name to same name

https://github.com/Satellite-im/Uplink/assets/63157656/aa551775-ac60-444f-967c-c07fee131834

- Resolve #983

### DIFF
--- a/ui/src/layouts/storage/mod.rs
+++ b/ui/src/layouts/storage/mod.rs
@@ -392,6 +392,7 @@ pub fn FilesLayout(cx: Scope<Props>) -> Element {
                         storage_controller.read().files_list.iter().map(|file| {
                             let file_name = file.name();
                             let file_name2 = file.name();
+                            let file_name3 = file.name();
                             let file2 = file.clone();
                             let file3 = file.clone();
                             let key = file.id();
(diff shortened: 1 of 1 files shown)


stavares843 (approved):
works great! 🎉 


we also need to fix for the folders, do you want a ticket for that to fix separately or you wanna address that here as well?

lgmarchi:
> works great! 🎉
> 
> we also need to fix for the folders, do you want a ticket for that to fix separately or you wanna address that here as well?

I think we can address here as well, small changes. Just a minute

stavares843:
neat, thanks! 🎉

lgmarchi:
Done.

stavares843 (approved):

sdwoodbury:
what happens if a file is given the same name as a folder (and vice versa)? Allowing that could be problematic

lgmarchi:
> what happens if a file is given the same name as a folder (and vice versa)? Allowing that could be problematic

It is allowed. Even with files, we can have test.mp4 and test.txt, it will not be a problem. And folder and files are like different types.

dariusc93 (approved):

PR #989, opened 2023-07-14
update(ChatBar): Add char counter

### LINKED ISSUE #977: task(chat): add char counter for textbox in chat
**Issue:**
In Core we had a char counter in the textbox for chat, it would be nice to have that in Uplink as well 

**Screenshot:**
image

lgmarchi:
### 1. Add char counter to the chatbar

https://github.com/Satellite-im/Uplink/assets/63157656/e85d68a9-8a73-4f81-892c-fc7fd8654fe5

- Resolve #977

### DIFF
--- a/kit/src/elements/textarea/mod.rs
+++ b/kit/src/elements/textarea/mod.rs
@@ -95,12 +95,17 @@ pub fn Input<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
     let update_char_counter_script = include_str!("./update_char_counter.js")
         .replace("$UUID", &id)
         .replace("$MAX_LENGTH", &format!("{}", max_length - 1));
+    let clear_counter_script =
+        r#"document.getElementById('$UUID-char-counter').innerText = "0/$MAX_LENGTH";"#
+            .replace("$UUID", &id)
+            .replace("$MAX_LENGTH", &format!("{}", max_length - 1));
 
     if *show_char_counter {
         eval(update_char_counter_script);
     }
 
     let cv2 = current_val.clone();
+    let cv3 = current_val.clone();
 
     let text_value = Rc::new(RefCell::new(value.to_string()));
     let text_value_onchange = Rc::clone(&text_value);
@@ -138,6 +143,9 @@ pub fn Input<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
                         let shift_key_as_modifier = evt.data.modifiers().contains(Modifiers::SHIFT);
 
                         if enter_pressed && !shift_key_as_modifier {
+                            if *show_char_counter {
+                                eval(clear_counter_script.to_string());
+                            }
                             onreturn.call((text_value_onreturn.borrow().clone(), true, evt.code()));
                         } else if enter_pressed && shift_key_as_modifier {
                             text_value_onkeyup.borrow_mut().push('\n');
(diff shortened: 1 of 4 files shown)


stavares843 (approved):
🎉 

works here but wasn't working for phil, he is re-checking

phillsatellite:
So I checked on my MacOS and Windows, Char counter appeared but did not change from 0/1024 for either as I typed

dariusc93 (approved):
LGTM although I do wonder if we could use the events from the modal instead of js for the character counter. Would be less js we would have to deal with. 

EDIT: This should not merge just yet until the issue with @phillsatellite is looked into

lgmarchi:
> LGTM although I do wonder if we could use the events from the modal instead of js for the character counter. Would be less js we would have to deal with.
> 
> EDIT: This should not merge just yet until the issue with @phillsatellite is looked into

I preferred to use js script to not re render every time, do you have a suggestion to improve that?

lgmarchi:
> So I checked on my MacOS and Windows, Char counter appeared but did not change from 0/1024 for either as I typed

I have tested on Both and worked normally, it had some problems yet and I updated, it should work now.

stavares843 (commented):
so was fine here, tested again, still fine here


but still doesn't work for phil

stavares843:
phil mentioned

> it got to 1/1024 but as I kept typing it did not keep counting now after I send message and type again Char counter doesnt leave from 0

phillsatellite:
Hello my friend! Okay so I tested with the latest commits on MacOS and on Windows and I had the same behavior for each. 

I type in some text and char counter changes from 0/1024 to 1/1024 but will never go higher than 1. When I remove all text goes back to 0 


https://github.com/Satellite-im/Uplink/assets/93608357/139b41ac-446e-4ea1-9b0f-9c360bba49b9

lgmarchi:
> Hello my friend! Okay so I tested with the latest commits on MacOS and on Windows and I had the same behavior for each.
> 
> I type in some text and char counter changes from 0/1024 to 1/1024 but will never go higher than 1. When I remove all text goes back to 0
> 
>  20230714-1710-33.5869428.mp4

This problem should not happen, but I got other problem here, I need to do a little refactor.

kit/src/elements/textarea/mod.rs
```diff
@@ -126,13 +143,24 @@ pub fn Input<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
                         let shift_key_as_modifier = evt.data.modifiers().contains(Modifiers::SHIFT);
 
                         if enter_pressed && !shift_key_as_modifier {
+                            if *show_char_counter {
```
sdwoodbury, line 146: if the space key increments the char counter, then enter + shift, which creates a newline, should probably also increment the char counter.

sdwoodbury:
> > LGTM although I do wonder if we could use the events from the modal instead of js for the character counter. Would be less js we would have to deal with.
> > EDIT: This should not merge just yet until the issue with @phillsatellite is looked into
> 
> I preferred to use js script to not re render every time, do you have a suggestion to improve that?

I was considering creating a child Element and passing it a hook (UseSharedState) but updating the hook  would probably cause the parent element to re-render too. JS may be the most efficient here.

kit/src/elements/textarea/mod.rs
lgmarchi: It is incrementing, code you commented is to clear char counter after send message

phillsatellite:
retested with same results as before

phillsatellite:
So I retested this using a new tmp account and it worked on all OS's but still wouldnt work on my older accounts. @stavares843 tested and it worked on her old account just fine. As long as it doesnt break for anyone else on old accounts and we wanna merge in I'm okay with that

Issue #972, opened 2023-07-11
bug(responsive): Sidebar will persist when switching pages 


phillsatellite:
**Issue:**
Adjust size of app to resemble a mobile device, when you switch tabs (ex going from files to friends) the sidebar will be open when you land on each page. When viewing on a mobile device we want the sidebar to be closed when switching pages. When a user switches to friends they should immediately see the friends page

**OS:**
MacOS

**Screen recording:**

https://github.com/Satellite-im/Uplink/assets/93608357/b1360477-d1be-4b01-ba4d-53731dddee86

phillsatellite:
Updated and replicated. MacOS only now

Issue #971, opened 2023-07-11
bug(responsive): Sidebar doesnt close when entering chat first time


phillsatellite:
**Issue:**
When you adjust your size of the app to resemble a mobile device and you try to enter a chat for the first time, the sidebar will not fully close. When you click the back button it will then fix this bug

**OS:**
MacOS
Windows 

**Screen recording:**

https://github.com/Satellite-im/Uplink/assets/93608357/ca88c6a5-cc90-4689-837a-005fa5bd1a8f

phillsatellite:
Updated title/description with more information

PR #1008, opened 2023-07-18
fix(sidebar): mobile view sidebar would stick

Jekrimo:
### What this PR does 📖

- when you downsize the app a "mobile view" on Linux everything works really well when you click a chat it goes into chat, switching pages shows correct page etc etc but on MacOS and Windows you cant click into chats, and sidebar will persist when switching between pages. 

- Resolve #1008

Currently only tested on Mac!

### DIFF
--- a/common/src/state/mod.rs
+++ b/common/src/state/mod.rs
@@ -68,7 +68,6 @@ pub struct State {
     #[serde(skip)]
     id: DID,
     pub route: route::Route,
-    pub chats: chats::Chats,
     friends: friends::Friends,
     #[serde(skip)]
     pub storage: storage::Storage,
@@ -493,9 +492,6 @@ impl State {
                         }
                     }
                     chat.messages.retain(|msg| msg.inner.id() != message_id);
-                    if chat.unreads > 0 {
-                        chat.unreads -= 1;
-                    }
                 }
 
                 if should_decrement_notifications {
@@ -862,19 +858,6 @@ impl State {
     fn clear_unreads(&mut self, chat_id: Uuid) {
         if let Some(chat) = self.chats.all.get_mut(&chat_id) {
             chat.unreads = 0;
-        }
-    }
-    /// get unreads  within a given chat on `State` struct.
-    ///
-    /// # Arguments
-    ///
-    /// * `chat_id` - The chat to get.
-    ///
-    pub fn get_unread_count(&mut self, chat_id: Uuid) -> u32 {
-        if let Some(chat) = self.chats.all.get_mut(&chat_id) {
-            chat.unreads
-        } else {
-            0
         }
     }
     /// Adds the given chat to the user's favorites.
--- a/common/src/state/mod.rs
+++ b/common/src/state/mod.rs
@@ -68,7 +68,7 @@ pub struct State {
     #[serde(skip)]
     id: DID,
     pub route: route::Route,
-    chats: chats::Chats,
+    pub chats: chats::Chats,
     friends: friends::Friends,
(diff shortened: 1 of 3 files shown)


phillsatellite:
Hi @Jekrimo so I was hoping to just approve this since it fixes both #971 and #972 but I did run into a minor issue. On Linux it seems to constantly stay in the mobile view no matter the size of the window 😔 

https://github.com/Satellite-im/Uplink/assets/93608357/11b2abab-0156-4468-9a28-25d9e506d4cc

Jekrimo:
> Hi @Jekrimo so I was hoping to just approve this since it fixes both #971 and #972 but I did run into a minor issue. On Linux it seems to constantly stay in the mobile view no matter the size of the window 😔
> 
>  Screen.Recording.2023-07-18.at.4.16.33.PM.mov

I was afraid that would happen, lookin into it!

lgmarchi:
@Jekrimo I recorded a video to help you, it just back to normal after 1200px

https://github.com/Satellite-im/Uplink/assets/63157656/d3a59d3b-448f-461b-aa22-9c36b3ebc634

Jekrimo:
> @Jekrimo I recorded a video to help you, it just back to normal after 1200px
> 
>  Screen.Recording.2023-07-19.at.11.45.07.mov

Oh right! Yes, Linux is messed up as Phil found out yesterday

sdwoodbury:
would it be better if a compile time flag checked if the build was for mobile? resizing a window isn't something one would do on mobile - maybe this is an easier way to address the issue. 

I noticed that on Desktop, other chat apps don't let the window get resized so small that there isn't room for both the sidebar and main part of the UI.

Jekrimo:
> would it be better if a compile time flag checked if the build was for mobile? resizing a window isn't something one would do on mobile - maybe this is an easier way to address the issue.
> 
> I noticed that on Desktop, other chat apps don't let the window get resized so small that there isn't room for both the sidebar and main part of the UI.

I would be cool with setting a min-width that stops desktop and tablet users from making the window too small. I feel like we were going to try the mobile only build awhile back but dropped it for some reason, maybe @InfamousVague remembers?

lgmarchi:
> > would it be better if a compile time flag checked if the build was for mobile? resizing a window isn't something one would do on mobile - maybe this is an easier way to address the issue.
> > I noticed that on Desktop, other chat apps don't let the window get resized so small that there isn't room for both the sidebar and main part of the UI.
> 
> I would be cool with setting a min-width that stops desktop and tablet users from making the window too small. I feel like we were going to try the mobile only build awhile back but dropped it for some reason, maybe @InfamousVague remembers?

I think we can set the min-width to 950, the way is today when we open the app, especially if we identify whether the system is windows, linux or macOS.

Other than that, min-width will need to be the exact size of the width of the tablet or mobile phone

stavares843:
is this ready to test again? 🔨

Jekrimo:
> is this ready to test again? 🔨

Now it should be!!! This should/might also conveniently fix 973

dariusc93:
~~If this is closed, should https://github.com/Satellite-im/Uplink/pull/1038 also be closed or was this issue closed in error @Jekrimo?~~

Edit: Nevermind, thought this was an issue

Jekrimo:
> ~If this is closed, should #1038 also be closed or was this issue closed in error @Jekrimo?~
> 
> Edit: Nevermind, thought this was an issue

It's my fault Darius, I was fighting some old code that kept pushing with my commits so I got slop with my branches, my bad and thanks for checking!!

PR #1025, opened 2023-07-21
feat(state): New state system

Flemmli97:
### What this PR does 📖

- This is the PR for a new state system utilising a ~~local subscription~~ dioxus signals.
- For now it implements this system for the message struct in chats.
- Goal is to have a system that allows one to update only a selected set of components without updating the whole state.
- Something to note is that since this only implements it for messages the actual impact will probably not be noticed much since in a lot of places components still subscribe to the generic `State` in uplink. We will likely only notice it if the whole `State` uses this system.
~~- TODO:~~
~~- [x] Local Subscription system~~
~~- [ ] Wrap needed uplink state fields in `LocalSubscription`~~
~~- [ ] Change methods in state so they properly subscribe to their correct local subs~~

`Signals` are used here just as a wrapper around the actual value. Writing to/Reading from it is simply done via the simple read and write functions. Dioxus will handle updating the correct components


sdwoodbury:
this looks almost the same as the implementation of UseSharedState - calling `cx.scope_id()` and adding that to a list of consumers. How is the new state system different? 
image

Flemmli97:
> this looks almost the same as the implementation of UseSharedState - calling `cx.scope_id()` and adding that to a list of consumers. How is the new state system different?
> image

The thing with `use_stared_state` is that it does not support sub components.
If you use `use_stared_state` you will subscribe to the whole `State` and any changes in `State` will cause a rerender for the subscribed component regardless of what was updated.

We can circumvent it by simply splitting `State` into multiple independent structs and register each of them with `use_stared_state` but that is hard to manage.

With this we can still mostly keep our current structure.

At the end we would have instead of
```rust
State {
  chat: Chats,
  some_other_field: String
}
```

```rust
State {
  chat: LocalSubscription<Chats>,
  some_other_field: String
}
```

and accessing the local subscription will automatically listen to changes in chat only. If now `some_other_field` is updated it will not cause a rerender for the component listening to chat

phillsatellite:
A lot of recently merged PR's features are not in here, if we can fix the conflicts so then I can update with the latest dev I can continue testing 👍

InfamousVague:
I think this would be good to give a try and see how we like this, it's a nice middle ground between subscribing to specific "nodes" in the state tree, but that sounds a bit difficult and perhaps super verbose, if we like this we can start working on moving more things over to improve performance.

phillsatellite (approved):

ui/src/components/chat/compose/messages.rs
```diff
@@ -591,39 +591,69 @@ struct MessagesProps<'a> {
...
+            state: state,
+        }))
+    })))
+}
 
-        // todo: add onblur event
-        rsx!(ContextMenu {
+#[derive(Props)]
+struct MessageContextProps<'a> {
+    message: &'a GroupedMessage<'a>,
```
jkelleyrtp: BTW @Demonthos and I were looking at this, and in an upcoming update we're going to disallow internal references here.

Your usecase is not problematic, but there are potential bugs abound with things like `RefCell<'a>` that we can't autoheuristic.

jkelleyrtp: Also, @Demonthos has been working on making the local subscriptions implicit by building some of the logic into Dioxus itself.

https://github.com/DioxusLabs/dioxus/pull/1300

That will likely be merged within the 0.4 cycle so you guys can benefit from it without another big version bump.

Flemmli97:
for now waiting for dioxus pr. seems to be more reasonable than merging this and then going back to this again later

ui/src/components/chat/compose/messages.rs
ealmloff: We found a way around breaking internally references. A new minor version of Dioxus core has been published, but it will not break this code

dariusc93:
Changing to draft until #1081 is complete

stavares843:
https://github.com/Satellite-im/Uplink/pull/1081 was merged 🎉

phillsatellite:
Ive been experiencing something a little odd on my account that I use daily. (this is happening on a m2 macbook running Ventura 13.5.1 on both `release` and `debug`)

```
0.4.2/src/scopes.rs:400:53
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fatal runtime error: Rust cannot catch foreign exceptions
zsh: abort      cargo run --release
```

So I typed in my PIN to login and app immediately crashes and shows this error in terminal 

thread '<unnamed>' panicked at 'already borrowed: BorrowMutError', /Users/phillwisniewski/.cargo/registry/src/index.crates.io-6f17d22bba15001f/dioxus-core-0.4.2/src/scopes.rs:400:53
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fatal runtime error: Rust cannot catch foreign exceptions
zsh: abort      cargo run --release

And showed this on the crash report 

Screenshot 2023-09-01 at 5 37 23 PM

I did a bit more testing and tried to see what would happen if I deleted my state.json, when I deleted I could now open the app again but as soon as I tried to open a chat it would crash again 

So I was thinking maybe a file I sent in one of the chats is whats causing it to crash so I again deleted my state.json and logged in, and removed all my old friends and had Sheldon add me as a friend, once we were friends I tried to start a new chat with Sheldon and it crashed as soon as I tried to chat 

Why I thought this was a little odd is because all the other friends that had the same chats as the crashing account were totally fine. And I even started up a fresh new account on m2 mac and added Sara and was able to send messages and files back and forth totally fine

https://github.com/Satellite-im/Uplink/assets/93608357/330536f3-a5dc-4aae-9ac8-94da422f3e70

Flemmli97:
can you try reproducing it again with the new crash log changes so it says more?

sdwoodbury:
@phillsatellite your error looks like it's related to extensions. if you delete your extensions  and recompile them and re-add them, does that error still happen?

luisecm:
Looks like somehow during the automated tests the emoji picker from the chat input bar is not loading:
image

Probably something with the emoji extension build is changing and the current steps from the workflow to copy the files are not working for this PR

phillsatellite:
> @phillsatellite your error looks like it's related to extensions. if you delete your extensions and recompile them and re-add them, does that error still happen?

I think you are right @sdwoodbury I deleted my emoji file from my extensions, recompiled and app opened fine and I could click on any chat without a crash. Although I need to test more because when I re-add the emoji extension I can still go to all the chats without the app closing but emoji's never appear for me

Flemmli97:
> > @phillsatellite your error looks like it's related to extensions. if you delete your extensions and recompile them and re-add them, does that error still happen?
> 
> I think you are right @sdwoodbury I deleted my emoji file from my extensions, recompiled and app opened fine and I could click on any chat without a crash. Although I need to test more because when I re-add the emoji extension I can still go to all the chats without the app closing but emoji's never appear for me

i actually had this bug rn with no emoji button too but that was on dev branch so its not related to this. restarting the app fixed it though before i that i updated the extension to a new built one. so maybe it was just because of and outdated extension

luisecm:
> > > @phillsatellite your error looks like it's related to extensions. if you delete your extensions and recompile them and re-add them, does that error still happen?
> > 
> > 
> > I think you are right @sdwoodbury I deleted my emoji file from my extensions, recompiled and app opened fine and I could click on any chat without a crash. Although I need to test more because when I re-add the emoji extension I can still go to all the chats without the app closing but emoji's never appear for me
> 
> i actually had this bug rn with no emoji button too but that was on dev branch so its not related to this. restarting the app fixed it though before i that i updated the extension to a new built one. so maybe it was just because of and outdated extension

When testing on Windows, I did with both DEV and your branch:

1. cargo update
2. cargo clean
3. cargo build --release -F production_mode (to create the .exe). 
4. Then copied the extensions files. (emoji_selector files with different extensions from target/release)

Finally, I saw that the emoji button is displayed on dev:

image

Not displayed in your branch:

image

Probably, I am missing any step above?

Flemmli97:
> > > > @phillsatellite your error looks like it's related to extensions. if you delete your extensions and recompile them and re-add them, does that error still happen?
> > > 
> > > 
> > > I think you are right @sdwoodbury I deleted my emoji file from my extensions, recompiled and app opened fine and I could click on any chat without a crash. Although I need to test more because when I re-add the emoji extension I can still go to all the chats without the app closing but emoji's never appear for me
> > 
> > 
> > i actually had this bug rn with no emoji button too but that was on dev branch so its not related to this. restarting the app fixed it though before i that i updated the extension to a new built one. so maybe it was just because of and outdated extension
> 
> When testing on Windows, I did with both DEV and your branch:
> 
> 1. cargo update
> 2. cargo clean
> 3. cargo build --release -F production_mode (to create the .exe).
> 4. Then copied the extensions files. (emoji_selector files with different extensions from target/release)
> 
> Finally, I saw that the emoji button is displayed on dev:
> 
> image
> 
> Not displayed in your branch:
> 
> image
> 
> Probably, I am missing any step above?

after some tests i found out that that is caused by an dioxus update. the update also affects the main branch so its not the signals branch causing this.

common/src/state/chats.rs
```diff
@@ -131,12 +139,41 @@ impl Chat {
...
+    fn from(value: SendableChat) -> Chat {
+        Chat {
+            id: value.id,
+            participants: value.participants,
+            conversation_type: value.conversation_type,
+            conversation_name: value.conversation_name,
+            creator: value.creator,
+            messages: {
+                let mut vec = VecDeque::new();
+                for msg in value.messages.iter() {
```
sdwoodbury: the `msg.clone()` could probably be avoided if `value` was `mut` and instead of `.iter()`, `value.messages.drain(..)` was used.

common/src/state/mod.rs
```diff
@@ -500,9 +501,9 @@ impl State {
                     if let Some(msg) = chat
                         .messages
                         .iter_mut()
-                        .find(|msg| msg.inner.id() == message.inner.id())
+                        .find(|msg| msg.read().inner.id() == message.inner.id())
```
sdwoodbury, line 518: I worry about `Signal::read()` here because the warp events are processed near the top of the VDOM. `process_warp_event()` is called by `use_app_coroutines()` which is called on line 169 of `main.rs`, in `app_layout`. 

After reading the comments in the [dioxus repository](https://github.com/DioxusLabs/dioxus/tree/master/packages/signals) it loos like `Signal::read()` automatically subscribes the caller's context. 

I suspect that storing `Signal`s in state isn't what we want. I could be wrong, but it seems like a `Signal` is better if used in a view rather than in a model.

Cargo.toml
sdwoodbury, line 1: I found [this module](https://github.com/DioxusLabs/dioxus/tree/master/packages/signals) in the master branch of Dioxus. I'm wondering if we still need Evan's branch here.

common/src/state/local_state.rs
sdwoodbury, line 1: I didn't see this used in the PR and am wondering if it is still needed now that there's a `dioxus_signals::Signal`.

Cargo.toml
ealmloff: There are some bug fix PRs that this repo needs that have not been merged. improved-signals is just a branch with all the signal bug PRs merged

common/src/state/mod.rs
ealmloff: Signals will only subscribe inside of a component or a selector. Inside of async futures spawned in Dioxus' runtime (like in a use_coroutine future), or event handlers, they will not cause any new subscriptions

sdwoodbury: That’s good news

sdwoodbury:
on Linux I can't build the project. I get this error
image

and found that in the [Dioxus branch](https://github.com/ealmloff/dioxus/blob/improved-signals/packages/desktop/Cargo.toml), the `tray` feature is indeed absent from `dioxus-desktop`.

dariusc93:
Pending #1229

phillsatellite (approved):

phillsatellite (approved):

sdwoodbury:
This looks cool but I'm not sure how much it will help at the moment. 

If a message is updated, the messages still need to be grouped and I believe that would still cause a re-render. Fortunately the chats page has been refactored to not store so many messages in memory. 

I wonder if this could help in other parts of the codebase. For example - if a friend comes online, it might be nice to not re-render stuff that doesn't include that friend. Still, for stuff like that I have a theory that it may be easier to have a top level element that pulls everything it needs from State and passes it to a child Element via props. That way if State is updated, in theory the child elements may have the same props and if props don't change, the element doesn't get re-rendered. (this is easier said than done because State gets used everywhere).

Flemmli97:
unless we do it across the whole app (and not just messages) the gain atm isn't really big. 
but e.g. editing messages should only update that one message component and not the whole chat

tooshel:
Closing this for now and we will revisit after the next release of Dioxus.

PR #1033, opened 2023-07-24
feat(groups): Allow any char in group names

### LINKED ISSUE #1030: improvement(groups): allow symbols when creating group
image

Flemmli97:
### What this PR does 📖

- Allows any character in group names. 

- Resolve #1030

### Additional comments 🎤
I am not sure if allowing any characters will cause any problems down the line but I see how it would.
If it does we probably want just a list of allowed special characters. (e.g. like `/._-` etc.)


dariusc93:
It might be best to allow only specific characters instead of any just to be safe since I can imagine it being problematic down the road

stavares843:
added label due to darius comment

Flemmli97:
changed to specific list. current allowed are ` .,!?_&+~(){}[]+-/*`. Which should cover the most commonly used ones

phillsatellite (approved):

dariusc93 (approved):

stavares843:
Captura de ecrã 2023-07-25, às 18 29 43

we should update the copy also, i can add a PR for that

stavares843 (approved):

PR #1038, opened 2023-07-24
fix(sidebar): mobile sidebar would stick on chat click

Jekrimo:
### What this PR does 📖

- when you downsize the app a "mobile view" on Linux everything works really well when you click a chat it goes into chat, switching pages shows correct page etc etc but on MacOS and Windows you cant click into chats, and sidebar will persist when switching between pages.

- Resolve #1008 

!Tested only on Windows & Mac currently!


phillsatellite:
So on Windows and Linux this is working really well, on MacOS its almost there. When you switch pages the sidebar will persist and sometimes when you try to open the sidebar while in a chat the sidebar will only open halfway. 

I know this has been a pain of a ticket would it be easier to fix MacOS in a separate PR?

Freaking awesome progress so far though! Thanks Jeff! 

https://github.com/Satellite-im/Uplink/assets/93608357/75114703-b2dc-4c86-a128-7058cb5f540f


https://github.com/Satellite-im/Uplink/assets/93608357/6fdc007a-5a74-43f2-bd67-3242fff2d319

Jekrimo:
I bet another close() was added somewhere that is confusing it again. I'll look into it!

phillsatellite:
Hey Jeff! So with the latest changes on Mac/Windows/Linux when in Desktop view the chats will not appear. Once user minimizes into the mobile view and they click into the chat it will re appear 

https://github.com/Satellite-im/Uplink/assets/93608357/59cec5d4-449d-4d06-b9d6-889ad4ba6850

Jekrimo:
I think I just pushed a fix that should be good!

stavares843:
thanks!

stavares843:
Captura de ecrã 2023-08-11, às 00 19 08

🔨

kit/src/components/nav/mod.rs
```diff
@@ -131,8 +137,12 @@ pub fn Nav<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
                         aria_label: aria_label.to_lowercase() + "-button",
                         icon: route.icon,
                         onpress: move |_| {
+                            //if route is friends or files then close, else leave sidebar open
+                            if f64::from(size.width) <= 600.0 * scale && !(route.to != "/friends" && route.to != "/files"){
```
sdwoodbury, line 141: if the code in `kit` is supposed to be reusable and not dependent on a particular UI, then you can't hard code the `/friends` route here. This logic should be called from the callback - `cx.props.onnavigate`. We should probably make a `Nav` element in `ui/src/components`. 

related - `kit` shouldn't rely on `State` either (line 108).

ui/src/components/chat/sidebar.rs
```diff
@@ -510,8 +512,7 @@ pub fn Sidebar(cx: Scope<Props>) -> Element {
                                 with_badge: badge,
                                 onpress: move |_| {
                                     state.write().mutate(Action::ChatWith(&chat_with.id, false));
-
-                                    if state.read().ui.is_minimal_view() {
```
sdwoodbury, line 514: `is_minimal_view` is used in a lot of places. if it doesn't work, it may be better to calculate `WindowMeta` and use `Action::SetMeta` to update state so that `is_minimal_view` works correctly.

sdwoodbury (changes requested):

kit/src/components/nav/mod.rs
Jekrimo: Dangit, I knew I was gunna put something dumb in kit at some point, thanks for the catch dude!

ui/src/components/chat/sidebar.rs
Jekrimo: I think you're right, prob going to need a state handler to calculate everything and rely on that

Jekrimo:
Closing as issue is resolved

PR #1042, opened 2023-07-24
chore(State): Includes some default values for the UI

InfamousVague:
### What this PR does 📖
Implements default for UI to better control default state values.

### Additional comments 🎤
This also includes some small fixes here and there as well as refactoring how we do the cargo.toml (because the old way was technically invalid TOML which caused most IDEs to complain.)


phillsatellite (approved):

phillsatellite:
Approved by me waiting for Sara too then good to go

stavares843 (approved):

PR #828, opened 2023-05-16
chore(cargo): Update dependency

dariusc93:
### What this PR does 📖

- Generate thumbnails for images internally rather than in uplink
- Remove function generating thumbnail for image
- Added small helper to generate base64 based on internal file type

### Which issue(s) this PR fixes 🔨

- N/A

### Special notes for reviewers 🗒️
- Not every image will generate a thumbnail. At the time, only PNG, JPG/JPEG, GIF, ICO, BMP, and WEBP will be supported for generating a thumbnail
- Clicking on an thumbnail in the file explorer will show the image/thumbnail at 500x500 (default, but configurable though recommend to leave it at this size or lower it down). A separate ticket may need to be opened to grab the image itself to display it in the modal rather than the thumbnail

### Additional comments 🎤
- Due to changes to `File` in warp, this will not be compatible with previous index (does not require new identity).
- Should unblock #728

PR #1045, opened 2023-07-25
fix(Attachments): Limit files per message and remove persistance when switching chat

### LINKED ISSUE #1026: bug(chat): If too many files are attached to a message, it will never send
**Issue:**
Like the title says, within chat if you send a message with attachments but you attach too many files to it they will never send or appear in chat 

**OS:**
All

**Screen recording:**

https://github.com/Satellite-im/Uplink/assets/93608357/9400178b-0902-453e-9029-e1ff429161ee

### LINKED ISSUE #1040: bug(chat): when sending a file filename is barely visible
SRT:
- go to chat
- send file
- name isnt visible

Captura de ecrã 2023-07-24, às 23 10 51

Captura de ecrã 2023-07-24, às 23 11 43

macOS, m1

### LINKED ISSUE #1046: bug(chat): Files will persist when switching chats
**Issue:**
If you attach a file to send in chat, then switch chats the file will persist 

**OS:**
all

**Screen recording:**

https://github.com/Satellite-im/Uplink/assets/93608357/242fd1d6-a1e2-4544-868e-72707fbffada

lgmarchi:
### 4. Improve UI when attach files 
a. Let files attached in vertical now, with document icon with extension or thumbnail if file is image 
image

b. When files in chat, they are aligned now, with document and extension when they do not have thumbnail
image

c. Files aligned when they are loading to be sent as well

image

- Resolve #1026 
- Resolve #1046 
- Resolve #1040

### DIFF
--- a/common/src/state/mod.rs
+++ b/common/src/state/mod.rs
@@ -857,7 +857,7 @@ impl State {
 
     fn clear_chat_attachments(&mut self, chat_id: &Uuid) {
         if let Some(c) = self.chats.all.get_mut(chat_id) {
-            c.files_attached_to_send = Vec::new();
+            c.files_attached_to_send.clear();
         }
     }
 
--- a/common/src/state/mod.rs
+++ b/common/src/state/mod.rs
@@ -233,6 +233,10 @@ impl State {
             }
             Action::SetChatDraft(chat_id, value) => self.set_chat_draft(&chat_id, value),
             Action::ClearChatDraft(chat_id) => self.clear_chat_draft(&chat_id),
+            Action::SetChatAttachments(chat_id, value) => {
+                self.set_chat_attachments(&chat_id, value)
+            }
+            Action::ClearChatAttachments(chat_id) => self.clear_chat_attachments(&chat_id),
             Action::AddReaction(_, _, _) => todo!(),
             Action::RemoveReaction(_, _, _) => todo!(),
             Action::MockSend(id, msg) => {
@@ -851,6 +855,12 @@ impl State {
         }
     }
 
+    fn clear_chat_attachments(&mut self, chat_id: &Uuid) {
+        if let Some(c) = self.chats.all.get_mut(chat_id) {
+            c.files_attached_to_send = Vec::new();
+        }
+    }
+
     /// Clear unreads  within a given chat on `State` struct.
     ///
     /// # Arguments
@@ -1028,6 +1038,12 @@ impl State {
     ) {
         if let Some(chat) = self.chats.all.get_mut(&conv_id) {
             chat.remove_pending_msg(msg, attachments, uuid);
+        }
+    }
+
+    fn set_chat_attachments(&mut self, chat_id: &Uuid, value: Vec<PathBuf>) {
+        if let Some(c) = self.chats.all.get_mut(chat_id) {
+            c.files_attached_to_send = value;
         }
     }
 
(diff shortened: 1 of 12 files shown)


dariusc93:
I wonder if we could just handle the from error instead since this will likely change in the near future where there can be more than 8 files

lgmarchi:
> I wonder if we could just handle the from error instead since this will likely change in the near future where there can be more than 8 files

I was thinking about it as well, is not better to already five a feedback for user before he tries to send message? 

He will already know, maximum possible is 8, and we are going let user just attach 8 files per time as well.

stavares843 (commented):
the toast isn't very visible because user is looking at the bottom of the chat and the toast is in the top right corner maybe we could use this type of error instead

Captura de ecrã 2023-07-25, às 23 01 39

instead of that would say ''Maximum of 8 files can be sent in a single message.''

phillsatellite:
Wanted to also note

on MacOS and Linux I would get this panic each time I sent over 8 files in chat 

Screenshot 2023-07-26 at 6 14 12 AM

And when I tried to compile on Windows I would get this error and app wouldn't open 

image

lgmarchi:
> Wanted to also note
> 
> on MacOS and Linux I would get this panic each time I sent over 8 files in chat
> 
> Screenshot 2023-07-26 at 6 14 12 AM
> And when I tried to compile on Windows I would get this error and app wouldn't open
> 
> image

Which method did you use to attach files? Drag and Drop, File Explorer ou Paste from clipboard? 

And this panic happened when did you click in send files button?

stavares843 (approved):

phillsatellite (approved):

kit/src/components/embeds/file_embed/mod.rs
```diff
@@ -244,3 +300,39 @@ pub fn FileEmbed<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
         }
     ))
 }
+
+fn get_file_thumbnail_if_is_image(filepath: PathBuf, filename: String) -> String {
```
dariusc93, line 304: This shouldnt be needed since the type and thumbnail is determined from within constellation unless there is some other use for it (see https://github.com/Satellite-im/Uplink/pull/828)

lgmarchi: It is necessary, because is using before send messages, using thumbnail on files attachments

stavares843:
missing linter 🔨 
https://github.com/Satellite-im/Uplink/actions/runs/5684449306/job/15407188943?pr=1045

kit/src/components/embeds/file_embed/mod.rs
dariusc93: I will allow it for now but might want to rethink this for the future as this is moreso redundant as it might misrepresent the thumbnail that is actually generated. A couple of things to note of would be in the PR mentioned above. A workaround might be to upload the images directly to constellation and change the `Location` to `Location::Constellation` and point to the file itself when calling `RayGun::attach`.

common/src/state/mod.rs
```diff
@@ -851,6 +855,12 @@ impl State {
         }
     }
 
+    fn clear_chat_attachments(&mut self, chat_id: &Uuid) {
+        if let Some(c) = self.chats.all.get_mut(chat_id) {
+            c.files_attached_to_send = Vec::new();
```
dariusc93:
```suggestion
            c.files_attached_to_send.clear();
```

dariusc93 (approved):
Just one small suggestion and it can merge

PR #1053, opened 2023-07-27
fix(messages): show past messages in correct order

### LINKED ISSUE #1024: bug(chat): Messages are not coming in, in order
**Issue:** 
If you notice when you log into the app and scroll through some of your chats the messages dont always appear as they should. Example in the video shows some of the messages Jeff and I sent at the bottom there are messages sent from 2 days ago, then in the middle it says 3 days ago, and in the very top its back to 2 days ago. Im also attaching the video Sheldon had as well 

**OS:**
All

**Screen recording:**

https://github.com/Satellite-im/Uplink/assets/93608357/e660a4d6-776c-404a-9ed4-eefbae6405e1

https://github.com/Satellite-im/Uplink/assets/93608357/e23f5b0e-22a5-4112-8826-b78534281584

sdwoodbury:
### What this PR does 📖

- fixes the lazy loading of messages. Chats are still only initialized with a few messages at first but now as more messages are fetched, they are displayed in order. 
- This fix isn't as lazy as it could be, but doing it this way prevents bugs arising from extra events getting triggered. 
- This may be good for a temporary solution, until we figure out how to determine when the user has scrolled to the end of a container. Without that, infinite scrolling will be hacky at best. 

- Resolve #1024

### DIFF
--- a/common/src/warp_runner/ui_adapter/mod.rs
+++ b/common/src/warp_runner/ui_adapter/mod.rs
@@ -154,15 +154,17 @@ pub async fn dids_to_identity(
 pub async fn fetch_messages_from_chat(
     conv_id: Uuid,
     messaging: &mut super::Messaging,
-    to_skip: usize,
     to_take: usize,
 ) -> Result<Vec<Message>, Error> {
-    let total = to_take + to_skip;
-    let max_messages = messaging.get_message_count(conv_id).await?;
-    let to_skip = std::cmp::min(to_skip, max_messages);
-    let total = std::cmp::min(total, max_messages);
+    let total_messages = messaging.get_message_count(conv_id).await?;
+    let to_take = std::cmp::min(total_messages, to_take);
+    let to_skip = total_messages.saturating_sub(to_take + 1);
+
     let messages = messaging
-        .get_messages(conv_id, MessageOptions::default().set_range(to_skip..total))
+        .get_messages(
+            conv_id,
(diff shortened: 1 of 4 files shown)


dariusc93 (approved):
LGTM, though it makes me wonder if we should switch over to using `MessagePage` instead and as one scroll up (or down from the current page of messages) that it would get the next (or previous) set.

sdwoodbury:
> LGTM, though it makes me wonder if we should switch over to using `MessagePage` instead and as one scroll up (or down from the current page of messages) that it would get the next (or previous) set.

I'd like that. I'm hoping we could use an approach like that once I figure out how to handle scrolling in Dioxus.

stavares843 (changes requested):
ignore

stavares843 (approved):

common/src/warp_runner/ui_adapter/mod.rs
```diff
@@ -154,38 +154,52 @@ pub async fn dids_to_identity(
...
     .await;
 
-    Ok(messages)
+    let has_more = to_skip > 0;
+    // log::debug!(
+    //     "fetched messages. total: {}, num taken: {}, has_more: {}",
+    //     total_messages,
+    //     to_take,
+    //     has_more
+    // );
```
stavares843, line 185: do we need this part? 🔨

phillsatellite (approved):

stavares843:
added a minor comment but merging

PR #1067, opened 2023-08-02
fix(ImageFrame): Fix rectify frames in images

### LINKED ISSUE #1062: bug(chat): rectify frame in images
STR:
- go to chat
- send image
- appears like this

image

macOS, m1

2 things here:
- should have frame in the left side as well
- the frame around the image should be all the same size

lgmarchi:
### 1. Fix rectify frames in images
image

- Resolve #1062


stavares843 (approved):

sdwoodbury (approved):
approved though I recommend renaming `is_file_extension_empty` to `file_extension_is_empty`, which would make the variable name like a declaration instead of a question.

phillsatellite (approved):

lgmarchi:
> approved though I recommend renaming `is_file_extension_empty` to `file_extension_is_empty`, which would make the variable name like a declaration instead of a question.

I do that way because I always think like a question in conditionals to take an action later, but no problem, I will change here!

PR #1070, opened 2023-08-02
chore(groupchat): topbar groupchat stuff

### LINKED ISSUE #851: chore(groups): need UI update for Group Chats
Group chats can have a single person or multiple people.

Here is what a group chat looks like for multiple (3) people:
image

Here is a group with 1 other person (the default group name will be the username unless you manually type something like i did here):
image

Here is a DM:
image

We need to make it easier to distinguish when you are in a group, and when you are in a DM, which is especially an issue when you are in a group that is just two people.

WanderingHogan:
### What this PR does 📖
Changes around the top bar and the group edit controls to condense things a little bit. Made button to show group list instead of clickable area. Moved some functionality from edit_group to compose where the topbar of the chat lives.

not owner of group
Screenshot 2023-08-02 at 11 39 33 AM

Screenshot 2023-08-02 at 11 40 29 AM

owner of group
Screenshot 2023-08-02 at 11 41 35 AM

Screenshot 2023-08-02 at 11 41 46 AM

Screenshot 2023-08-02 at 11 42 00 AM

- Resolve #851 - (ONLY PART) need to add a profile page still


sdwoodbury (commented):
I think it would be helpful to split up the `friend-list vertically-scrollable` div into multiple functions (as it was before) because otherwise that part of the `rsx!` macro is 70 lines of code, the indentation takes up half the screen, and  the `EditGroup` function is already enormous.

phillsatellite:
add changes label for Stu's comment above

ui/src/components/chat/compose/mod.rs
```diff
@@ -547,6 +572,36 @@ fn get_topbar_children(cx: Scope<ComposeProps>) -> Element {
         get_local_text("uplink.members"),
         all_participants.len()
     );
+
+    let conv_id = state.read().get_active_chat().unwrap().id;
```
lgmarchi: It is better to avoid unwrap() here. Maybe return if unwrap() fails;

ui/src/components/chat/edit_group.rs
```diff
@@ -33,28 +29,21 @@ enum EditGroupAction {
...
 }
 
-#[derive(Props)]
-pub struct Props<'a> {
-    onedit: EventHandler<'a, MouseEvent>,
-}
+// #[derive(Props)]
+// pub struct Props<'a> {
+//     onedit: EventHandler<'a, MouseEvent>,
+// }
```
lgmarchi: Could we remove this comment?

ui/src/components/chat/edit_group.rs
```diff
@@ -116,312 +105,149 @@ pub fn EditGroup<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
...
+                                                    } else {
+                                                        get_local_text("uplink.remove")
+                                                    },
+                                                    onpress: move |_| {
+                                                        let friend_did = friend.did_key();
+                                                        let mut friends = selected_friends.get().clone();
+                                                        friends.clear();
+                                                        selected_friends.set(friends.clone());
+                                                        friends.insert(friend_did);
+                                                        selected_friends.set(friends);
```
lgmarchi: If I am not wrong, is not better to use just one line like this:

`selected_friends.set(vec![friend.did_key()]);`

What do you think?

common/locales/en-US/main.ftl
```diff
@@ -2,6 +2,8 @@ uplink = Uplink
     .unknown = Unknown
     .home = Home
     .add = Add
+    .add-member = Add Members
+    .current-member = Current Members
```
lgmarchi: I believe is better name to use in plural like its value. 

```
.add-members
.current-members
```

ui/src/components/chat/group_users.rs
```diff
@@ -18,6 +18,7 @@ use warp::{crypto::DID, logging::tracing::log};
 pub struct Props {
     #[props(!optional)]
     active_chat: Option<Chat>,
+    is_owner: bool,
```
lgmarchi: My friend, Will this value be used yet? I saw that it is not been used on this page

WanderingHogan: removed, thank you

ui/src/components/chat/edit_group.rs
WanderingHogan: great idea, this lets me remove 4 other lines.

had to be `selected_friends.set(vec![friend.did_key()].into_iter().collect());` because it's a hashset

ui/src/components/chat/compose/mod.rs
WanderingHogan: good call, I realized i didn't even need to go to state here though, and could just use the data prop that gets passed in

ui/src/components/chat/compose/mod.rs
```diff
@@ -547,6 +570,37 @@ fn get_topbar_children(cx: Scope<ComposeProps>) -> Element {
         get_local_text("uplink.members"),
         all_participants.len()
     );
+
+    let conv_id = data.active_chat.id.clone();
```
lgmarchi: Just to avoid clippy, we can remove clone() here. 


```suggestion
    let conv_id = data.active_chat.id;
```

lgmarchi (approved):
Good one! 

Just left some suggestions.

WanderingHogan:
@stu  I split `friend-list vertically-scrollable` back into it's own function per your request

phillsatellite (approved):

tooshel (approved):

stavares843 (approved):

PR #1076, opened 2023-08-03
chore(cargo): update dependency

sdwoodbury:
#### Blink features:
- record call to mp4 file
- auto-mute: eliminates feedback from speakers to microphone. 
- volume control: make a peer sound louder/quieter
- use forks of opus related crates to allow proper cross compiling, needed for mobile builds

#### Other features:
- Multipass::import_identity

- Resolve #

### Additional comments 🎤
`MultiPass::import_identity` does not have any functionality at this time, but will in the future

### DIFF
--- a/common/src/state/mod.rs
+++ b/common/src/state/mod.rs
@@ -604,6 +604,11 @@ impl State {
                 // todo
                 log::info!("audio degradation for peer {}", peer_id);
             }
+            BlinkEventKind::AudioOutputDeviceNoLongerAvailable
+            | BlinkEventKind::AudioInputDeviceNoLongerAvailable => {
+                // todo: notify user
+                log::info!("audio I/O device no longer available");
+            }
         }
     }
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -623,7 +623,7 @@ dependencies = [
 [[package]]
 name = "audiopus_sys"
 version = "0.2.0"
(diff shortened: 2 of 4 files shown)


common/src/state/mod.rs
```diff
@@ -604,6 +604,11 @@ impl State {
                 // todo
                 log::info!("audio degradation for peer {}", peer_id);
             }
+            BlinkEventKind::AudioOutputDeviceNoLongerAvailable
+            | BlinkEventKind::AudioInputDeviceNoLongerAvailable => {
+                // todo: notify user
+                log::info!("audio I/O device no longer available");
+            }
```
lgmarchi, line 611: Is it not better to let already 2 separate error treatment here?

lgmarchi (approved):

common/src/state/mod.rs
sdwoodbury: yes - not better :grin: After all, the behavior is the same as far as the user is concerned and the task of handling those events is planned to be implemented in a separate PR.

dariusc93 (approved):
I did update warp to revert the initialization changes (already updated the commits here) but everything else should be good to go.

phillsatellite:
Getting this error when trying to compile on windows, no issues when trying to compile on MacOS or Linux 

image

stavares843 (approved):

phillsatellite (approved):

PR #1121, opened 2023-08-11
feat(chat): Pinning messages

### LINKED ISSUE #1017: feat(chat): Pinning Messages
Implement the ability to pin/unpin messages in chat. This can be done by using `RayGun::pin`, with `PinState::Pin` to pin a message, and `PinState::Unpin` to unpin a message. 

Note:
- ~~At the time of this issue, there is no way to filter out what messages were pinned in the current implementation. This come a little later (implemented in a separate branch)~~ This can be done using `MessageOption::set_pinned`, which would only yield messages that are pinned.
- There is no hard limit (yet) on how many messages can be pinned in a chat, however, I would recommend limiting it to no more than 100 pinned messages per chat (again, no hard limit internally but may have one in the near future since pinned messages are mostly for important messages one may wish to remember)
- At this time, pinning a message will also pin it for the member(s) of the chat with no localized pinning (will be implemented a little later but although its a simple implementation, at this time its not priority)

Flemmli97:
### What this PR does 📖

- Adds the ability to pin messages in chat
  - Messages can be pinned from the context menu (right click) and viewed in the top bar by clicking on the pinned message button

- Resolve #1017


stavares843 (changes requested):
i pinned some messages but clicking in the pin icon doesn't open the pinned messages


https://github.com/Satellite-im/Uplink/assets/29093946/93cc0c3b-2b5a-400d-9636-7a5eb93d93fd

macOS, m1

stavares843:
Captura de ecrã 2023-08-11, às 19 22 23

linter is failing 🔨

phillsatellite:
Just a few things from my side that I found on MacOS/Linux/Windows

1. Maybe we can have the "There are no Pinned messages in this chat" text be white instead of black. Its a little hard to see when User is in Dark mode 

Screenshot 2023-08-11 at 2 11 22 PM

2.The download button showing when you pinned a message with file attachment is a little odd, maybe for pinned messages we dont show download button

Screenshot_2023-08-11_at_2 13 44_PM

3. When you click on the Pinned Messages icon, it should be highlighted like the Favorites/Edit Group buttons 

https://github.com/Satellite-im/Uplink/assets/93608357/3bcbbe7c-183b-4446-a6ae-ce4d282728ec

4. If you change theme from Dark to Light or Light to Dark then try to open Pinned Messages, it will not open (similar to Sara's problem how box would never open for her)

https://github.com/Satellite-im/Uplink/assets/93608357/0cfbb0aa-4e6c-47b5-a253-83c147248ace

5. Maybe we could add this in a future PR but I think it would be nice if we had a indictor visible in chat to show a message is pinned 

https://github.com/Satellite-im/Uplink/assets/93608357/59c837de-7f25-41b5-88df-5cd3ccb3facb


This is such an awesome start though! Thanks @Flemmli97 🔥

dariusc93 (approved):

Flemmli97:
all points should be addressed now.
pinned message not opening should also be fixed.

common/src/state/mod.rs
```diff
@@ -973,6 +987,20 @@ impl State {
             }
             message.set_reactions(reactions);
             msg.inner = message;
+            if pinned_change {
```
sdwoodbury: I believe the logic here is backwards. if `message` is pinned then it should be added to `conv.pinned_messages` but on line 993 `.retain()` is being used to remove messages from `pinned_messages`. 


I believe you also may not need to call `.filter()` on `conv.messages` (line 999). I think it should be enough to either add or remove the message from `conv.pinned_messages`. 

in `MessageEvent::Deleted` (line 497), need to check if the message was pinned and if so, update `conv.pinned_messages`

common/src/warp_runner/manager/commands/raygun_commands.rs
```diff
@@ -209,7 +227,28 @@ pub async fn handle_raygun_cmd(
...
+            conv_id,
+            to_fetch,
+            current_len,
+            rsp,
+        } => {
+            let r = fetch_messages_from_chat(
+                conv_id,
+                messaging,
+                MessageOptions::default().set_pinned(),
+                to_fetch + current_len,
```
sdwoodbury: this is probably wrong. looks like `FetchMessages` was copy/pasted, but that command doesn't fetch all messages in the conversation. I'm guessing that `FetchPinnedMessages` actually wants to fetch all pinned messages.

common/src/warp_runner/ui_adapter/message_event.rs
```diff
@@ -147,6 +149,18 @@ pub async fn convert_message_event(
         } => MessageEvent::ConversationNameUpdated {
             conversation: messaging.get_conversation(conversation_id).await?,
         },
+        MessageEventKind::MessagePinned {
```
sdwoodbury, line 154: I'd prefer the new enumerations be `MessageEventKind::Pinned` and `MessageEventKind::Unpinned` or  `MessageEventKind::MessagePinned` and `MessageEventKind::MessageUnpinned`. The latter is redundant but the `MessageReactionRemoved/Added` events have that format too. (maybe they should be changed too)

common/src/warp_runner/ui_adapter/mod.rs
```diff
@@ -154,17 +154,15 @@ pub async fn dids_to_identity(
 pub async fn fetch_messages_from_chat(
     conv_id: Uuid,
     messaging: &mut super::Messaging,
+    options: MessageOptions,
```
sdwoodbury: It may be confusing to have a function that accepts a `MessageOptions` which lets the caller specify a range, and then override that range (on line 165). It may be better to either calculate the `MessageOptions` before calling `fetch_messages_from_chat` (and remove the `to_take` argument, or make a new function to fetch pinned messages. 

You probably want a new function to fetch all pinned messages because normal messages will be paged and pinned messages will probably be fetched all at once.

common/src/warp_runner/ui_adapter/mod.rs
```diff
@@ -211,6 +209,12 @@ pub async fn conversation_to_chat(
     .collect()
     .await;
 
+    let pinned: VecDeque<_> = messages
+        .iter()
+        .cloned()
```
sdwoodbury: you probably want to call `.cloned()` after `.filter()` so that you don't clone every message in the conversation just to throw it away.

ui/src/components/chat/pinned_messages/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+    #[props(!optional)]
+    active_chat: Option<Chat>,
+}
+
+#[allow(non_snake_case)]
+pub fn PinnedMessages(cx: Scope<Props>) -> Element {
+    log::trace!("rendering pinned_messages");
+    let state = use_shared_state::<State>(cx)?;
+    let _loading = use_state(cx, || true);
+    let newely_fetched_messages: &UseRef<Option<(Uuid, Vec<ui_adapter::Message>, bool)>> =
```
sdwoodbury: I'm not sure that `PinnedMessages` needs to worry about `newely_fetched_messages`. I believe this and the related code was erroneously copied from `compose/messages.rs`

sdwoodbury (changes requested):

phillsatellite:
Hey @Flemmli97 thanks for adding all those changes! Works a lot better now! One last thing from QA then it looks perfect

If you edit message that is already Pinned, the new edited message will not update in Pinned Messages 

https://github.com/Satellite-im/Uplink/assets/93608357/03efb075-fb35-4dc4-a983-8de29b0c0b73

common/src/state/mod.rs
Flemmli97: ye retain is correct. if the updated message isnt pinned anymore it gets removed.
adding and removing does not work as else order is not kept.

Flemmli97: delete is fixed

common/src/warp_runner/ui_adapter/message_event.rs
Flemmli97: idk about that. since every other enum value follows the same principle. or do you mean `MessageEvent` instead of `MessageEventKind`?

sdwoodbury: I stand by what I wrote. Including the part about the other enums following the same naming pattern.

common/src/state/mod.rs
sdwoodbury: i read lines 1000-1001 as "if incoming message is pinned and the pin state has changed..."  But I see that was a mistake on my part due to line 999. 

~~I'm confused about where `pinned_messages` gets updated. I'll keep looking for that.~~ 

Now that I think about it, that entire `if` statement from lines 1001 to 1012 could be replaced by lines 1005 to 1011. That would eliminate one possible source of confusion. 

It would probably be less confusing if line 999 was done away with. It looks like `msg` isn't assigned to anything and I think it would be more clear if `message` was used to update `conv.pinned_messages`.

common/src/warp_runner/ui_adapter/message_event.rs
Flemmli97: the thing is though for `MessageEventKind` all commands that deal with message have `Message` prefixed.
`MessageSent, MessageReceived ... etc`. so should they all be changed?

should i then update the other enums also? or should that be in a separate pr?

sdwoodbury: I think it would be better to rename them to remove the `Message` prefix.  Adding that there was my mistake. Probably should be done in a separate PR just to make this PR easier to read.

phillsatellite (approved):

phillsatellite:
Good by me, waiting for Sara to put QA tested in case she wants to test as well 👍

dariusc93:
Hold merging pending @sdwoodbury review

common/src/warp_runner/ui_adapter/message_event.rs
dariusc93: It can be done in a separate PR for now if it havent been done in this PR. I remember the reasoning for me having the `Message` prefix in the enum was due to some conflicts earlier on when streams were introduced, which looking over the code isnt applicable now so it may change in the future on warp end but not exactly priority.

common/src/state/mod.rs
Flemmli97: actually line 999 is not for pinned things. `update_message` was originally only used for when reactions are updated. i kinda just hijacked the function and added the pin update handling there since else i would need to copy all the things into a new function.

common/src/warp_runner/manager/commands/raygun_commands.rs
Flemmli97: im just wondering here if we want to fetch all pinned message at once? since if someone has a lot of pinned messages (for whatever reason) this might take a while

should we limit pinned messages to 100 max as darius suggested in the linked issue to maybe mitigate this?

common/src/warp_runner/ui_adapter/message_event.rs
Flemmli97: ~~i will then for now adjust the enum related to this pr. and the other enums can be changed in another pr~~
just noticed but `MessageEventKind` is from warp not uplink. but i did change `MessageEvent:MessagePinnedUpdate`

ui/src/components/chat/pinned_messages/mod.rs
Flemmli97: see comment https://github.com/Satellite-im/Uplink/pull/1121#discussion_r1294466623

common/src/state/mod.rs
sdwoodbury: true, and `msg` is a `mut` reference to `conv.messages`.

sdwoodbury:
@Flemmli97  I pushed some commits that I think improve the code. It isn't ready to merge as-is because there's still the question of if and how to handle pagination. I did this because it was easier to show the changes instead of try to explain them.

sdwoodbury:
as for the question of pagination, I think we should wait for dioxus 0.4 to do that. It would be a good bit of work to implement the hacky mouseover based fetching mechanism and that work would then get thrown away anyway after the dioxus update.

common/src/warp_runner/manager/commands/raygun_commands.rs
sdwoodbury: that could be fine for a temporary solution. 

i'm not entirely sure about this but we may want to show pinned messages in order of most recent to  least recent too. Currently the oldest message appears at the top of the list and  one has to scroll down for more recent messages.

common/src/state/mod.rs
Flemmli97: hmm wondering then if this can be closed as resolved?

ui/src/components/chat/pinned_messages/mod.rs
Flemmli97: removed that whole part as pinned msg are now fetched at once

common/src/warp_runner/ui_adapter/mod.rs
Flemmli97: i split it now into a separate function `fetch_pinned_messages_from_chat` (in case this is needed in future) and rolled back `fetch_messages_from_chat` to how it is in dev

common/src/state/mod.rs
Flemmli97: closing it then. seems its addressed

common/src/warp_runner/manager/commands/raygun_commands.rs
Flemmli97: app will fetch up to 100 pinned message now and will have them in reverse order

Flemmli97:
For now i limited pinned messages to 100 messages. If the user reaches that a toast notification will be displayed when the user attempts to pin more

Flemmli97:
just a notice with unloaded messages:
if you click on a pinned message that is atm not loaded in the current chat it will simply go to the oldest current loaded message

sdwoodbury (changes requested):
I like the changes! About the case where someone wants to jump to a pinned message that hasn't been loaded - I think that just scrolling to the top wouldn't show the pinned message and it's probably better to not add a feature that doesn't work. I think it better to first load the necessary messages (plus some extra so that the subsequent scroll doesn't just trigger another reload) and then scroll to the pinned message.

Flemmli97:
implemented scrolling into unloaded message. if a message is not loaded it will try to load all messages till that point and then scroll to it. this might potentially cause lots of lag if you want to scroll to a message that is really far back. 
maybe this can be a point in a future issue though?

stavares843:
Captura de ecrã 2023-08-18, às 16 35 59


linter is failing 🔨

sdwoodbury (approved):

phillsatellite (approved):

stavares843 (approved):

PR #1139, opened 2023-08-16
chore(cargo): Update Dependency 

### LINKED ISSUE #1134: chore(blink): blink not always initialized correctly
cargo run -- --path=/Users/<profile>/.uplink_test1
cargo run -- --path=/Users/<profile>/.uplink_test2

make two new accounts with those names test1 and test2

In both enable developer tools

try to call from one to the other, if it works, great

if it doesn't, stop and restart just the uplinks with the red blink initialization error in devtools
usually one works and one doesn't for me the first start, then it takes 1 or 2 restarts of one for the second to work

dariusc93:
### What this PR does 📖

- Relocate logic in initialization

- Resolves #1134

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9168,7 +9168,7 @@ dependencies = [
 [[package]]
 name = "warp"
 version = "0.1.0"
-source = "git+https://github.com/Satellite-im/Warp?rev=f6f48c42293396f9db8cbdd07283f092701b9188#f6f48c42293396f9db8cbdd07283f092701b9188"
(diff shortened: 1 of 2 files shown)


WanderingHogan:
this rocks, this makes the pieces of warp that don't load retry and load after a few moments. It fixes this ticket: https://github.com/Satellite-im/Uplink/issues/1134

We will probably want to find a way to visually show if warp has loaded or not but this is great!

sdwoodbury (approved):

phillsatellite (approved):

phillsatellite:
Good by me, do not merge until Sara also approves it when she gets back from dinner and is able to finish her testing

stavares843 (approved):

PR #1140, opened 2023-08-17
fix(extensions): Fix auto enable extension and emoji selector not working

Flemmli97:
### What this PR does 📖

- Fixes the extension setting "Automatically enable extensions" not working
- Fixes emoji extension not enabled by default

- In dev since the emoji extension is not bundled in it will only be enabled if you copy paste the extension into the extension folder.
- In release this is not needed


stavares843 (commented):
i used cargo run --release and this button doesn't work

Captura de ecrã 2023-08-17, às 16 32 28


macOS

Flemmli97:
idk if that is different but i tested it by building the app and using the Uplink.app.

tooshel:
> idk if that is different but i tested it by building the app and using the Uplink.app.

Is there something about building that enables the extension? If the extension was disabled at one point, deleted, and then put back will it be saved in state.json as disabled so when I copy the extension there again it will be enabled? 

I guess what I'm really asking is . . . is there a way to test this without having to build and install the app?

Flemmli97:
not sure. but im guessing that simply running the app does not have the extension in the right place. e.g. in the Uplink.app the extension is in it (in the `Frameworks` folder).

one way to test auto activation would be to copy paste the extension into the `uplink/extensions` folder which should auto enable it. so meaning if the extension is in the right place it should auto enable

tooshel:
I tried to simulate being a new user and even then the extension wasn't enabled by default and I think it's because of this line: 

```if self.configuration.extensions.enable_automatically```

it will only enable this particular extension if some other option that is normally false is actually true. Do I understand that right?

Flemmli97:
was the extension in the right folder?

tooshel:
> was the extension in the right folder?

I think so or else it would not have showed up right!  I created a new folder, .uplink_e22, and then made a new folder in there called extensions and then I copied the extension there like this.

image

When I ran the app, I saw this: 
image

Other times when I made new accounts and didn't copy it right (i.e. I copied it to an extensions file because the folder wasn't there) the extension just wasn't there. That said, it could be that extensions can exist somewhere beside the .uplink folder.

Flemmli97:
can you try testing it with an Uplink.app so we can test if at least that works?

stavares843:
asked hogan to make a .app and i tried here
https://github.com/Satellite-im/test-builds-uplink/releases/tag/0.1.7

but was the same

Flemmli97:
Fixed the problem with initial extension registration not respecting setting etc.
`cargo run --release` will still not work though since the extension is not put in the right place

sdwoodbury (commented):
The code accomplishes the task. Still, I wonder if we really want to do this. With this change, anyone can name their extension `emoji_selector` and it will be automatically enabled.

tooshel:
> The code accomplishes the task. Still, I wonder if we really want to do this. With this change, anyone can name their extension `emoji_selector` and it will be automatically enabled.

That seems low risk in that anyone that would install an extension would also probably want to enable it. Unless of course it was someone trying to get another person to do something malicious. Also, I couldn't get it to work! Did something change since last week @phillsatellite and @stavares843?

Flemmli97:
> The code accomplishes the task. Still, I wonder if we really want to do this. With this change, anyone can name their extension `emoji_selector` and it will be automatically enabled.

i dont think this will be a big issue since the extension is bundled with the app by default. unless users can override extensions that are already installed?

Flemmli97:
> > The code accomplishes the task. Still, I wonder if we really want to do this. With this change, anyone can name their extension `emoji_selector` and it will be automatically enabled.
> 
> That seems low risk in that anyone that would install an extension would also probably want to enable it. Unless of course it was someone trying to get another person to do something malicious. Also, I couldn't get it to work! Did something change since last week @phillsatellite and @stavares843?

how did you try to test it?

tooshel:
I tested it like this . . . . 
First, I built the project or otherwise just make sure the emoji_selector.dll file exists
cargo build --release

Create a new folder  C:\uplinkprofiles\user1_test_ext\
And then another folder C:\uplinkprofiles\user1_test_ext\extensions

And then I put the extension into that folder:
cp .\target\release\emoji_selector.dll C:\uplinkprofiles\user1_test_ext\extensions

then I ran the project using that user
cargo run --profile=release -- --path C:\uplinkprofiles\user1_test_ext trace2

And I get this:
image

And of course, NOW it works. But when I did that the last time it didn't.

I dug in and just discovered that in an INSTALLED app, the extension goes in c:\program files\uplink\extensions and thought maybe THAT was an issue? But apparently not.

tooshel:
@Flemmli97 I also noticed that after I disable it manually, when I launch the app again, it's enabled again. Is that what we expect?

phillsatellite:
> @Flemmli97 I also noticed that after I disable it manually, when I launch the app again, it's enabled again. Is that what we expect?

Also replicated on my side with m2 macOS

Flemmli97:
fixed

phillsatellite (approved):

stavares843 (approved):

sdwoodbury (changes requested):
I don't like that `UplinkExtension.is_enabled` is used as a default value by `State.ui.extensions.insert`. I think that function should be given a 3rd argument for the default value (when the extension name isn't found in the hash map) and that the field `is_enabled` should be removed from `UplinkExtension`. 

I believe that this change would make the code easier to follow. I was initially confused because when I saw `ext.set_enabled(true)`, I expected that the extension would be enabled. But in this case it isn't - that really just sets the default value. 

The presence of `is_enabled` in `UplinkExtension` also requires synchronization between `state.ui.extensions.enabled.is_enabled` and `state.ui.extensions.map`, as seen in `Extensions.insert()`. This is wasted effort because once the extension is registered, `UplinkExtension.is_enabled` is no longer needed. 

Affected areas:  
image
image
image

sdwoodbury (approved):

phillsatellite (approved):

PR #1141, opened 2023-08-17
fix(emoji): Consistent emoji in context menu

### LINKED ISSUE #1115: task(chats): Emoji reactions from context menu are constantly changing on new accounts
When you right click on a message, you will see a context menu displaying the 4 most used emoji reactions by the user. 

However, when a new account is created, there are 5 emojis assigned by default with counter = 1 in state.json file, but the app only has 4 UI buttons for the top reactions on the context menu, and the 4 emojis displayed there are constantly changing between new accounts created.

image

Steps to reproduce:

1. Create several tests accounts and add friends to them
2. Go to chat on these accounts and send a message
3. Right click on the message to see the context menu
4. The context menu emoji reactions displayed (top used emoji reactions) are constantly changing on new accounts.

Note: This is issue was discussed during today's standup meeting and it was suggested to use different default counter values for each of the emojis listed initially when creating accounts. Therefore, the emojis would be sorted correctly by the number given in this counter.

Flemmli97:
### What this PR does 📖

- Makes emoji list in context menu consistent across all accounts if use count is equal

- Resolve #1115


stavares843:
https://github.com/Satellite-im/Uplink/assets/29093946/37c36f1a-0c61-4fbc-9b66-b3550dd16f08

phil noticed this issue while testing, where the order is somewhat random in the emoji usage

Flemmli97:
it will show the most used emoji first in that list. i did fix the issue though where the count starts at 1 causing it to have inconsistency when using a new emoji the first time.

Flemmli97:
i have also expanded the emoji amount in that menu to 5 since the default amount of emoji without extension is 5.
else the user can only use 4 emojis ever (instead of the default 5)

sdwoodbury (approved):

stavares843 (approved):

phillsatellite (approved):

PR #1148, opened 2023-08-21
fix(chat): Add indicator for disabled emoji extension

### LINKED ISSUE #1083: task(reactions): Add indicator showing you need to add emoji extension to use all reactions
**Issue:**
When you dont have emoji's enabled there is no way to use all the emoji's to react to a message, we should add a message or indicator so the user is aware of this and knows to install the extension to react with all emoji's 

**Screenshot:**
image

Flemmli97:
### What this PR does 📖

- Adds an indicator in form of a tooltip upon hovering the emoji button in quickprofile that tells the user the extension is disabled (if true)

- Resolve #1083


sdwoodbury (approved):

stavares843 (commented):
in this case, i don't have the emoji extension in place at all so saying it's deactivated is incorrect because it isn't there at all


Captura de ecrã 2023-08-22, às 14 45 16



Captura de ecrã 2023-08-22, às 14 43 38


maybe we can say that user doesn't have that extension instead

stavares843 (approved):

stavares843:
after discussing with the team, we will merge 🔨

PR #1157, opened 2023-08-23
feat(ui): Improve call UIs

### LINKED ISSUE #1135: chore(calling-ui-siderbar): make it look like figma
current:
Screenshot 2023-08-15 at 11 13 48 AM

screenshot of desired: 

Screenshot 2023-08-15 at 11 14 33 AM

Screenshot 2023-08-15 at 11 14 56 AM

figma link: https://www.figma.com/file/vYKm91y1PjZPW0NRuzj0u9/Video%2F-Voice-Chat?type=design&node-id=0-1&mode=design&t=2el4JshSVnT0WvPU-0

Flemmli97:
### What this PR does 📖

- Adjust various call ui to match the figma in the linked issue
~~- Missing: Call ui if in caller chat~~

- Resolve  #1135

### Note
Some buttons might be different than the figma since functionality isnt there yet.
Video call things are also not implemented as videos as a whole is not there yet


stavares843 (commented):
image

if you go to a group > call > remove people > avatar doesn't get updated in the call

macOS, m1

stavares843 (commented):
if you go to a group where is only 1 person > call > no avatar appears

image


macOS, m1

phillsatellite:
Somethings I noticed

On light mode, should we have the call background be black? Was thinking when we have video the white background might be intrusive  

Screenshot 2023-08-24 at 6 35 43 PM

Also noticed if the user only has 1 chat and they go to a different page the call pop out never appears. This also happens if the user navigates to a different page from the current call chat 

https://github.com/Satellite-im/Uplink/assets/29093946/24b223be-1029-4d1b-9d83-6a88bb82c26e



Also talked with Sara and we thought the call pop out could be a little smaller 

Screenshot 2023-08-24 at 6 39 30 PM

Screenshot 2023-08-24 at 6 38 53 PM

Flemmli97:
> if you go to a group where is only 1 person > call > no avatar appears
> 
> image
> 
> macOS, m1

yes cause it displays all members in the call except oneself

Flemmli97:
> image
> 
> if you go to a group > call > remove people > avatar doesn't get updated in the call
> 
> macOS, m1

ive made it so it gets updated now but while doing so i found some problems with it:
currently if you remove someone from a group while they are in the call it does not terminate the call. regardless on the person who was removed or the person in the call.
so the person removed from the group is still in the call and if the group only has 1 user (the owner e.g.) left the call is still ongoing.

the reverse also has a problem:
if you add someone else to a group while in the call that other person will not be send a call request and thus also is not added to the active call. idk what behavior we want here.

also since those problems are not ui related but functionality related my suggestion is to put this in a new issue with a new pr

stavares843:
even thought the UI added also includes functionality, after discuss with the team, we have decided to fix the issue where you remove/add users in the call for a separate ticket

Flemmli97:
> Somethings I noticed
> 
> On light mode, should we have the call background be black? Was thinking when we have video the white background might be intrusive
> 
> Screenshot 2023-08-24 at 6 35 43 PM
> Also noticed if the user only has 1 chat and they go to a different page the call pop out never appears. This also happens if the user navigates to a different page from the current call chat
> 
>  Screen_Recording_2023-08-24_at_5.48.43_PM.mov 
> 
> Also talked with Sara and we thought the call pop out could be a little smaller
> 
> Screenshot 2023-08-24 at 6 39 30 PM Screenshot 2023-08-24 at 6 38 53 PM

what is take on the background color now? as it seems you have phrased it like a question.

stavares843:
yes @Flemmli97 could you add the same background color as the one we use here - the dark grey in the light mode please?


image

this one

Captura de ecrã 2023-08-25, às 21 49 06

common/src/state/call.rs
```diff
@@ -197,6 +212,12 @@ impl Call {
         }
     }
 
+    fn remove_participants(&mut self, id: DID) {
```
sdwoodbury: the function name should use the singular noun  - `remove_participant` unless it takes a `Vec` or similar.

sdwoodbury (dismissed):
We should probably move `components/calldialog` into `components/media`. 

I found that once both sides join the call, the UI looks different depending on who offered the call.

if they offer and you answer - looks like this wasn't updated to reflect the new UI
image

if you offer and they answer - should probably show both participants but I could be wrong
image

Flemmli97:
> We should probably move `components/calldialog` into `components/media`.
> 
> I found that once both sides join the call, the UI looks different depending on who offered the call.
> 
> if they offer and you answer - looks like this wasn't updated to reflect the new UI image
> 
> if you offer and they answer - should probably show both participants but I could be wrong image

can you show a more concrete example? as for me it doesnt matter who called. the ui looks the same
when getting called:
Screenshot 2023-08-28 at 17 46 02
Screenshot 2023-08-28 at 17 46 09
Screenshot 2023-08-28 at 17 46 17
Screenshot 2023-08-28 at 17 46 23

when calling:
Screenshot 2023-08-28 at 17 46 35
Screenshot 2023-08-28 at 17 46 42
Screenshot 2023-08-28 at 17 46 55
Screenshot 2023-08-28 at 17 47 00

sdwoodbury:
@Flemmli97 I think I tested wrong. The screenshots I showed where it didn't work involved a group call instead of a direct call. When I used direct calls, the UI was the same after both sides joined the call. 

On the bright side, now we know that the UI for group calls hasn't been updated yet :grin:

Flemmli97:
seems to be the same for groups too for me hmm

Flemmli97:
@sdwoodbury asking again here for the above

phillsatellite (approved):

common/src/state/mod.rs
```diff
@@ -585,6 +589,19 @@ impl State {
             }
             MessageEvent::RecipientRemoved { conversation } => {
                 if let Some(chat) = self.chats.all.get_mut(&conversation.id()) {
+                    // Also remove the recipient from the active call if present
```
sdwoodbury: if we do this, we should also remove the participant from all pending calls.

also, and more importantly, this won't actually boot the person from the call, just from the UI. @stavares843 Blink would need a new feature to boot someone from a call. I think that if we really want this, we should make a separate ticket for it and add the support to Blink first.

ui/src/components/media/calling.rs
```diff
@@ -0,0 +1,456 @@
...
+            )),
+            p {
+                class: format_args!("call-time {}", if cx.props.in_chat {"in-chat"} else {""}),
+                format_timestamp_timeago(active_call.answer_time.into(), &state.read().settings.language_id()),
+            }
+        },
+        div {
+            class: "controls",
+            Button {
+                // TODO: we need to add an icon for this `if state.read().ui.silenced { Icon::Microphone } else { Icon::Microphone }`
```
sdwoodbury: the TODO has been addressed. can delete it now.

sdwoodbury (changes requested):

InfamousVague:
> > if you go to a group where is only 1 person > call > no avatar appears
> > image
> > macOS, m1
> 
> yes cause it displays all members in the call except oneself

Maybe we can just display "Nobody is here" or "You're all alone"

stavares843:
> Maybe we can just display "Nobody is here" or "You're all alone"


good idea 🔨

stavares843:
missing addressing stu requested changes 🔨

sdwoodbury:
@Flemmli97  I still think that removing someone from a group shouldn't remove them from the call UI at this time because it wouldn't remove them from the call. Is this OK with @stavares843?

stavares843:
if we can't remove from the call, not much point in just removing from the UI then

Flemmli97:
should i just for now comment out that part so we can just uncomment it later? or delete it

sdwoodbury:
> should i just for now comment out that part so we can just uncomment it later? or delete it

Comment it out

phillsatellite (approved):

phillsatellite:
Functionality seems good when tested across Linux/Mac/Windows

stavares843 (approved):

sdwoodbury (approved):

PR #1179, opened 2023-08-30
feat(crash): Improve crash handling

Flemmli97:
### What this PR does 📖

Currently when a panic occurs uplink will throw the generic message that the app paniced at a line of code. Something like `panic occured at file {}` not showing the actual reason. This makes debugging crashes really hard. It seems that `PanicInfo::payload` does not properly work as it doesn't display the reason.

This PR changes it back to the default print format which does change it to something like `panicked at {reason}, {file}`.
Additionally this PR also includes the feature that the crash it saved to a file.
The file will be saved to `./uplink/crash-logs` and will also include a backtrace leading up to the crash so hopefully this will make debugging easier


ui/src/bootstrap.rs
```diff
@@ -59,18 +62,24 @@ pub(crate) fn use_boostrap<'a>(
...
+        let backtrace = Backtrace::force_capture();
+        let time = Local::now().format("%d-%m-%Y_%H-%M-%S").to_string();
+        let path = STATIC_ARGS.crash_logs.join(format!("{}.log", time));
+        let log = format!(
+            "Uplink crashed:\n{}\nBackTrace:\n{}",
+            crash_report, backtrace
+        );
         println!("{crash_report}");
+        if let Some(p) = path.parent() {
+            if fs::create_dir_all(p).is_ok() {
```
dariusc93: It might be best to do `let _ = fs::create_dir_all(p);` and ignore any errors since `fs::write` would give the reason for why it failed to write in that path. If we want to still check to make sure it does create the directories, then we should catch the error instead by doing something along the lines of 

```rust
if let Err(e) = fs::create_dir_all(p) {
   println!("Unable to create directory: {e}");
   return;
}

match fs::write(&path, log) {
    Ok(_) => println!("The crash has been saved to {}", path.to_string_lossy()),
    Err(e) => println!("Unable to save crash log {}", e),
}
```

ui/src/bootstrap.rs
```diff
@@ -59,18 +62,24 @@ pub(crate) fn use_boostrap<'a>(
...
+                match fs::write(path.clone(), log) {
```
dariusc93: If I recall, `std::fs::write` accepts `impl AsRef<Path>`. If thats the case, we can do `&path` or `path` but since you are also reusing it, the reference should suffice.

dariusc93 (commented):
LTGM but just a couple of comments on the logic.

sdwoodbury (approved):
nice

dariusc93 (approved):

dariusc93:
We can merge this in, but if QA wants to test, a separate branch can be created with inserted panics at some points that could produce a crash report.

tooshel:
I added a panic to ui\src\layouts\chat.rs and the app didn't actually shut down it kept running and whenever I clicked on it a new report would be generated. Where and how is a good place to put a panic? 

image

image

dariusc93:
> I added a panic to ui\src\layouts\chat.rs and the app didn't actually shut down it kept running and whenever I clicked on it a new report would be generated. Where and how is a good place to put a panic?
> 
> image
> 
> image

https://github.com/Satellite-im/Uplink/blob/ce71141ddde39c519fbf6a820fa2221d82b93c5f/ui/src/components/friends/add.rs#L188 might be a good place to try

PR #1193, opened 2023-09-05
feat(chat): Emoji autofill

### LINKED ISSUE #1145: feat(emojis): auto filler list picker
In discord or github or any other apps when you do ` :  joy  :`

Captura de ecrã 2023-08-18, às 17 14 27

appears a list with the picker auto filled, would be neat to add that into Uplink as well

### LINKED ISSUE #1167: bug(emoji): emoji counter not working properly on m2
From Phil:

So I tried on a m1 Mac and a m2 Mac. The m1 Mac worked great but on the M2 I had an issue of the emoji counter wouldn't show count until after you insert the emoji. The count would not clear out again until after you send another message with an emoji 

https://github.com/Satellite-im/Uplink/assets/93608357/f228f8cb-cb22-4a9a-84a5-b2a72e1494c5

Flemmli97:
### What this PR does 📖

- Adds ability to type emoji shortcut to have it either autofill the emoji or open up a selection menu similar to how a lot of other apps handle emojis
- If the user types e.g. `:jo` it will show all emojis starting with `jo`

- Also fixes char counter not updating correctly

- Resolve #1145
- Resolve #1167


native_extensions/emoji_selector/src/lib.rs
```diff
@@ -287,6 +287,34 @@ fn render_1(cx: Scope, _unused: bool) -> Element {
...
+    use_effect(cx, (), |_| {
+        to_owned![state];
+        async move {
+            state.write_silent().ui.emojis.register_emoji_filter(
+                String::from("emoji_picker"),
+                |pattern, exact| {
+                    emojis::Group::iter()
+                        .flat_map(|group| group.emojis())
+                        .filter_map(|emoji| {
+                            if emoji.shortcode().is_none() {
```
sdwoodbury: (not blocking comment): this is a good place to use `and_then`. ex:
```
emoji.shortcode().and_then(|short| {
    if (exact && short.eq(pattern))
        || (!exact && short.starts_with(pattern))
    {
        Some((emoji.to_string(), short.to_string()))
    } else {
        None
    }
})
```

common/src/state/ui.rs
```diff
@@ -52,6 +61,51 @@ impl EmojiCounter {
...
+            .filter_map(|(emoji, alias)| {
+                if (exact && (*alias).eq(pattern)) || (!exact && alias.starts_with(pattern)) {
+                    Some((emoji.to_string(), alias.to_string()))
+                } else {
+                    None
+                }
+            })
+            .collect();
+        for (_, matcher) in self.emoji_filters.iter() {
+            matcher(pattern, exact).iter().for_each(|(emoji, alias)| {
```
sdwoodbury, line 80: will the function `matcher` on line 80 ever be different from the function on lines 71-66? 

I'm wondering if the filter function could/should be built in instead of being a `Fn` trait. 

Also, I think the `matcher` function should have a 3rd argument - the emoji being matched to. Then there isn't a need to make a separate `Fn` closure for each emoji. This would require storing the emoji "short" alongside the unicode (which i also think would be an improvement). 

`EmojiList` could become a struct like this
```
struct EmojiList {
   emojis: HashSet<Emoji>,
   times_used: HashMap<String, u64>
}

struct Emoji {
  unicode: String,
  short: & 'static str, // may want to make this a list, same as the emojis crate
}
```
or perhaps use the `Emoji` struct from the `emojis` crate.

common/src/state/ui.rs
```diff
@@ -52,6 +61,51 @@ impl EmojiCounter {
...
+                matches.insert(emoji.clone(), alias.clone());
+            });
+        }
+        let mut matches: Vec<(String, String)> = matches
+            .iter()
+            .map(|(emoji, alias)| (emoji.clone(), alias.clone()))
+            .collect();
+        matches.sort_by(|(emoji, _), (emoji2, _)| {
+            let first = self.list.get(emoji).unwrap_or(&0_u64);
```
sdwoodbury: the name `list` confused me. I see that `EmojiCounter`'s `list` is really a `HashMap` of integers. This name is bad.

kit/src/elements/textarea/mod.rs
```diff
@@ -100,12 +107,25 @@ pub fn Input<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
         let _ = eval(&update_char_counter_script);
     }
 
-    let cv2 = current_val.clone();
-    let cv3 = current_val.clone();
+    let eval = use_eval(cx);
```
sdwoodbury: this line is redundant due to line 57

kit/src/elements/textarea/mod.rs
```diff
@@ -100,12 +107,25 @@ pub fn Input<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
...
+    let cursor_eval = include_str!("./caret_position.js")
```
sdwoodbury: could be better named. ex: `cursor_script`.

kit/src/elements/textarea/mod.rs
```diff
@@ -100,12 +107,25 @@ pub fn Input<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
...
+            *text_value.write_silent() = val;
+        }
+    });
+
+    let do_cursor_update = oncursor_update.is_some();
 
-    let text_value = Rc::new(RefCell::new(value.to_string()));
-    let text_value_onchange = Rc::clone(&text_value);
-    let text_value_onreturn = Rc::clone(&text_value);
+    if let Some(val) = cursor_position.write_silent().take() {
```
sdwoodbury, line 119: instead of writing to `cursor_position`, re-rendering the textarea, and checking if `cursor_position` is `Some`, you could use a `use_coroutine` and send a command (with the cursor position) over the channel. This approach would require a hook (a `UseRef<bool>`) to ensure one command isn't submitted until the previous one has finished. One would set the hook to `true` before sending the command and the `use_coroutine` would set it back to `false` after the callback.

kit/src/layout/chatbar/mod.rs
```diff
@@ -141,11 +148,44 @@ pub fn Chatbar<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
...
+                        if let Some(e) = cx.props.oncursor_update.as_ref() {
+                            e.call((v,p))
+                        }
+                        *cursor_position.write_silent() = Some(p)
+                    },
                     is_disabled: cx.props.is_disabled,
+                    prevent_up_down_arrows: !cx.props.emoji_suggestions.is_empty(),
+                    onup_down_arrow: 
+                        move |code| {
+                            let current = &mut *selected_emoji.write_silent();
```
sdwoodbury, line 178: you should obtain the `Some` value from `selected_emoji` here, and if `selected_emoji` is `None`, then `return`. basically stop using `unwrap` unless absolutely necessary.

kit/src/layout/chatbar/mod.rs
```diff
@@ -157,7 +197,48 @@ pub fn Chatbar<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
             div {
                 class: "controls",
                 cx.props.controls.as_ref()
-            }
+            },
+            (!cx.props.emoji_suggestions.is_empty()).then(|| rsx!(EmojiSuggesions {
+                suggestions: cx.props.emoji_suggestions,
+                on_emoji_click: move |(emoji, alias)| {
+                    if let Some(e) = cx.props.on_emoji_click.as_ref() {
+                        e.call((emoji, alias, cursor_position.read().unwrap()))
```
sdwoodbury: remove the call to `unwrap()`. `kit` shouldn't panic if  the code using `Chatbar` doesn't correctly call `oncursor_update` before calling `on_emoji_click`

ui/src/components/chat/compose/chatbar.rs
```diff
@@ -438,6 +445,58 @@ pub fn get_chatbar<'a>(cx: &'a Scoped<'a, super::ComposeProps>) -> Element<'a> {
             value: state.read().get_active_chat().as_ref().and_then(|d| d.draft.clone()).unwrap_or_default(),
             onreturn: move |_| submit_fn(),
             extensions: cx.render(rsx!(for node in ext_renders { rsx!(node) })),
+            emoji_suggestions: emoji_suggestions,
+            oncursor_update: move |(mut v, p): (String, i64)| {
+                if let Some(id) = &active_chat_id {
+                    let sub: String = v.chars().take(p.try_into().unwrap()).collect();
```
sdwoodbury: remove the call to `unwrap`

ui/src/components/chat/compose/chatbar.rs
```diff
@@ -438,6 +445,58 @@ pub fn get_chatbar<'a>(cx: &'a Scoped<'a, super::ComposeProps>) -> Element<'a> {
...
+            },
+            on_emoji_click: move |(emoji, _, p): (String, String, i64)| {
+                if let Some(id) = &active_chat_id {
+                    let mut draft = state
+                        .read()
+                        .get_active_chat()
+                        .as_ref()
+                        .and_then(|d| d.draft.clone())
+                        .unwrap_or_default();
+                    let sub: String = draft.chars().take(p.try_into().unwrap()).collect();
```
sdwoodbury: remove the call to `unwrap`

sdwoodbury (dismissed):
I think it would be better if in `state/ui`, emojis were represented as a struct containing both the unicode string and a list of aliases. We may want to just use the `Emoji` struct from the emojis crate for this. 

`EmojiCounter.list` isn't a list and should just be deleted. A field should be added to contain all the `Emoji` structs, perhaps a `HashSet` (you would probably have to implement a hash function for the `Emoji` struct to do this). Another field should be added to track the number of times each emoji was used. The `emoji_filters` field should be be removed and a function should be used to compare a `&str` to each emoji's aliases. This function could be included in the `impl` for the `Emoji` struct and could be used by `EmojiCounter.get_matching_emoji`.

`register_emoji_filter` would then be removed.

Flemmli97:
> I think it would be better if in `state/ui`, emojis were represented as a struct containing both the unicode string and a list of aliases. We may want to just use the `Emoji` struct from the emojis crate for this.
> 
> `EmojiCounter.list` isn't a list and should just be deleted. A field should be added to contain all the `Emoji` structs, perhaps a `HashSet` (you would probably have to implement a hash function for the `Emoji` struct to do this). Another field should be added to track the number of times each emoji was used. The `emoji_filters` field should be be removed and a function should be used to compare a `&str` to each emoji's aliases. This function could be included in the `impl` for the `Emoji` struct and could be used by `EmojiCounter.get_matching_emoji`.
> 
> `register_emoji_filter` would then be removed.

i dont see the benefit of turning the current emoji tracker hashmap into a  hashset. instead it could have `HashMap<String,Emojistruct` but ye. i agree that the name is wrong. 
i also dont agree to have that map already containing all emojis (if thats what you mean) instead of lazily add them.
`emoji_filters` is used to allow e.g. extensions to add custom filtering option. since without the emoji extension we only have 5 default emoji we need some kind of way to obtain more emojis which is also what the filter is used for.
if you want to have a simple function to get emojis from alias you would need a way to tell the app of the additional emojis which in turn means e.g. registering all emojis to the emoji tracker which seems rather redundant when the extension itself already contains all emojis.

stavares843:
linter is failing 🔨

common/src/state/ui.rs
Flemmli97: see mostly this comment: https://github.com/Satellite-im/Uplink/pull/1193#issuecomment-1708444194

as for 3rd arg: i dont see how that is possible? as the goal of the matcher is to find a matching emoji given a prefix. it doesnt know what emoji to match to.

sdwoodbury:
@Flemmli97 you said you didn't agree to having a map that contains all the emojis instead of lazily adding them, but you did exactly that already

image

Furthermore, I doubt anyone would ever want to filter emojis (by typing ` :<name>:` ) except for name. Seems that having `ui` contain the filtering function should be fine. 

Finally, another benefit of restructuring the emoji tracker is that it would allow associating an emoji with multiple aliases, which would allow a better user experience. For example, if someone wanted the heart emoji, they could search for `:heart:` or `:love:`.

kit/src/elements/textarea/mod.rs
Flemmli97: i dont see how though? that would require `oncursor_update` EventHandler to be send async which is not really possible?

Flemmli97:
> @Flemmli97 you said you didn't agree to having a map that contains all the emojis instead of lazily adding them, but you did exactly that already
> 
> image
> 
> Furthermore, I doubt anyone would ever want to filter emojis (by typing ` :<name>:` ) except for name. Seems that having `ui` contain the filtering function should be fine.
> 
> Finally, another benefit of restructuring the emoji tracker is that it would allow associating an emoji with multiple aliases, which would allow a better user experience. For example, if someone wanted the heart emoji, they could search for `:heart:` or `:love:`.

no that doesnt add any emoji to the main app. it registers a function that gets called that returns a list of matching emojis.
instead of the extension being like: i provide all these emoji, register them it is instead uplink being: i need all emojis that match this pattern

allowing multiple alias would still be possible with this approach then. would just need to adjust the registered function to account for that

sdwoodbury:
@Flemmli97 ok, i'm convinced you're right about the emoji filter  :+1: 
looks like using the aliases would only require changing from `shortcode()` to `shortcodes()` 
I wonder if `EmojiCounter` should handle eviction too, using some combination of most used and most recently used.

Flemmli97:
> @Flemmli97 ok, i'm convinced you're right about the emoji filter 👍 looks like using the aliases would only require changing from `shortcode()` to `shortcodes()` I wonder if `EmojiCounter` should handle eviction too, using some combination of most used and most recently used.

👍 . one more thing though: if we allow multiple aliases how should we handle the emoji suggestion? as that way it will show the same emoji multiple times potentially if the aliases are very similar

sdwoodbury:
> > @Flemmli97 ok, i'm convinced you're right about the emoji filter 👍 looks like using the aliases would only require changing from `shortcode()` to `shortcodes()` I wonder if `EmojiCounter` should handle eviction too, using some combination of most used and most recently used.
> 
> 👍 . one more thing though: if we allow multiple aliases how should we handle the emoji suggestion? as that way it will show the same emoji multiple times potentially if the aliases are very similar

good question. Hopefully the aliases aren't similar. if they are, one could filter the duplicates.

kit/src/elements/textarea/mod.rs
sdwoodbury: use_coroutine returns a channel that can be used from a synchronous callback. the coroutine can execute async code.

common/src/state/ui.rs
sdwoodbury: I was thinking of using the emoji short code as an argument but it seems that isn't necessary. I see that the emoji extension provides a function that takes the search string and a bool and returns a list of matches. I think the idea of having multiple emoji filters threw me off. But having multiple filters seems ok.

stavares843:
[07-09-2023_15-19-08.log](https://github.com/Satellite-im/Uplink/files/12550129/07-09-2023_15-19-08.log)

while searching `:smile:` the app crashed

macOS

stavares843:
linter is failing 🔨

phillsatellite:
Couple things I noticed while I was testing it out (I replicated these on Linux/MacOS/Windows

- If you have 2+ chats favorited then go to seach emoji list, it will not appear 

https://github.com/Satellite-im/Uplink/assets/93608357/a9922056-d033-4a1b-88b9-9f3d9f05486b

*I think this could easily be fixed with a new ticket since its very minor*
- If you search emoji and keep list open then switch chats, the emoji list will persist 

https://github.com/Satellite-im/Uplink/assets/93608357/37b4abe4-2154-4be9-9192-03619a9f87da

luisecm:
Automated tests are failing because the emoji counter is not increaasing after adding an emoji: 

https://satellite-im.github.io/Uplink/1173/#suites/142cb8a08da9c771c94850184f339ed1/89ebe5f767bc27bb/

image

phillsatellite:
Added label for Luis's comment above 👍

Flemmli97:
wondering if i should merge this pr #1197 into this one?
as that might deal with it? and since i changed a few things in `textarea` the other pr might need changes anyway
@phillsatellite

stavares843:
we can merge the other PR first and then update this PR id say 🔨

Flemmli97:
the thing done in #1197 are now also done in this pr instead

Flemmli97:
should also fix emoji counter from extension not updating right

kit/src/elements/textarea/mod.rs
sdwoodbury: i found that dioxus doesn't allow calling callbacks from an async context. it's an unfortunate limitation.

sdwoodbury (approved):

Flemmli97:
hmm idk why clippy is complaining again. the things are os specific vars but it shouldn't error

stavares843 (approved):

PR #1204, opened 2023-09-08
feat(markdown): improve markdown handling

### LINKED ISSUE #1199: bug(chat): when editing a message that has a < appears the html
STR:
- go to chat
- send <

Captura de ecrã 2023-09-07, às 15 27 16

- edit
- appears like this

Captura de ecrã 2023-09-07, às 15 26 46

macOS

sdwoodbury:
### What this PR does 📖

- in `kit/src/components/message/mod.rs`, when creating `ChatText`, give it the text before it is converted to markdown (or before urls are wrapped with anchor links). this makes editing the message less buggy. 
- make `cargo test` stop complaining about our docs. 

- Resolve #1199


kit/src/components/message/mod.rs
```diff
@@ -451,89 +381,127 @@ pub fn ChatText(cx: Scope<ChatMessageProps>) -> Element {
...
+        },
+    }
+}
 
-    let mut options = Options::empty();
-    options.insert(Options::ENABLE_STRIKETHROUGH);
+fn multiline_code_regex(target: &str) -> Option<String> {
+    let re = Regex::new(r"(?<code_block>```(?<language>[a-z]+)(\s|\n)+(?<code>(.|\s)+)```)")
+        .expect("invalid regex");
+    if let Some(caps) = re.captures(target) {
```
dariusc93: because the function returning a `Option`, wouldnt it be better to do `let caps = re.captures(target)?;`? It would remove the `if/else` clause

kit/src/components/message/mod.rs
```diff
@@ -163,15 +162,18 @@ pub fn Message<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
         })
     });
 
-    // if markdown support is enabled, we will create it, otherwise we will just pass text.
-    let mut formatted_text = if cx.props.parse_markdown {
-        markdown(&text)
+    let loading_class = loading.then_some("loading").unwrap_or_default();
+    let remote_class = is_remote.then_some("remote").unwrap_or_default();
+    let order_class = if cx.props.order.is_some() {
```
dariusc93: Wouldnt this be redundant due to `order` using `unwrap_or`? In this case, we could do `let order_class = order.to_string();`

sdwoodbury: you're right

lgmarchi:
1. If I send text and code in same message, code not work

image

2. If I jump line in code block, code not work as well

image

3. If jump line, bold not work 

image

kit/Cargo.toml
```diff
@@ -20,7 +20,7 @@ humansize = { workspace = true }
 common = { workspace = true }
 warp = { workspace = true }
 derive_more = { workspace = true }
-pulldown-cmark = "0.9.2"
+# pulldown-cmark = "0.9.2"
```
lgmarchi: If it will not be used anymore, we can delete it.

kit/src/components/message/mod.rs
```diff
@@ -451,89 +381,124 @@ pub fn ChatText(cx: Scope<ChatMessageProps>) -> Element {
...
-            if line.starts_with('>') {
-                format!("\\{}", line)
-            } else {
-                line.to_string()
-            }
-        })
-        .collect();
+fn triple_backtick_regex(target: &str) -> Option<String> {
+    let re = Regex::new(r"(?<code_block>```(?<code>(.|\s)+)```)").expect("invalid regex");
+    if let Some(caps) = re.captures(target) {
```
lgmarchi: triple_backtick_regex and single_backtick_regex are basically same code. Just an additional parameter for has a different regex, and we can save some lines here.

ui/src/components/chat/sidebar.rs
```diff
@@ -563,7 +562,8 @@ pub fn Sidebar(cx: Scope<SidebarProps>) -> Element {
                     };
 
                     let subtext_val = match unwrapped_message.value().iter().map(|x| x.trim()).find(|x| !x.is_empty()) {
-                        Some(v) => markdown(v),
+                        // todo: does this really need markdown?
+                        Some(v) => v.to_string(),
```
lgmarchi: It does, because sometimes user send code block, and it will appear just a single text. Or bold text. 

image


image

ui/src/main.rs
```diff
@@ -348,7 +348,7 @@ fn use_app_coroutines(cx: &ScopeState) -> Option<()> {
                 }
                 let size = webview.inner_size();
                 let metadata = state.read().ui.metadata.clone();
-                log::debug!("resize {:?}", size);
+                //log::debug!("resize {:?}", size);
```
lgmarchi, line 350: If we will not use, better to delete already.

ui/src/misc_scripts.rs
```diff
@@ -4,16 +4,23 @@ pub const PRISM_SCRIPT: &str = include_str!("../extra/assets/scripts/prism.js");
 pub const PRISM_STYLE: &str = include_str!("../extra/assets/styles/prism.css");
```
lgmarchi, line 4: PRISM_SCRIPT doesn't need to be public.

lgmarchi (changes requested):
This PR is good, but some changes are necessary before approve.

ui/src/components/chat/sidebar.rs
sdwoodbury: after commit 90b298ed91848d470e419559eeed779677c61bd4 it now behaves same as before.

ui/src/main.rs
sdwoodbury: Matt was using that to debug. I think he might want it later.

sdwoodbury:
> 1. If I send text and code in same message, code not work
> 
> 
> image
> 
>     2. If I jump line in code block, code not work as well
> 
> 
> image
> 
>     3. If jump line, bold not work
> 
> 
> image

fixed now.

ui/src/main.rs
dariusc93: Might be better to uncomment it. Maybe put it at a trace log

dariusc93 (approved):

lgmarchi (approved):

stavares843 (approved):

PR #1207, opened 2023-09-11
feat(call): Speaking indicator

### LINKED ISSUE #1177: feat(calling): add visual indicator when someone is speaking
Blink emits the following events: `BlinkEventKind::SelfSpeaking` and `BlinkEventKind::ParticipantSpeaking`. When these events are emitted, we need a visual indicator, such as changing the border on the user's profile picture.

Flemmli97:
### What this PR does 📖

- Adds indicator for when a user is speaking in a call
- Username will be enlarged and highlighted
- Adds system to check if a participant stopped speaking as there is only the blinkevent for when someone actually speaks `BlinkEventKind::ParticipantSpeaking` and not when someone stops.
   - This is achieved by keeping track of when a SpeakEvent comes on and checking if the last time a participant spoke was longer than 3 sec

- Resolve #1177


stavares843:
linter is failing 🔨

common/src/state/call.rs
```diff
@@ -231,7 +241,20 @@ impl Call {
...
 
     fn participant_speaking(&mut self, id: DID) {
-        self.participants_speaking.insert(id);
+        self.participants_speaking.insert(id, Instant::now());
+    }
+
+    fn update_speaking_participants(&mut self) -> bool {
+        let delay = Duration::from_secs(3);
+        let mut removed = false;
+        self.participants_speaking.retain(|_, time| {
```
sdwoodbury: it looks like this will remove at most one participant. I think it would be better if multiple participants could be removed at once. If you want to know if `retain()` modified the list, you could check the `len()` of `participants_speaking` before and after the call to `retain()`.

common/src/state/mod.rs
```diff
@@ -323,7 +323,10 @@ impl State {
     }
 
     pub fn process_warp_event(&mut self, event: WarpEvent) {
-        log::debug!("process_warp_event: {event}");
+        // Blink events are very frequent
+        if !matches!(event, WarpEvent::Blink(_)) {
```
sdwoodbury: I think it would be better to only filter out `BlinkEventKind::ParticipantSpeaking` and `BlinkEventKind::SelfSpeaking` instead of filtering out all blink events.

ui/src/components/media/calling.rs
```diff
@@ -220,7 +223,15 @@ fn ActiveCallControl(cx: Scope<ActiveCallProps>) -> Element {
                             class: "lonely-call",
                             get_local_text("remote-controls.empty")
                         })
-                    } else {
+                    } else if cx.props.in_chat {
+                        let call_participants: Vec<_> = other_participants
```
sdwoodbury, line 227: Maye this wasn't in the mock-ups but I believe users will want to see their own icon light up when they speak.

common/src/state/call.rs
Flemmli97: it wasnt but ye i do agree that approach would be more clearer

ui/src/components/media/calling.rs
Flemmli97: hmm ye idk. we currently have it so the user does not have an indicator (or profile pic) in calls. this was also the case in the figmas irrc

common/src/state/mod.rs
```diff
@@ -323,7 +323,14 @@ impl State {
     }
 
     pub fn process_warp_event(&mut self, event: WarpEvent) {
-        log::debug!("process_warp_event: {event}");
+        // Blink events are very frequent
+        if !matches!(
+            event,
+            WarpEvent::Blink(BlinkEventKind::ParticipantSpeaking { peer_id: _ })
```
lgmarchi: Just a suggestion
```suggestion
            WarpEvent::Blink(BlinkEventKind::ParticipantSpeaking { .. })
```

common/src/state/mod.rs
```diff
@@ -323,7 +323,14 @@ impl State {
...
+                | WarpEvent::Blink(BlinkEventKind::SelfSpeaking)
+        ) {
+            log::debug!("process_warp_event: {event}");
```
lgmarchi, line 332: It is not better process_blink_event?

ui/src/main.rs
```diff
@@ -434,6 +434,20 @@ fn use_app_coroutines(cx: &ScopeState) -> Option<()> {
...
     });
 
+    //Update active call
+    use_future(cx, (), |_| {
+        to_owned![state];
+        async move {
+            loop {
+                sleep(Duration::from_secs(1)).await;
+                log::trace!("updating active call");
+                if state.write_silent().ui.call_info.update_active_call() {
```
lgmarchi, line 443: Here we need to use read(), right? 


```suggestion
                if state.read().ui.call_info.update_active_call() {
```

lgmarchi (approved):
Looks good to me. Just left small suggetions.

common/src/state/mod.rs
sdwoodbury: i'm not sure what the question is but the answer is probably no :grin:

ui/src/components/media/calling.rs
sdwoodbury: maybe adding an indicator for the local user should be handled in another ticket then.

common/src/state/mod.rs
dariusc93: I believe he referring to replacing `process_warp_event` with `process_blink_event`, however I think `process_warp_event` is appropriate since blink is still apart of warp.

ui/src/main.rs
dariusc93: `update_active_call` is a mutable function as outlined in [common/src/state/call.rs](https://github.com/Satellite-im/Uplink/pull/1207/files#diff-134fd3263262466455426cd37bd253a755ba88c9e9e1a201578de7b86c4ace62) so calling `read` would cause an error.

ui/src/components/media/calling.rs
Flemmli97: 👍

sdwoodbury (approved):

PR #1209, opened 2023-09-12
fix(lang): Lang update

Flemmli97:
### What this PR does 📖

- Makes using translation with args easier to use be providing a method where you can just pass an args vector to it instead of the current map function
- Changes all places where instead of using translation args it still uses rusts `format!` as not every language will follow the same layout


luisecm:
Hi, I see that the automated tests are failing when typing a pin with more than 32 chars:

Then, I replicated the issue manually and it failed:
https://github.com/Satellite-im/Uplink/assets/35935591/1493bc93-27e4-4c81-9385-12337fde0348

Crash Log:
[12-09-2023_15-33-04.log](https://github.com/Satellite-im/Uplink/files/12591097/12-09-2023_15-33-04.log)

stavares843:
```
Expected: "Please enter at least 1 character."
Received: "Please enter at least 1 character"
``` 


the tests are failing because is missing a dot in the end of the error message

common/src/language/mod.rs
```diff
@@ -62,18 +62,19 @@ pub fn get_local_text(text: &str) -> String {
...
+        for (key, val) in args {
+            m.insert(key, val);
+        }
+    })
 }
 
-pub fn get_local_text_args_builder<F, T: AsRef<str>>(text: &str, builder: F) -> String
+// Looks and formats a local text using the given args
+// Provides the undelying args map
+pub fn get_local_text_args_builder<'a, F, T: AsRef<str>>(text: &str, builder: F) -> String
```
lgmarchi, line 75: Can we let it not public?

common/src/language/mod.rs
```diff
@@ -62,18 +62,19 @@ pub fn get_local_text(text: &str) -> String {
...
 
 // Looks and formats a local text using the given args
-pub fn get_local_text_with_args<T: AsRef<str>>(
-    text: &str,
-    args: &HashMap<T, FluentValue>,
-) -> String {
-    LOCALES
-        .lookup_with_args(&APP_LANG.read().0, text, args)
-        .unwrap_or_default()
+pub fn get_local_text_with_args(text: &str, args: Vec<(&str, FluentValue<'_>)>) -> String {
```
lgmarchi, line 65: Not for this PR, but could we let arguments in translated file as key1, key2? To be more generic and work for every text?

And instead of a Vec, use a HashMap.

ui/src/layouts/storage/send_files_components.rs
```diff
@@ -52,9 +53,7 @@ pub fn send_files_from_chat_topbar(
                     },
                 },
                 Button {
-                    text: get_local_text_args_builder("files.send-files-text-amount", |m| {
-                        m.insert("amount", format!("{}/{}", files_selected_to_send.read().len(), MAX_FILES_PER_MESSAGE).into());
-                    }),
+                    text: get_local_text_with_args("files.send-files-text-amount", vec![("amount", format!("{}/{}", files_selected_to_send.read().len(), MAX_FILES_PER_MESSAGE).into())]),
```
lgmarchi, line 56: Here for example, would be nice a HashMap with Key1, and Key2. 

And those Keys could be a enum as well, just for us avoid Strings.

But all those suggestions for a next PR, don't worry for this one.

lgmarchi (approved):
LGTM, just left small suggestions for a next PR.

stavares843 (approved):

stavares843:
```
Expected: "Maximum of 128 characters exceeded."
Received: "Maximum of 128 characters exceeded"
```

the tests are failing because is missing a dot in the end of the error message

common/locales/en-US/main.ftl
```diff
@@ -59,9 +58,10 @@ updates = Updates
     .download-label = Download Update
 
 warning-messages = Warning Messages
-    .please-enter-at-least = Please enter at least
-    .maximum-of = Maximum of
-    .disallowed-characters = Not allowed character(s)
+    .please-enter-at-least-one = Please enter at least 1 character.
+    .please-enter-at-least = Please enter at least { $num } characters
```
stavares843:
```suggestion
    .please-enter-at-least = Please enter at least { $num } characters.
```

stavares843:
added commit for the failing test 🔨

common/locales/en-US/main.ftl
```diff
@@ -59,9 +58,10 @@ updates = Updates
     .download-label = Download Update
 
 warning-messages = Warning Messages
-    .please-enter-at-least = Please enter at least
-    .maximum-of = Maximum of
-    .disallowed-characters = Not allowed character(s)
+    .please-enter-at-least-one = Please enter at least 1 character.
+    .please-enter-at-least = Please enter at least { $num } characters.
+    .maximum-of = Maximum of { $num } characters exceeded
```
stavares843:
```suggestion
    .maximum-of = Maximum of { $num } characters exceeded.
```

PR #1214, opened 2023-09-14
chore(chats): refactor get_messages

sdwoodbury:
### What this PR does 📖

all this does is create new files for `use_effect`, `use_coroutine`, and some scripts. It makes `get_messages` a lot easier to read. I'm planning to change this file more and want to make future pull requests as small as possible. 

- Resolve #

### DIFF
--- a/ui/src/components/chat/compose/messages/effects.rs
+++ b/ui/src/components/chat/compose/messages/effects.rs
@@ -1,11 +1,10 @@
-
+use crate::utils;
 use common::state::State;
 use dioxus_core::Scoped;
 use dioxus_hooks::{to_owned, use_effect, UseRef, UseSharedState};
 use uuid::Uuid;
-use crate::utils;
 
-use crate::components::chat::compose::messages::{
+use crate::components::chat::compose::messages::scripts::{
     SCROLL_BOTTOM, SCROLL_TO, SCROLL_UNREAD, SETUP_CONTEXT_PARENT,
 };
 
--- a/ui/src/components/chat/compose/messages/effects.rs
+++ b/ui/src/components/chat/compose/messages/effects.rs
@@ -8,7 +8,33 @@ use crate::components::chat::compose::messages::scripts::{
     SCROLL_BOTTOM, SCROLL_TO, SCROLL_UNREAD, SETUP_CONTEXT_PARENT,
 };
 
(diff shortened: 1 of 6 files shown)


ui/src/components/chat/compose/messages/effects.rs
```diff
@@ -0,0 +1,90 @@
...
+    currently_active: &Option<Uuid>,
+) {
+    use_effect(cx, scroll_to, |_| {
+        to_owned![state, eval, currently_active];
+        async move {
+            let currently_active = match currently_active {
+                Some(r) => r,
+                None => return,
+            };
+            if let Some(uuid) = state.write_silent().check_message_scroll(&currently_active) {
```
lgmarchi, line 53: Here, we need to use read(), right? 
```suggestion
            if let Some(uuid) = state.read().check_message_scroll(&currently_active) {
```

dariusc93: > Here, we need to use read(), right?

If I recall, `check_message_scroll` is a mutable function so calling `read` would cause an error

ui/src/components/chat/compose/messages/mod.rs
```diff
@@ -469,7 +231,7 @@ pub fn get_messages(cx: Scope, data: Rc<super::ComposeData>) -> Element {
                                 identity_profile.set(id);
                             }
                             //Dont think there is any way of manually moving elements via dioxus
-                            let script = include_str!("../show_context.js")
+                            let script = include_str!("../../show_context.js")
```
lgmarchi: Just a suggestion, I think it is better to import like a const variable in the beginning of the file, and use this variable here, instead of let file path directly.

lgmarchi (approved):
LGTM, nice changes to organize code. Just left 2 small suggestions.

ui/src/components/chat/compose/messages/effects.rs
lgmarchi: Hum... so it is changing a value inside state, and return other value to use in conditional, right?

sdwoodbury: Darius is correct. `read()` is inappropriate here.

dariusc93 (approved):

PR #1220, opened 2023-09-15
feat(calling): Implement volume configs for users

### LINKED ISSUE #1176: feat(calling): allow changing participants' volume
If someone is quiet, you might want to make their audio track sound louder without turning up the volume on your speakers/headset. Blink provides an API for this: `set_peer_audio_gain(...)`.  Suggested input options: [0.25, 0.5, 0.75, 1, 2, 3, 4, 5]. Note that without a getter, the UI will have to track what the gain was set to.

Flemmli97:
### What this PR does 📖

- Implements the ability to change other users volume. This is done via the quickprofile menu by adding a volume slider there. Volume can be adjusted with a scale of 0-5

- Resolve #1176


ui/src/components/chat/compose/quick_profile.rs
```diff
@@ -299,6 +335,18 @@ pub fn QuickProfileContext<'a>(cx: Scope<'a, QuickProfileProps<'a>>) -> Element<
...
+                        min: 0.0,
+                        max: 5.0,
+                        step: 0.1,
+                        no_num: true,
+                        icon_left: Icon::Speaker,
+                        icon_right: Icon::SpeakerWave,
+                        onchange: move |val| {
+                            ch.send(QuickProfileCmd::AdjustVolume(did_cloned.clone(), val));
+                        }
+                    }
```
sdwoodbury: the `min` should probably be non-zero. Recommend one half or one quarter. Recommend a step size of one half or  one quarter as well.

ui/src/components/settings/sub_pages/audio.rs
sdwoodbury, line 1: i'm not sure we need settings to change the local microphone and speaker gain. It's probably better to let the host OS handle this.

kit/src/elements/range/mod.rs
```diff
@@ -5,17 +5,26 @@ use dioxus::prelude::*;
...
     icon_right: Option<Icon>,
 }
 
 #[allow(non_snake_case)]
 pub fn Range<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
     let internal_state = use_state(cx, || cx.props.initial_value);
+    use_effect(cx, &cx.props.initial_value, |val| {
+        to_owned![internal_state];
+        async move {
+            internal_state.set(val);
```
sdwoodbury, line 24: we should check `internal_state` and only update it if it differs from `val`.

sdwoodbury (dismissed):

kit/src/elements/range/mod.rs
Flemmli97: thats what it does. if initial_value is different the async part will be called

ui/src/components/settings/sub_pages/audio.rs
Flemmli97: hmm im unsure what you mean here?
you mean where during fetching it will update the selected device again?
i just though that would be better as if e.g. a device was unplugged it will otherwise not show the updated device but functionally will use the updated one which might cause confusion if it still shows the old one

ui/src/components/chat/compose/quick_profile.rs
```diff
@@ -227,6 +239,30 @@ pub fn QuickProfileContext<'a>(cx: Scope<'a, QuickProfileProps<'a>>) -> Element<
...
+                            volume,
+                            rsp: tx,
+                        })) {
+                            log::error!("failed to send blink command: {e}");
+                            continue;
+                        }
+
+                        match rx.await {
+                            Ok(_) => {
+                                state
```
lgmarchi: This code is as write_silent, will it change UI?

ui/src/components/chat/compose/quick_profile.rs
```diff
@@ -299,6 +335,18 @@ pub fn QuickProfileContext<'a>(cx: Scope<'a, QuickProfileProps<'a>>) -> Element<
                             }*/
                         )
                     }
+                    Range {
+                        initial_value: volume,
+                        min: 0.25,
```
lgmarchi: I would recommend all those numbers as const variables for Volume settings.

ui/src/components/chat/compose/quick_profile.rs
Flemmli97: no. only quickprofile which gets done before

ui/src/components/settings/sub_pages/audio.rs
```diff
@@ -149,15 +149,14 @@ pub fn AudioSettings(cx: Scope) -> Element {
             },
             SettingSectionSimple {
                 Range {
-                    initial_value: 100,
-                    min: 0,
-                    max: 200,
+                    initial_value: 100.0,
```
lgmarchi, line 155: Those numbers as variables as well.

lgmarchi (approved):

Flemmli97:
@sdwoodbury asking about the reply to your points there

kit/src/elements/range/mod.rs
sdwoodbury: I suspect that the entire `use_effect` could be deleted. You probably don't need a hook either. I think that `audio.rs` should update the `initial_value` prop in the `oninput` callback. currently it gets reset to `VOL_MIN` every time. 
image


https://github.com/Satellite-im/Uplink/assets/8636602/49d1b9f8-af66-4072-8004-8fe716efcb34

ui/src/components/settings/sub_pages/audio.rs
sdwoodbury: at the very least, blink doesn't support said feature yet.

kit/src/elements/range/mod.rs
Flemmli97: > I suspect that the entire `use_effect` could be deleted. You probably don't need a hook either. I think that `audio.rs` should update the `initial_value` prop in the `oninput` callback. currently it gets reset to `VOL_MIN` every time. image
> 
>  audio-setting-bug-2023-09-21_15.17.13.mp4

this only applies to the setting here as the audio settings here are not implemented yet.
e.g. in the quickprofile menu which also uses a range slider the initial value is the current user volume.
if we dont do it this way the slider will not get updated correctly as `oninput` gets fired when the slider changes. not when it gets loaded

sdwoodbury: ok, i see how the `value` field of the `input` element is set to `internal_state`. my mistake in asking for that hook to be removed. I still think the `use_effect` is basically a no op though.

Flemmli97: it allows to check when `initial_value` gets updated. you could maybe achieve it by having a `use_ref` or something that keeps the last value and checks if thats different but i dont see the need for that when `use_effect` achieves the same without doing lot of manual stuff

Flemmli97:
think all points were addressed. but @sdwoodbury should prob check to make sure

sdwoodbury:
> think all points were addressed. but @sdwoodbury should prob check to make sure

as long as we're ok with adding a UI for a feature that isn't implemented yet, then it's fine.

Flemmli97:
> > think all points were addressed. but @sdwoodbury should prob check to make sure
> 
> as long as we're ok with adding a UI for a feature that isn't implemented yet, then it's fine.

the ui is already in dev. the changes show up in this pr as i changed the underlying `Range` structure from int to float

phillsatellite (approved):

PR #1226, opened 2023-09-20

dariusc93:
### What this PR does 📖

- Simplify discovery, reducing logic 
- Remove nat status check and allow relay to be enabled regardless of port forwarding status
- Introduce a secondary discovery flag
- Refactored Blink interface, removing codec from `Blink::offer_call` and added `Blink::get_audio_device_config`
- Adds the capability (to Blink) to test audio I/O devices before selecting them for future use. 
- Blink can now be used with 3 channel microphones. 

- N/A

- Though not perfect, connectivity status should have improve compared to before.

### Additional comments 🎤
- Discovery has a secondary point (which is disabled by default) for discovering peers outside of DHT or direct connection, which would reduce cpu and memory consumption, however this will be refactored in the very near future
- Discovery can be disable with `--no-discovery`, which would tunnel and connect to peers directly via common relays. Note that any additional relays besides the default may not be recognized initially.


phillsatellite (approved):

ui/src/components/chat/compose/mod.rs
sdwoodbury, line 1: this file shouldn't exist anymore

sdwoodbury (changes requested):
everything else looks good. just that one file needs to be deleted.

sdwoodbury (approved):

PR #1227, opened 2023-09-20
chore(chats): refactor the chat layout to be self contained

sdwoodbury:
### What this PR does 📖

- This PR incorporates elements of CleanArch into the chat layout (renamed to chats). It should not change the behavior of the app. 
- Everything that goes into the chats layout can now be easily found in one place. other layouts which need the chats sidebar can still access it and use it the same way as before. 
- Not everything has been reorganized as much as possible. For example, later the chatbar might become its own folder, with separate files for the coroutines and use_futures. My main focus is on `messages` so some things (like the chatbar) were just moved. things could be further reorganized later. 
- note that the `style.scss` files were copied over and placed in the same hierarchy as before. 

- Resolve #


phillsatellite (approved):

dariusc93 (approved):

stavares843 (changes requested):
missing linter 🔨

sdwoodbury:
> missing linter 🔨

image

stavares843 (approved):

PR #1238, opened 2023-09-22
feat(logging): Add additional logging profiles

dariusc93:
### What this PR does 📖

- Adds `LogProfile::DebugAll`,`LogProfile::TraceAll`, 'LogProfile::TraceWarp` and `LogProfile::TraceDioxus`
- Removes `LogProfile::Trace2`

- N/A

- To use, use `debug-all` or `trace-all` instead of `trace2`. 
- This will not output anything to terminal but instead would output to `$UPLINK/.user/debug.log` (replacing `$UPLINK` with the path to the uplink instance).  

### Additional comments 🎤
- Using `debug-all` or `trace-all` will output to a file, while all other logging profiles will output to terminal.


sdwoodbury (approved):

lgmarchi (approved):

InfamousVague:
Perhaps this is out of scope for this ticket, but how difficult would it be to show this output in the built-in console in app? We could then toggle on and off different log levels in real time kinda like chrome debug console (or most major browsers)

phillsatellite (approved):

phillsatellite:
Added changes requested for Matts comment

sdwoodbury:
> Added changes requested for Matts comment

anything written to the console is also written to the logger display in the app.

dariusc93:
Just to add, the additional logs from all other crates would be to much for displaying in the built-in console (which is why its written to a file instead), so while it technically couldve fallen within the scope of this PR, I dont think I can recommend it. All other logging commands though (eg dioxus, warp, uplink) is fine and iirc should display in the console.

PR #1250, opened 2023-09-28
fix(friends): Fix did hash for outgoing/incoming being reversed

### LINKED ISSUE #1235: bug(friends): hash is displayed backwards
image

FR someone and the last X digits of their id (everything after the #) are shown backwards on the friends page.

Flemmli97:
### What this PR does 📖

- Fixes hash being displayed in reverse for outgoing/incoming friend requests.
  - Reason `Identity#short_id` is not used like in friend page is the short id is not present if the identity is not resolved fully

- Resolve #1235


ui/src/components/friends/incoming_requests/mod.rs
```diff
@@ -97,7 +97,8 @@ pub fn PendingFriends(cx: Scope) -> Element {
                 let _status_message = friend.status_message().unwrap_or_default();
                 let did = friend.did_key();
                 let did2 = did.clone();
-                let did_suffix: String = did.to_string().chars().rev().take(6).collect();
+                let mut did_suffix: String = did.to_string();
+                did_suffix = did_suffix.char_indices().nth_back(6).map(|(i,_)|did_suffix[i..].to_string()).unwrap_or_default();
```
dariusc93: Instead, you can use `Identity::short_id` in which you could do something like

```rust
let did_suffix = friend.short_id().to_string();
```

ui/src/components/friends/outgoing_requests/mod.rs
```diff
@@ -61,7 +61,8 @@ pub fn OutgoingRequests(cx: Scope) -> Element {
             friends_list.into_iter().map(|friend| {
                 let did = friend.did_key();
                 let did2 = did.clone();
-                let did_suffix: String = did.to_string().chars().rev().take(6).collect();
+                let mut did_suffix: String = did.to_string();
+                did_suffix = did_suffix.char_indices().nth_back(6).map(|(i,_)|did_suffix[i..].to_string()).unwrap_or_default();
```
dariusc93: ditto

ui/src/components/friends/incoming_requests/mod.rs
Flemmli97: already in the description but it seems the short_id is not always present during requests?

phillsatellite:
Noticed one thing while testing, the did will still appear reversed in the blocked tab, this happens on all os's. Other than that seemed good! 

Screenshot 2023-09-28 at 3 50 46 PM

lgmarchi (approved):

phillsatellite (approved):

ui/src/components/friends/incoming_requests/mod.rs
dariusc93: If thats the case, maybe we can also set it in `get_uninitialized_identity`?

sdwoodbury: i concur. it can be done like this: 

```
let pk = did.public_key_bytes();
let short: [u8; 8] = pk[pk.len() - 8..].try_into().map_err(|_e| warp::error::Error::OtherWithContext("did to short".into()))?;
default.set_short_id(short);
```

dariusc93 (approved):

PR #1260, opened 2023-09-29
feat(call): Impl calling ringtone

### LINKED ISSUE #1236: task(calling): Add ringtone for incoming calls
**Issue:**
Scope of this ticket is pretty much what the title says we want to add a ringtone for incoming calls

**OS:**
All

Flemmli97:
### What this PR does 📖

- Implements ringtone for incoming calls.
  - Adds ability to play a sound on loop which uses an `Arc<AtomicBool>` to stop the sound so 
    1. Stops the sound if there are no refs for the `Arc` left e.g. a component is not rendered anymore
    2. Allows also manually stopping the sound (in general. not this case)

Updated:
- Ringtone is from https://freesound.org/people/Knox1/sounds/572112/ in case it matters

- Resolve #1236


phillsatellite (approved):

stavares843 (approved):

lgmarchi (approved):

common/src/sounds.rs
```diff
@@ -38,3 +50,39 @@ pub fn Play(sound: Sounds) {
...
+                        sound_inst = Some(sink);
+                    }
+                }
+                Some(sound) => {
+                    if sound.empty() {
+                        sound_inst = None
+                    }
+                }
+            }
+            if Arc::strong_count(&condition) <= 1 || condition.load(Ordering::Relaxed) {
```
dariusc93:
```suggestion
            if Arc::strong_count(&condition) == 1 || condition.load(Ordering::Relaxed) {
```

ui/src/components/media/calling.rs
```diff
@@ -490,6 +494,11 @@ fn PendingCallDialog(cx: Scope<PendingCallProps>) -> Element {
             }
         };
     }
+    let alive = use_ref(cx, || Arc::new(AtomicBool::new(false)));
```
dariusc93, line 506: because of this being atomic, does this have to set behind a `use_ref`? Also, where is `AtomicBool` being set at for it to trigger in the condition?

sdwoodbury, line 506: instead of tracking when the `UseRef<Arc<AtomicBool>>` gets dropped, it may be better to to set the variable in the handlers for `with_accept_btn` and `with_deny_btn`. I'm guessing it gets dropped correctly because it passed testing but I think it would be more clear to explicitly set the variable to `false`.

sdwoodbury: it might be better to use a `UseRef` instead of an `Arc`, to ensure that if somehow the Element re-renders, it doesn't make a new `AtomicBool`.

ui/src/components/media/calling.rs
```diff
@@ -490,6 +494,11 @@ fn PendingCallDialog(cx: Scope<PendingCallProps>) -> Element {
...
+    use_effect(cx, (), |_| {
```
sdwoodbury, line 507: the `use_effect` needs to re-run for subsequent calls. the call id should be passed as a dependency to ensure this happens. 

edit: i'm assuming that testing consisted of a single call. If it worked for multiple calls then no problem.

sdwoodbury (dismissed):

InfamousVague:
Small note, we can't accept this PR with the current tone due to the licenses on the site (
image
)

I'll search for a royalty free tune we can use for now.

InfamousVague:
It seems this is royalty free (albeit very basic) https://freesound.org/people/snzl/sounds/595932/

It would be excellent if we could make the sound itself configurable with an "open sounds folder" much like we do for fonts.

ui/src/components/media/calling.rs
Flemmli97: useref gets only initialized once.
also needs to be a useref as to get "bound" to the component. else it will indeed create a new atomicbool everytime it rerenders.

ui/src/components/media/calling.rs
Flemmli97: i dont see a reason as to why though? you can only have one active call or an incoming call.

ui/src/components/media/calling.rs
Flemmli97: imo isnt it better if its done automatically instead of needing to manually stop it?

ui/src/components/media/calling.rs
sdwoodbury: I was thinking the same `use_effect` would be used across multiple instantiations of the `Element`. but that's probably not how it works. comment withdrawn.

ui/src/components/media/calling.rs
sdwoodbury: unless the dev forgets about that part months later when they need to work on it again.

stavares843:
added changes requested label due to matt comment above

ui/src/components/media/calling.rs
Flemmli97: `PlayUntil` has a comment indicating the behavior. isnt that enough?

ui/src/components/media/calling.rs
Flemmli97: `use_effect` gets run for every new instance. but not for rerenders

Flemmli97:
updated sound file. and left some comments for @sdwoodbury points

stavares843 (approved):

stavares843:
resolved the chats to be able to merge

PR #1290, opened 2023-10-05
chore(appium): close update alert from gh runner before tests

luisecm:
### What this PR does 📖

- Recently the MacOS GH runners started to show an alert for update available on the OS that is being positioned in front of the application
- The problem is that webdriverio is not ignoring this alert and its failing test execution when trying to interact with elements located behind of this alert
- Asked to chatgpt for an applescript that could close these alerts before starting tests, so trying adding a new step before running tests. Example of the alert displayed on GH runner:
image

- Resolve #


luisecm:
Closing this since it did not work. Trying another approaches on the testing repo

PR #1299, opened 2023-10-09
feat(ui): Window size setting

### LINKED ISSUE #1111: bug(ui): Remember size of app from when previously logged in
**Issue:**
When you adjust size of app and log out/back, state is not remembering the size and adjusts to default 

**OS:**
All

**STR:**
- Adjust size of app 
- Log out
- Log in 

**Screen recording:**

https://github.com/Satellite-im/Uplink/assets/93608357/1577c223-d575-4c28-a2ee-df1c3753028f

Flemmli97:
### What this PR does 📖

- Makes it so window size is remembered and reloaded on app start

- Resolve #1111


phillsatellite:
While testing I noticed a bug on MacOS, I was not able to replicate on Windows or Linux. 

When you re open the app on MacOS the width will be the same as when you closed app but the height will stretch from the top of the screen to the bottom 

https://github.com/Satellite-im/Uplink/assets/93608357/e1ab823e-fdfa-4da4-b204-2cccdcfe8a81

sdwoodbury (approved):
If I remember correctly, resizing the window emits a large number of events. Saving state every time will hit disk a lot. oh well - don't really have a better way to do it at the moment.

Flemmli97:
ye unless there is a way to detect when someone stops resizing (which i doubt) i dont think there is a better way.

phillsatellite:
Getting this error when trying to compile on Windows and Mac. Linux was able to compile fine 

Screenshot 2023-10-11 at 11 13 43 AM

stavares843:
missing linter 🔨

phillsatellite:
Noticed something else when re-tested, on MacOS and Windows if you enter fullscreen, exit app and re-enter, the app will be halfway off the screen 


https://github.com/Satellite-im/Uplink/assets/93608357/d8cf473d-ba5d-4995-993b-b3052de4cea6

Flemmli97:
should we also save the window position? as the position will be default value so depending on app size the window can still go out of the screen

luisecm:
Looks like there is any issue when compiling on MacOS and Windows:
image

phillsatellite:
> should we also save the window position? as the position will be default value so depending on app size the window can still go out of the screen

Would this be hard to do? It might be nice to add but if it might come with a lot of bugs and be a lot of hard work it might be best to hold that off for a later time

sdwoodbury:
> > should we also save the window position? as the position will be default value so depending on app size the window can still go out of the screen
> 
> Would this be hard to do? It might be nice to add but if it might come with a lot of bugs and be a lot of hard work it might be best to hold that off for a later time

@phillsatellite  @Flemmli97 
looks like this is easy enough to do once the window position is saved.  
image

Flemmli97:
done. might not work on linux though as the api handling the position does not work right there

phillsatellite (approved):

PR #1300, opened 2023-10-09
feat(chat): Quickprofile layout

### LINKED ISSUE #1255: chore(did): DID should change to a small profile picture and username inline and should display the profile card
DID should change to a small profile picture and username inline and should display the profile card below

image

Flemmli97:
### What this PR does 📖

- When typing a did in chat a profile card gets displayed instead with the option to add that person to ones friend
- Also changes profile name + status location in quickprofile menu

- Resolve #1255


common/src/warp_runner/manager/commands/multipass_commands.rs
```diff
@@ -395,6 +400,23 @@ pub async fn handle_multipass_cmd(cmd: MultiPassCmd, warp: &mut super::super::Wa
                     rsp.send(Err(e))
                 }
             };
+        }
+        MultiPassCmd::GetIdentity { did, rsp } => {
+            let pic = warp.multipass.identity_picture(&did).await;
+            let r = warp.multipass.get_identity(Identifier::DID(did)).await; //(dids, &mut warp.multipass).await;
```
dariusc93: Maybe use `map` and `and_then` chain to have something to supply to `rsp`?

common/src/warp_runner/manager/commands/multipass_commands.rs
```diff
@@ -395,6 +400,23 @@ pub async fn handle_multipass_cmd(cmd: MultiPassCmd, warp: &mut super::super::Wa
...
+            let r = match r {
+                Ok(ids) => {
+                    if ids.is_empty() {
+                        Err(Error::IdentityExist)
```
dariusc93: if the vector is empty, then the identity wouldnt exist

```suggestion
                        Err(Error::IdentityDoesntExist)
```

common/src/warp_runner/manager/commands/multipass_commands.rs
```diff
@@ -395,6 +400,23 @@ pub async fn handle_multipass_cmd(cmd: MultiPassCmd, warp: &mut super::super::Wa
                     rsp.send(Err(e))
                 }
             };
+        }
+        MultiPassCmd::GetIdentity { did, rsp } => {
+            let pic = warp.multipass.identity_picture(&did).await;
```
dariusc93: It might be best to call this if the identity exist

common/src/warp_runner/manager/commands/multipass_commands.rs
```diff
@@ -395,6 +400,23 @@ pub async fn handle_multipass_cmd(cmd: MultiPassCmd, warp: &mut super::super::Wa
...
+            let r = warp.multipass.get_identity(Identifier::DID(did)).await; //(dids, &mut warp.multipass).await;
+            let r = match r {
+                Ok(ids) => {
+                    if ids.is_empty() {
+                        Err(Error::IdentityExist)
+                    } else {
+                        let mut id = Identity::from(ids[0].clone());
+                        id.set_profile_picture(&pic.unwrap_or_default());
```
dariusc93: We could do an `if let` and set it if the profile picture exist

dariusc93: Maybe we should also set the banner too?

ui/build.rs
```diff
@@ -9,6 +10,7 @@ use std::{
 use walkdir::WalkDir;
 
 fn main() -> Result<(), Box<dyn Error>> {
+    env::set_var("RUST_BACKTRACE", "1");
```
dariusc93: any reason to set this variable while building?

Flemmli97: oops that was from testing a thing...

common/src/warp_runner/manager/commands/multipass_commands.rs
Flemmli97: dont really see a point specially since gotta check if returned ids are empty or not

dariusc93: Not really as you could do something like

```rust
let id = warp.multipass.get_identity(Identifier::DID(did))
        .await
        .map(|list| list.first().cloned())
        .and_then(|id| id.ok_or(Error::IdentityDoesntExist));
```

Especially since youre not expecting a list but a single identity.

kit/src/components/message/mod.rs
```diff
@@ -460,3 +470,147 @@ pub fn markdown(text: &str) -> String {
...
+}
+
+#[derive(Props, PartialEq)]
+pub struct IdentityMessageProps {
+    id: DID,
+}
+
+#[allow(non_snake_case)]
+pub fn IdentityMessage(cx: Scope<IdentityMessageProps>) -> Element {
+    let state = use_shared_state::<State>(cx)?;
```
sdwoodbury, line 575: you can't use `State` in `kit`. you have to make a new `component` in `ui`.

sdwoodbury (changes requested):

common/src/warp_runner/manager/commands/multipass_commands.rs
Flemmli97: hmm you cant use await in a closure so calling e.g. `warp.multipass.get_identity(Identifier::DID(did))` in a mapping operation doesn't really work? so you still need to do something like `match`

kit/src/components/message/mod.rs
Flemmli97: i am seeing quite a few state usages in `kit` though? why not?

phillsatellite (approved):

phillsatellite:
Awesome new feature @Flemmli97 🔥

kit/src/components/message/mod.rs
sdwoodbury: per design by @InfamousVague.

stavares843 (changes requested):
very cool 🎉 

btw i was trying out with the did's in the current did channel and i found some that don't work
 only display a empty message


```did:key:z6MkrCoZv9BamVr6WnMiSsoMEHKa1Et8CWeD18TeuvJXx4Gr```

```did:key:z6MkrxMxNMw7iu281szkP6DCx7E3XJ43C9zfkFHDdBazahoY``` 


image


if this did's are displaying a empty message due to not existing we could add ''user not found'' as @phillsatellite suggested

stavares843:
i also wonder if we should display the default avatar 

example instead of this


Captura de ecrã 2023-10-12, às 01 21 21


should we display the dots thing?

stavares843:
Captura de ecrã 2023-10-12, às 01 21 46


also the status i wonder if should appear, but if is supposed to appear it should be a bit bigger and more similar to the rest of the avatars

Captura de ecrã 2023-10-12, às 01 22 13



like this

stavares843:
also something is breaking the styles in this PR


dev

Captura de ecrã 2023-10-12, às 01 24 14

Captura_de_ecra_2023-10-12_as_01 25 21


this PR
Captura de ecrã 2023-10-12, às 01 16 12

Captura de ecrã 2023-10-12, às 01 23 08

stavares843:
all the feedback is in macOS, m1

InfamousVague:
Looks great! There is one additional requirement to wrap the actual DID key and convert it to a <span class="did-to-username">Username</span> but i'm wondering if this would be best handled in the new text parser @sdwoodbury

kit/src/components/message/mod.rs
Flemmli97: i just dont really see a way to separate it from kit (or not have state) since its so deeply nested in kit components

common/src/warp_runner/manager/commands/multipass_commands.rs
dariusc93: Although it is possible to await in a map, In this case you wouldnt be since `MultiPass::get_identity` returns `Vec<Identity>` and not a future, unless im misunderstanding your question.

Flemmli97:
> i also wonder if we should display the default avatar
> 
> example instead of this
> 
> Captura de ecrã 2023-10-12, às 01 21 21
> should we display the dots thing?

hmm idk why its displaying it empty. warp is the one handling the default img so if someone doesn't have a pic warp should provide the default pic. at least as far as i see.
@dariusc93 ^ ?

common/src/warp_runner/manager/commands/multipass_commands.rs
Flemmli97: oh sry i meant like `identity_picture`. (though both return a future)

dariusc93: Hmm in this case what we could do is import `TryFuture` and make the calls inside `map` or `and_then` using a async block, but in this case, we can leave it as for now

dariusc93:
> > i also wonder if we should display the default avatar
> > example instead of this
> > Captura de ecrã 2023-10-12, às 01 21 21
> > should we display the dots thing?
> 
> hmm idk why its displaying it empty. warp is the one handling the default img so if someone doesn't have a pic warp should provide the default pic. at least as far as i see. @dariusc93 ^ ?

Sounds related to https://github.com/Satellite-im/Uplink/issues/1314. Could you confirm if the image been cropped any?

common/src/warp_runner/manager/commands/multipass_commands.rs
Flemmli97: is there a benefit for `map` and `and_then` or is it just for readability/looking nice?

dariusc93: Just readability and reducing the amount of indents.

Flemmli97:
> Captura de ecrã 2023-10-12, às 01 21 46
> also the status i wonder if should appear, but if is supposed to appear it should be a bit bigger and more similar to the rest of the avatars
> 
> Captura de ecrã 2023-10-12, às 01 22 13
> like this

can you confirm/check what darius said @stavares843 ?

kit/src/components/message/mod.rs
Flemmli97: @sdwoodbury you have some approach ideas regarding this? only way i see it is like passing either passing some function or the component down the line but due to the depth might seem ugly.
or move the whole message thing to `ui` which isn't nice either

stavares843:
no in this case there was no image there

kit/src/components/message/mod.rs
sdwoodbury: @Flemmli97  sorry i forgot about this. let's make an exception for this case and go ahead with the current implementation.

sdwoodbury (approved):

Flemmli97:
> no in this case there was no image there

might be related to #1401. so should we handle it there instead? this way this pr isn't hold up by this

stavares843 (approved):

PR #1288, opened 2023-10-04
chore(cargo): Update dependency

dariusc93:
### What this PR does 📖

- Update dependency
- Refactor logic when downloading attachments
- Implemented `MessageOptions::limit` when fetching messages

### Which issue(s) this PR fixes 🔨
- N/A

### Special notes for reviewers 🗒️
- The change in the logic removes saving and downloaded attachments to constellation. This will be reimplemented in the near future
- If there is no response when downloading an attachment, it will give an error. This will usually happen within 60 seconds but will reset if it continues to resolve.

PR #1307, opened 2023-10-10
chore(messaging): fetch messages on demand

### LINKED ISSUE #1021: feat(chats): optimize chats layout
make an entire conversation not re-render when a new message arrives. Currently the "message grouping" causes this to happen. messages should be groups on the fly instead of using many nested elements to accomplish this.

### LINKED ISSUE #1208: feat(chats): implement infinite scroll
Now that Dioxus allows for communication between rust and js, it's probably possible to implement infinite scrolling properly.

sdwoodbury:
### What this PR does 📖

- Uplink used to load messages for each conversation on startup and then load more messages for a conversation as the user scrolled up, keeping every message ever rendered in memory. 
- Now, Uplink fetches only the most recent message for each conversation, for the sidebar, plus whatever is displayed for the active conversation. 
- This fixes the bug regarding pinned messages - if you hadn't scrolled up to a message but wanted to jump to it, now you can. 
- if  you have scrolled up in a conversation and that conversation receives new messages, the unread messages will now be correctly incremented. If you then switch chats and switch back, the unread message count will no longer be incorrectly cleared - it will remain until the user scrolls to the bottom of the conversation. 

- Hopefully this new design also prevents the app from bogging down when the app is in use for a long time. This used to happen when switching between chats which had many messages loaded. 

- Resolve #1021 
- Resolve #1208 

### Additional comments 🎤

~~waiting on https://github.com/Satellite-im/Uplink/pull/1288~~


phillsatellite:
Noticed when testing the "scroll to bottom" never goes away. 

https://github.com/Satellite-im/Uplink/assets/93608357/01583c30-52e9-4b1d-b944-70d40a43aa4c

Also when testing when you scroll up through the messages something seems off, midway through it gets extremely choppy when scrolling 

https://github.com/Satellite-im/Uplink/assets/93608357/1413ad2b-9bf0-477b-be5c-402e7883f883

sdwoodbury:
https://github.com/Satellite-im/Uplink/assets/8636602/b83a437b-e7e0-4bab-bb68-ae5a7a70f622

@phillsatellite I noticed that on Mac, If I double touch the touchpad and quickly drag, the messages window scrolls extra - it will go past the boundary and then "bounce" back to the boundary. I believe that this messes up message fetching because if the user scrolls like that, instead of a nice slow controlled scroll, the page will do a little jitter when it reloads, probably due to the bounce.


edit: another source of jitter has to do with message detection. It triggers when the message is X% in view (X is around 95 currently), but on reload, the message is scrolled entirely into view.

sdwoodbury:
found an bug with new conversations. will fix and then re-open the PR.

phillsatellite:
Went through and re tested everything again and its working a lottttt smoother. There seems to be like a 2 second delay on sending files in chat compared to dev but still works and files appear. Dont think that a 2 second delay is worth holding up the pr for. 

I wanted to mention though I did have Luis also try to replicate that _scroll to bottom_ bug and he was able to replicate it 

Was the same flow:
-Start a chat 
-Make sure the last most recent message is one that you sent that contains images 
-Scroll up till button appears 
-Manually scroll back down to the bottom of chat and the button will persist 

https://github.com/Satellite-im/Uplink/assets/93608357/ea76a4dd-4cbc-4811-82d5-d28d23266c94

Since not everyone was able to replicate this, if it would be better to approve this and then create a new ticket for that issue I am totally cool with that. Other than that everything else that was found has been resolved 👍

phillsatellite (approved):

common/src/lib.rs
```diff
@@ -165,11 +165,8 @@ pub static WARP_CMD_CH: Lazy<WarpCmdChannels> = Lazy::new(|| {
 // allows the UI to receive events to Warp
 // pretty sure the rx channel needs to be in a mutex in order for it to be a static mutable variable
 pub static WARP_EVENT_CH: Lazy<WarpEventChannels> = Lazy::new(|| {
-    let (tx, rx) = tokio::sync::mpsc::unbounded_channel();
-    WarpEventChannels {
-        tx,
-        rx: Arc::new(Mutex::new(rx)),
-    }
+    let (tx, _) = broadcast::channel(128);
```
dariusc93, line 168: Interesting to use broadcast here. I dont know the amount of events that are emitted at one time in uplink but keep in mind of the amount in case of an overflow. It is better to use though instead of anchoring the receiver behind a lock primitive.

dariusc93 (approved):

PR #1325, opened 2023-10-13

dariusc93:
### What this PR does 📖

- Consolidate modules into one
- Separated handling of identity and root documents into their own respected task
- Implemented timestamp to `Identity` 

- N/A

- There should be no difference in functionality compared to dev, but insure that the identity update properly, as well as making sure request (incoming and outgoing) persist after every restart. 

### Additional comments 🎤
- This PR adds two fields to `Identity`, being `created` and `modified`, which are respected timestamps to show when the identity was first created and the last time it was updated. Because of the introduction of the fields, the identity profile may not have backward compatibility. If the identity does not contain these timestamps, it will default to the present date and time until there is an update to the profile in which the `created` timstamp will remain and the `modified` timestamp will provide the timestamp of the last update.

Note:
- The `created` timestamp can be used externally but `modified` timestamp may not have much relevance externally


lgmarchi (approved):

phillsatellite:
Sent Darius a message this morning, on MacOS when I try to set a profile picture as soon as I click the "+" button I get a crash with this error in terminal 

register Accelerator { id: None, mods: SUPER, key: KeyV }
libunwind: stepWithCompactEncoding - invalid compact unwind encoding
zsh: abort      cargo run

Screenshot 2023-10-16 at 11 29 02 AM

https://github.com/Satellite-im/Uplink/assets/93608357/9b786664-2b30-44e5-a331-91ef086d0d66

dariusc93:
> Sent Darius a message this morning, on MacOS when I try to set a profile picture as soon as I click the "+" button I get a crash with this error in terminal
> 
> register Accelerator { id: None, mods: SUPER, key: KeyV } libunwind: stepWithCompactEncoding - invalid compact unwind encoding zsh: abort cargo run
> Screenshot 2023-10-16 at 11 29 02 AM
> Screen.Recording.2023-10-16.at.11.33.24.AM.mov

Looks related to rfd and the emoji selector. Was the extension built with the PR or used from a prior build?

dariusc93:
The issue mentioned above seem to stem from macos 14.x, however doing `cargo clean` seems to fix the issue (which for me happened on dev too).

phillsatellite:
Wanted to comment that Luis also had tried replicating with doing cargo clean and was able to replicate this bug (and found that it was most buttons that open native OS windows would force a crash - except for _open extension_) he tried on dev and seemed to be all good on there.

dariusc93:
At this point, it is determine that the issue is related to macos 14, which happens on both dev and this PR (at least I can reproduce it on both branches while in release, things do work fine), which makes the issue itself outside the scope of the PR.

stavares843 (approved):

PR #1326, opened 2023-10-13
fix(log): Dont load debug file

dariusc93:
### What this PR does 📖
- Dont load the debug.log file into the debug logger
- Create parent directory if it doesnt exist when using `debug-all` or `trace-all` 

- N/A

### Additional comments 🎤
- Loading the debug.log file into the debuglogger component can cause high memory usage if the log file is big, or cause the UI to slow down. Even if we were to read only several lines from the file instead of the whole file, that file may contain irrelevant logging info not pertaining to uplink, in which case it would be best not to read the file.
- Something that was caught was that the logger will pass along along any log level selected into uplink dev console when using any logging commands. This is a bad behaviour as the ui logger should only be logging anything from uplink, dioxus and maybe warp. Any others would be to much, however this change may come in a separate PR to filter the logs into its appropriate location.  

## Summary by CodeRabbit

Refactor:
- Removed a redundant warning message in the `update_identity` function, simplifying the code and reducing unnecessary log entries.
- Increased the channel capacity in the `log` method from 10 to 100, enhancing the logging system's performance.

Bug Fixes:
- Fixed a potential issue in the `log` method where the log entries could exceed the `max_logs` limit.

Style:
- Updated the `Logger` struct to use `PathBuf` instead of `String` for the `log_file` field, aligning with Rust's best practices.

Chores:
- Enhanced the `log` method to create the parent directory of `log_file` if it doesn't exist, ensuring the logging system's robustness.
- Simplified the `load_debug_log` function by directly using log entries from the `LOGGER` struct, improving code readability and maintainability.


sdwoodbury (approved):

ui/src/logger/mod.rs
```diff
@@ -276,26 +278,14 @@ pub fn set_write_to_stdout(b: bool) {
...
-    let raw_file = match std::fs::read_to_string(&STATIC_ARGS.logger_path) {
-        Ok(l) => l,
-        Err(e) => {
-            log::error!("failed to read debug.log: {}", e);
-            return vec![];
-        }
-    };
-
-    let mut in_memory: Vec<_> = LOGGER
+    //Note: We shouldnt read from the file since it may be too big or contain irrevelent information related to uplink
```
stavares843:
```suggestion
    //Note: We shouldnt read from the file since it may be too big or contain irrelevant information related to uplink
```

stavares843 (approved):

phillsatellite (approved):

Issue #1319, opened 2023-10-12
bug(favorites): heart button on top of chats screen for Favorites is not modified to Add/Remove quickly


luisecm:
Description: The tooltip for Favorites button on top from Chats Screen will always show "Add to Favorites". It looks like the heart button is taking too much time to change to Add/Remove after the latest commit to Dev.

Environment: MacOS and Windows

Prerequisites:
Having an account with friends

Steps to Reproduce:
1. Login with account with friends
2. Go to any chat that is not Favorite
3. Hover on the heart button at the top. It will show a tooltip with text "Add to Favorites"
4. Click on the Heart button and user will be added as Favorite on Slimbar
5. Hover again on the heart button at the top. You will see that the button was not updated and tooltip is still the same

Expected Result: "Remove from favorites" tooltip should be displayed and Button should change to Remove from favorites
Actual Result: "Add to Favorites" tooltip is still displayed even if user is already favorited and button is still the same (Add To Favorites)

Video:
https://github.com/Satellite-im/Uplink/assets/35935591/c31dea48-a3e4-4dbc-8ff6-5f5b66eab14d

sdwoodbury:
so strange 
![image](https://github.com/Satellite-im/Uplink/assets/8636602/8e09ae03-6f65-4667-ba43-055d53bc0406)

![image](https://github.com/Satellite-im/Uplink/assets/8636602/0ad91ce8-3b69-4eb8-a9a4-6b22ab65f0c5)

A favorite was loaded from the State.json file but it still isn't showing up in the sidebar.

luisecm:
I noticed that if you:
1. Add to favorites
2. Then, just switch to another chat
3. Return quickly to the conversation that was just added to favorites, then you will see the heart button/tooltip updated

(same happens with pinning messages, opening reply modal and updating name of group chat) 
Like if the elements from top are not refreshing quickly, and only when loading again the conversation all is loaded fresh

PR #1327, opened 2023-10-13
fix(chat): Fix message reply not working

### LINKED ISSUE #1321: bug(chats): reply modal is not being displayed quickly
Description: Right click on message, select reply and reply modal is no longer displayed as before. Issue started happening yesterday

Environment: MacOS and Windows

Prerequisites:
Having an account with friends

Steps to Reproduce:
1. Login with account with friends
2. Go to any chat
3. On a sent or received message try to reply
4. Modal is shown but it takes a lot of time to load

Video:

https://github.com/Satellite-im/Uplink/assets/35935591/c698811d-646e-4c00-942d-b9d2c84ba306

Flemmli97:
### What this PR does 📖

- This PR fixes message replying not working anymore
- The way messages (or chat in general) was obtained changed with #1307. For now i made it so replying goes back to the state field that is set (the way it was previously used). Im not sure if this is a good way though. Another way i see it is similar to how the new `ChatData` listens to warp event to also make a new channel that listens to state changes. 
    We would need to create a new enum for it possibility duplicating some things due to `Action` having some enum values containing non thread safe values making the usage of it directly in as a new channel command input not possible. Do we want to do it that way? 

Edit:
Adding on to it this also affects favorite chats not updating correctly (#1319) and possibly other things so we might need an additional channel after all.

- Resolve #1321

### DIFF
--- a/common/src/state/chats.rs
+++ b/common/src/state/chats.rs
@@ -11,7 +11,10 @@ use warp::{
     raygun::{self, ConversationType, Location},
 };
 
-use crate::{warp_runner::ui_adapter, STATIC_ARGS};
+use crate::{
+    warp_runner::ui_adapter::{self},
+    STATIC_ARGS,
+};
 
 use super::pending_message::{progress_file, PendingMessage};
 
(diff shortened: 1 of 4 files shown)


common/src/state/chats.rs
```diff
@@ -217,12 +220,9 @@ impl Chats {
     }
 
     /// returns the UUID of the message being replied to by the active chat
-    pub fn get_replying_to(&self) -> Option<Uuid> {
-        self.active.and_then(|id| {
-            self.all
-                .get(&id)
-                .and_then(|chat| chat.replying_to.as_ref().map(|msg| msg.id()))
-        })
+    pub fn get_replying_to(&self) -> Option<&raygun::Message> {
```
sdwoodbury, line 223: i'd prefer cloning the `raygun::Message` so that it isn't necessary to keep `state.read()` in scope.  otherwise the PR looks good.

sdwoodbury: actually this should go in ChatData, not State. my bad

sdwoodbury:
gonna try an alternative approach in another branch and try to squash all the bugs in one go

sdwoodbury:
abandoning in favor of #1329. Thanks for finding these bugs, it was a big help.

PR #1336, opened 2023-10-16
feat(markdown): use new markdown crate

### LINKED ISSUE #1305: bug(markdown): markdown never worked
edit: Uplink uses something called Clear CSS and we never implemented the styling for the tags we used to render markdown. 

image

 A few notes:
- There are a few libraries (1 rust crate and a few javascript libraries) which can pretty reliably convert text to markdown. However, Uplink currently treats everything as markdown, which results in incorrect rendering of non-markdown text. We have a hack in place to mitigate this, but unfortunately most of the markdown stuff is still broken. 
- we're also using prismjs to allow rendering code blocks with a given language. The markdown processing libraries don't support this and extra hacks are required to make both work at once. 
- the most recent change to this part of the code involved ensuring that the processed code wasn't stored in State, which allows properly editing the message. 

Considering that it's difficult or impossible to make our current implementation work, I recommend choosing one of these approaches:
- the easy way: use regexes for each thing we want to capture (bold, italics, underline, code block, strikethrough)
- the hard but maybe more efficient way: build a state machine that can do this all in a single pass. 

Extra stuff that could be cool
- detecting lines that start with one or more of these: # and rendering a heading
- detecting bulleted and numbered lists

sdwoodbury:
### What this PR does 📖

- uses the `markdowns` crate, produced by Satellite.im. 
- Transforms ascii emojis to unicode ex:  :O  -> 😮 
- detects messages that consist of only emojis and whitespace and enlarges them
- makes markdown support configurable in settings
- doesn't support lists yet
- adds css for markdown relates classes, fixing the markdown bugs present in dev. (the markdown crate isn't technically needed to do this but the two have been tested together). 

- Resolve #1305


sdwoodbury:
edit: fixed

found a bug in message replies.
image

phillsatellite:
Hey Stu I found a couple things while testing with the most latest commit (Verified that the commit Im on has the reply fix that you just recently added in) These are able to be replicated on all OS's 

- It doesnt seem like all the markdowns show up in replies, Bold appeared fine for me but all the other markdowns wouldnt appear

https://github.com/Satellite-im/Uplink/assets/93608357/835df132-ece9-4b5d-81f2-f9df95938bda

- Not all markdowns are appearing in the sidebar, bold appears fine but none of the others will appear 

Screenshot 2023-10-17 at 12 45 23 PM

- Markdowns arent appearing in Pinned messages 

Screenshot 2023-10-17 at 12 44 43 PM

- I also couldn't get underline markdowns to work

- I used the code markdown and put "cargo run --release" in there and you can see in screenshot it just says "run --release" 

Screenshot 2023-10-17 at 1 19 37 PM

sdwoodbury:
@phillsatellite  underline isn't supported.

stavares843:
Captura de ecrã 2023-10-18, às 01 21 40


is there specific ascii that are allowed? some didnt converted into emoji, example the : P one

luisecm:
Looks like the existing markdown to send code in a message is not working with this PR and thats why the chats tests are failing

sdwoodbury:
> Captura de ecrã 2023-10-18, às 01 21 40
> 
> is there specific ascii that are allowed? some didnt converted into emoji, example the : P one

yes. i think i'll move that code into uplink instead of the crate. that way it will be more clear to us. as for the above list, it's too much. don't care about most of those. `:P` will be added.

sdwoodbury:
I found that code blocks don't display html tags - probably because messages are displayed using `dangerous_html`. needs some more thought.

PR #1339, opened 2023-10-17
fix(chatbar): use typing indicator from State

sdwoodbury:
### What this PR does 📖

- fixes a bug introduced in the chats refactor. 
- the typing indicator should once again appear on the chatbar. 

- Resolve #

### DIFF
--- a/ui/src/layouts/chats/data/chat_data/active_chat/mod.rs
+++ b/ui/src/layouts/chats/data/chat_data/active_chat/mod.rs
@@ -1,7 +1,4 @@
-use std::{
-    collections::{HashMap, VecDeque},
-    time::Instant,
-};
+use std::collections::VecDeque;
 
 use common::{
     state::{self, Identity, State},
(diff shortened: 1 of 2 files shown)


phillsatellite:
Hey Stu, so we had some weird behavior. On Windows it worked great. On MacOS for Sara it also worked great (she is still on ventura) but for MacOS and my Linux VM only the typing indicator on the sidebar would appear. Im on 14 Sonoma so Im thinking it could possibly be an issue with that maybe? 

Screenshot_2023-10-17_at_1 46 32_PM

sdwoodbury:
@phillsatellite the chatbar is using the exact same typing indicator (from state) as the sidebar. did you check out the correct branch? the picture shows the behavior before the fix and I tested on Linux and mac 13.x.

phillsatellite:
> @phillsatellite the chatbar is using the exact same typing indicator (from state) as the sidebar. did you check out the correct branch? the picture shows the behavior before the fix and I tested on Linux and mac 13.x.

Yes I was on the correct branch. I was on a very old beta version of Sonoma. I restored to the current up to date version and is all working as should now. Assuming that was the cause of the error.

phillsatellite (approved):

stavares843 (approved):

PR #1354, opened 2023-10-19
feat(settings): make the settings page remember where you were

sdwoodbury:
### What this PR does 📖

- go to the settings page, should init to `Profile`.
- click on a different settings page.
- leave the settings page
- return to the settings page - should remember where you left off. 

- Resolve #

### DIFF
--- a/ui/src/components/settings/sidebar.rs
+++ b/ui/src/components/settings/sidebar.rs
@@ -12,6 +12,7 @@ use kit::{
     layout::sidebar::Sidebar as ReusableSidebar,
 };
 
+#[derive(Clone, Eq, PartialEq)]
 pub enum Page {
     About,
     Audio,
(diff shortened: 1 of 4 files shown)


luisecm:
~~I noticed that a minor change would be required on Windows Chats - Quick Profile tests (test is looking for Settings Profile, but last Settings screen for the test user is Setting Notifications). To fix this, I would need to add one line in test to go from Settings Notifications to Settings Profile. Short comment is that chats tests will fail here~~

Update: Please look at Phil's comment below:

phillsatellite:
Along with what Luis had mentioned (might be a little different than what he is describing) If Im in Settings and go to say the About Page. Then I go to Chat/QuickProfile/EditProfile instead of being taken to the Edit Profile, I land on the About Page. Other than that I think this pr is awesome 👍

phillsatellite (approved):

ui/src/components/settings/sidebar.rs
```diff
@@ -12,6 +12,7 @@ use kit::{
     layout::sidebar::Sidebar as ReusableSidebar,
 };
 
+#[derive(Clone, Eq, PartialEq)]
```
dariusc93:
```suggestion
#[derive(Copy, Clone, Eq, PartialEq)]
```

ui/src/components/settings/sidebar.rs
```diff
@@ -26,6 +27,22 @@ pub enum Page {
     Licenses,
 }
 
+impl Page {
+    pub fn set(&mut self, p: Page) {
+        *self = p;
+    }
+    pub fn get(&self) -> Self {
+        self.clone()
```
dariusc93:
```suggestion
        *self
```

dariusc93 (approved):
Some minor suggestions but not blocking to the PR

stavares843 (approved):

stavares843:
cc @sdwoodbury regarding darius feedback above

PR #1356, opened 2023-10-19
fix(FileDIalog): Not panic anymore when open File Dialog in debug mode

### LINKED ISSUE #1351: bug(files): macos 14.x and file dialogs on debug mode
STR:
- use sonoma
- run app in debug mode
- go to app
- use file dialog such as sending a file
- app crash's

lgmarchi:

- Resolve #1351

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,9 @@ members = [
 
 resolver = "2"
 
+[profile.dev]
+panic = "abort"
+
 [profile.rapid]
 inherits = "dev"
 opt-level = 2


phillsatellite (approved):

dariusc93 (approved):
Personally, I would avoid abort as it would make it hard to debug in many situations as well as aborting on any panic regardless if the panic is isolated in a separate thread or not (which is why i didnt add it in and actually this profile was removed from release mode back in june). Ideally, it would be best to figure out the actual cause the crash than to avoid it, but I do not object to this.

PR #1357, opened 2023-10-20
feat(markdown): big emoji, ascii emoji, and new settings

sdwoodbury:
### What this PR does 📖
- copies over the stuff from the previous PR. No need to keep that stuff out of dev. 
- Transforms ascii emojis to unicode ex: :O -> 😮
- detects messages that consist of only emojis and whitespace and enlarges them
- makes markdown support configurable in settings

- Resolve #

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2512,13 +2512,22 @@ dependencies = [
  "common",
  "dioxus",
  "dioxus-desktop",
- "emojis",
+ "emojis 0.6.1",
  "extensions",
  "futures",
  "kit",
  "once_cell",
  "uuid",
  "warp",
+]
+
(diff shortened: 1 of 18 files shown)


phillsatellite:
Luis and I noticed while testing that in the reply modal the big emoji's are not appearing correctly 

image

luisecm:
There is one test failing on Reply Modal validation since somehow the "message to reply" text has an extra space:

Message Sent: "Testing...😀"
Reply Modal - Message To Reply Text: "Testing...😀 "

Since the test is looking for the exact match, it is failing

<p> html tag from element:

image

phillsatellite (approved):

PR #1358, opened 2023-10-20
fix(crash): fix crash on macos

sdwoodbury:
### What this PR does 📖

- forked dioxus, branched off the rev that was currently in use, and merged Evan's branch which updated wry
- changed the way the menu is constructed on macos. 
- linux needs a new dep - added that to the readme and github workflows. 
- verified that the mac .dmg doesn't crash on sonoma. Linux seems ok. untested on windows. 

- Resolve #

### DIFF
--- a/ui/src/webview_config.rs
+++ b/ui/src/webview_config.rs
@@ -4,10 +4,10 @@ use dioxus_desktop::Config;
 use warp::logging::tracing::log;
 
 pub(crate) fn webview_config() -> Config {
-    //let window = crate::window_builder::get_window_builder(true);
+    let window = crate::window_builder::get_window_builder(true);
 
     let config = Config::new()
-        //.with_window(window)
+        .with_window(window)
         .with_custom_index(
             r#"
 <!doctype html>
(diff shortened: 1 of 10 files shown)


ui/src/webview_config.rs
```diff
@@ -4,10 +4,13 @@ use dioxus_desktop::Config;
 use warp::logging::tracing::log;
 
 pub(crate) fn webview_config() -> Config {
-    let window = crate::window_builder::get_window_builder(true, true);
-
-    let config = Config::new()
-        .with_window(window)
+    let window = crate::window_builder::get_window_builder(true);
+    let config = if cfg!(not(target_os = "macos")) {
+        Config::new().with_window(window)
```
dariusc93: Does `Config` use `self` instead of `&self`? If so, maybe we could do something like

```rust
let mut config = Config::new();
if cfg!(not(target_os = "macos")) {
   config = config.with_window(window);
}
```

dariusc93 (approved):

PR #1365, opened 2023-10-20
chore(ci): adds a workflow to build linux in every PR

stavares843:
### What this PR does 📖

- Adds a workflow to build linux in every PR to make sure it builds properly in each PR in Linux

image

### DIFF
--- a/.github/workflows/build-linux-pr.yml
+++ b/.github/workflows/build-linux-pr.yml
@@ -1,0 +1,28 @@
+
+name: Build Linux in PR
+
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+      - reopened
+
+jobs:
+  check:
+    name: Build Linux
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


WanderingHogan (changes requested):
Is it possible to use something like this: https://github.com/marketplace/actions/toml-reader to pull the toolchain value from cargo.toml?

sdwoodbury:
the cargo workflow runs/builds on linux already. Clippy will fail if the project won't compile. i believe this workflow is redundant.

luisecm:
> the cargo workflow runs/builds on linux already. Clippy will fail if the project won't compile. i believe this workflow is redundant.

This means that the job that builds linux on automated tests is redundant too. I added this job recently just to ensure that app builds correctly on Linux. Since there are no automated tests for linux yet, I will add a PR to remove my build linux job too

stavares843:
got it, thanks! 🔨

PR #1366, opened 2023-10-23
fix(chat): Markdown strikethrough fix

### LINKED ISSUE #1359: bug(chat): Strikethrough stopped working
**Issue:**
Stu just recently added Markdowns into dev (thanks Stu!) only issue that I noticed this morning was that "~ ~ strikethrough ~ ~" seems to have stopped working 

**OS:**
All

**Screenshot:**
Screenshot 2023-10-20 at 2 07 45 PM
Screenshot 2023-10-20 at 2 07 36 PM

Flemmli97:
### What this PR does 📖

- Fixes strikethrough not working
- Also fixes an issue where strikethrough are only detected if they stand alone:
  - `test ~~test~~` this works
  - `test~~test~~` this didn't works (it treated it as a whole word)
- Adds the dependency `once_cell` to kit because of regex usage. and fixes an instance where the regex was constructed always instead of cached

- Resolve #1359

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4572,6 +4572,7 @@ dependencies = [
  "humansize",
  "linkify",
  "mime",
+ "once_cell",
  "pulldown-cmark",
  "regex",
  "reqwest",
--- a/kit/Cargo.toml
+++ b/kit/Cargo.toml
@@ -25,6 +25,7 @@ linkify = { workspace = true }
 reqwest = { workspace = true }
 base64 = { workspace = true }
 mime = { workspace = true }
+once_cell = { workspace = true }
 
 [dependencies.uuid]
 workspace = true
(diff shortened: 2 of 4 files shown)


sdwoodbury (approved):
looks good 
image

phillsatellite:
Strikethrough isnt appearing correctly in reply modals but other than that it appears fine everywhere else in the app. If we could fix that real quick looks good to merge

Screenshot 2023-10-23 at 1 43 46 PM

sdwoodbury:
> Strikethrough isnt appearing correctly in reply modals but other than that it appears fine everywhere else in the app. If we could fix that real quick looks good to merge
> Screenshot 2023-10-23 at 1 43 46 PM

I think the `big-emoji` pr might fix this issue. https://github.com/Satellite-im/Uplink/pull/1357

dariusc93 (approved):

phillsatellite (approved):

PR #1369, opened 2023-10-23
chore(cargo): Update Dependency

dariusc93:
### What this PR does 📖

- Adds `IdentityImage`
- Add timestamp to `Conversation`
- Rename `Message::value` to `Message::lines`

- N/A

- This PR will not be compatible with any profiles prior.

### Additional comments 🎤
- Internally, conversations and messages are stored into a map instead of having an id for each conversation in individual files, which will make it easier for importing and exporting in the future (which may migrate into a single document in the future). 
- In the current implementation, images are encoded and then passed into `MultiPass` only to be decoded in uplink. In this PR, `IdentityImage` removes that need of needing to encode then pass into warp by passing the images directly and when getting the image again, a `IdentityImage` would be returned, which would contain the data of the image as well as the type of the image. This would remove the complexity of needing to encode/decode and having to guess internally.


phillsatellite (approved):

phillsatellite:
re-testing pr to see why ci is failing

sdwoodbury (approved):

PR #1421, opened 2023-10-26
fix(settings): fix audio 

### LINKED ISSUE #1405: bug(settings): output device is set to zero automatically
image

to reproduce, drag the slider to a non-zero value, navigate away from the settings page, then return.

sdwoodbury:
### What this PR does 📖

image

- the sliders are now ~~disabled~~ commented out  because that stuff isn't implemented. 
- the options to change the audio sampling rate and level of echo cancellation are removed because they were not implemented. 
- there is now a switch to toggle echo cancellation (echo cancellation is binary, there aren't any levels for it). 
- calls blink less regarding the audio config - blink provides a struct that can poll the hardware without further interaction with blink. This is now used in the coroutine in the audio settings page.  

- Resolve #1405


luisecm:
The automated tests will fail on this pr since there are tests asserting for texts displayed on Settings Audio - Sample Rate and Noise Suppression. So, there is no need to wait on the CI for approving this PR. I will update the tests as soon as this one is merged.

stavares843 (commented):
in my opinion the sliders being there but deactivated seems a bug

i believe we should add a coming soon when hovering instead

sdwoodbury:
> in my opinion the sliders being there but deactivated seems a bug
> 
> i believe we should add a coming soon when hovering instead

or we could just comment them out

stavares843:
or that lol

common/src/state/configuration.rs
```diff
@@ -72,7 +72,7 @@ pub struct AudioVideo {
 impl Default for AudioVideo {
     fn default() -> Self {
         Self {
-            noise_suppression: false,
+            echo_cancellation: false,
```
lgmarchi: Make sense serde default be true, and impl Default be false?

stavares843 (approved):

ui/src/components/settings/sub_pages/audio.rs
```diff
@@ -150,17 +172,18 @@ pub fn AudioSettings(cx: Scope) -> Element {
...
-                    aria_label: "range-input-device".into(),
-                    initial_value: 100.0,
-                    min: VOL_MIN,
-                    max: VOL_MAX,
-                    icon_left: Shape::Microphone,
-                    icon_right: Shape::MicrophoneWave,
-                    onchange: move |_| {}
-                }
-            }
+            // SettingSectionSimple {
```
lgmarchi, line 175: Can we add TODO for why we let it commented?

ui/src/components/settings/sub_pages/audio.rs
```diff
@@ -169,57 +192,64 @@ pub fn AudioSettings(cx: Scope) -> Element {
...
-                    aria_label: "range-output-device".into(),
-                    initial_value: VOL_MIN,
-                    min: 0.0,
-                    max: VOL_MAX,
-                    icon_left: Shape::Speaker,
-                    icon_right: Shape::SpeakerWave,
-                    onchange: move |_| {}
-                }
-            }
+            // SettingSectionSimple {
```
lgmarchi, line 199: Same here about TODO.

ui/src/components/settings/sub_pages/audio.rs
```diff
@@ -1,28 +1,26 @@
...
 use warp::logging::tracing::log;
 
-use crate::components::settings::{SettingSection, SettingSectionSimple};
+use crate::components::settings::SettingSection;
 use common::state::{action::ConfigAction, Action, State};
 use common::{sounds, WARP_CMD_CH};
 
-pub const VOL_MIN: f32 = 0.0;
-pub const VOL_MAX: f32 = 200.0;
+// pub const VOL_MIN: f32 = 0.0;
```
lgmarchi, line 15: Add a TODO why we let it commented

lgmarchi (approved):
Just left a question and suggetions

dariusc93 (approved):

common/src/state/configuration.rs
sdwoodbury: "does this make sense?" no. i think that if  a struct implements Default, it shouldn't have `#[serde(default)]` on any of its attributes.

PR #1472, opened 2023-11-07
feat(friends): Add feature to share friends

### LINKED ISSUE #1424: feat(friends): add option to share someone's contact information
i imagine this working as follows - right click on a friend from the friend's list (maybe also the quickprofile menu) and select an option to share the contact. A list appears from which you select a friend (maybe multiple) and then a special message gets sent to the person/persons, like the one shown here 
https://github.com/Satellite-im/Uplink/issues/1255

for reference, the friends page
image

and quick profile 
image

Flemmli97:
### What this PR does 📖

- Adds a feature to share the did of a friend to a chat. This is possible either via the friends list page or via quickprofile
  If via quickprofile the current chat the user is in is excluded

- Resolve #1424


phillsatellite:
This is working great flem really awesome work, I think it would be cool if we could select multiple people to send did to, similar to what we have in the share files modal but besides that works great! 
Captura_de_ecra_2023-11-07_as_19 50 13
Captura_de_ecra_2023-11-07_as_19 50 04

ui/src/components/friends/friends_list/mod.rs
```diff
@@ -336,3 +363,138 @@ pub fn FriendsSkeletal(cx: Scope) -> Element {
...
+                    height: "80px",
+                    padding: "16px",
+                    display: "inline-flex",
+                    User {
+                        username: participants_name,
+                        subtext: subtext_val,
+                        timestamp: raygun::Message::default().date(),
+                        active: false,
+                        user_image: cx.render(rsx!(
+                            if chat.conversation_type == ConversationType::Direct {rsx! (
```
dariusc93: It would be more idiomatic to use a match condition instead of let. That way if there are any other conversation types, we can expand on it cleanly.

ui/src/layouts/chats/presentation/quick_profile/mod.rs
```diff
@@ -413,6 +429,17 @@ pub fn QuickProfileContext<'a>(cx: Scope<'a, QuickProfileProps<'a>>) -> Element<
             }
         ))
         ,
+        share_did.is_some().then(||{
```
dariusc93: No need to change, but you could possibly use `map` and add the components there so if `share_did` is `Some(_)`, it can be used, otherwise it will be ignored if its `None`

phillsatellite (approved):

sdwoodbury (approved):

PR #1488, opened 2023-11-08
refactor(EmojiSuggestion): Close emoji suggestion modal with X button or typing ESCAPE keycode

### LINKED ISSUE #1479: improvement(emojis): add a close modal button or ability to press Escape to close it
@luisecm suggestion:

when the emoji suggestions are displayed... I think there should be like a close modal button or ability to press Escape to close it
I see that now the only way to close suggestions is to delete your text on input bar

in here:

image

or clicking outside too

lgmarchi:
### 1. Close emoji suggestion modal with X button or typing ESCAPE keycode

https://github.com/Satellite-im/Uplink/assets/63157656/e6b4e93d-45f8-414a-a90a-fbb7bc6d3859

- Resolve #1479


phillsatellite (approved):

sdwoodbury (commented):
Question - if ChatBar's `emoji_suggestions` was a `UseState<Vec(String, String)>>` instead of a `&'a Vec<(String, String)>`, would it work to just clear `emoji_suggestions` instead of adding the new variable ` is_emoji_suggestion_modal_closed`? 

also, if I type `:que` and `question` is suggested, I then press escape, and then enter `s`, uplink again suggests `question`. would we want suggestions to be turned off for the remainder of the word if someone exits the modal?

lgmarchi:
> Question - if ChatBar's `emoji_suggestions` was a `UseState<Vec(String, String)>>` instead of a `&'a Vec<(String, String)>`, would it work to just clear `emoji_suggestions` instead of adding the new variable ` is_emoji_suggestion_modal_closed`?
> 
> also, if I type `:que` and `question` is suggested, I then press escape, and then enter `s`, uplink again suggests `question`. would we want suggestions to be turned off for the remainder of the word if someone exits the modal?

Probably it will work and close modal as well if we clear this variable, but it was not done, because the answer of your second question, to let exactly behavior you described.

sdwoodbury (approved):

PR #1499, opened 2023-11-10
task(groups): Change calling to only include joined participants

### LINKED ISSUE #1473: task(groups): Only show users who are in the call
**Task:**
In group calling, when the call starts every member of the group will show up in the calling UI, we should only show people who have answered the call. 

Screenshot 2023-11-07 at 1 27 43 PM

Flemmli97:
### What this PR does 📖

- Makes it so calls only show joined participants

- Resolve #1473### Additional comments 🎤

### DIFF
--- a/ui/src/components/media/calling.rs
+++ b/ui/src/components/media/calling.rs
@@ -319,7 +319,7 @@ fn ActiveCallControl(cx: Scope<ActiveCallProps>) -> Element {
     }
     let call = &active_call.call;
 
-    let participants = state.read().get_identities(&call.participants);
+    let participants = state.read().get_identities(&call.participants_joined);
     let other_participants = state.read().remove_self(&participants);
     let participants_name = State::join_usernames(&other_participants);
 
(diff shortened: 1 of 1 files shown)


phillsatellite:
@Flemmli97 Works well! Just had one suggestion but I guess it depends on what others think as well maybe @XileHorizon can give his opinion. Should we show users own profile within the call ui? Like when I place call and no one else has joined yet instead of showing "Nobody is here" the user would see just their profile avatar. We could also just add at a later more appropriate time in a different ticket if that works! 

https://github.com/Satellite-im/Uplink/assets/93608357/5b757dd8-a4a5-42a5-9f4b-c90ae2564ba0

Flemmli97:
irrc in the past we wanted to exclude the current user from it. maybe im wrong though?

phillsatellite:
> irrc in the past we wanted to exclude the current user from it. maybe im wrong though?

we can keep it like that  👍

sdwoodbury (approved):

XileHorizon:
> @Flemmli97 Works well! Just had one suggestion but I guess it depends on what others think as well maybe @XileHorizon can give his opinion. Should we show users own profile within the call ui? Like when I place call and no one else has joined yet instead of showing "Nobody is here" the user would see just their profile avatar. We could also just add at a later more appropriate time in a different ticket if that works!
> 
>  Screen.Recording.2023-11-10.at.2.14.12.PM.mov

In my designs, I have the user icon in the bottom left corner. 
image

phillsatellite:
@Flemmli97 if you want me to make a ticket for making call ui resemble figma designs and approve this Im cool with that

stavares843 (approved):

phillsatellite (approved):

PR #1509, opened 2023-11-13
fix(blink): fix signaling for group calls

### LINKED ISSUE #1409: bug(calling): sometimes users can't hear each other
during a group call, 2 participants could hear everyone and be heard by everyone, the other two could hear the other's voices but couldn't be heard.

sdwoodbury:
### What this PR does 📖

- changes the algorithm used to connect peers to each other. Mainly to allow re-sending signals upon failure. 
- changes how peers track each other's state. everything is now consolidated into a `PeerState` struct which tells if the peer is muted, deafened, or recording. It looks like Uplink isn't currently using this information anyway. 

- Resolve #1409 

### Additional comments 🎤
Testing:
- offer a group call when one of the participants doesn't have uplink running. they should be able to turn on uplink and receive the call, and connect to everyone


sdwoodbury:
@coderabbitai pause

sdwoodbury:
making it harder for ppl to merge dev into this branch.

PR #1522, opened 2023-11-15
fix(call): Improv user ui in call

### LINKED ISSUE #1493: bug(ui): Members render off Call UI when too many members
**Issue:**
When you have 10+ users in the group call, the members will start to render off the call UI 

**Screenshot:**
Screenshot 2023-11-09 at 12 17 24 PM

Flemmli97:
### What this PR does 📖

- Fixes the problem where having a lot of users in a call makes it render off screen

- Resolve #1493


kit/src/components/context_menu/context.js
```diff
@@ -1,37 +1,53 @@
+let click = (toggle) => function (ev) {
```
sdwoodbury, line 1: i'm not familiar with this syntax. based off of [this](https://www.w3schools.com/Js/js_es6.asp#mark_arrow) it looks like the arrow can be used to make a lambda like this `(..) => {}` but I'm not familiar with `(..)  => function(..) {}`. is this valid javascript? where can I read more about it?

kit/src/components/context_menu/context.js
```diff
@@ -1,37 +1,53 @@
...
+  const menus = document.getElementsByClassName("context-menu")
+  // Select the current one
+  const context_menu = document.getElementById("UUID-context-menu")
+  let hidden = context_menu.classList.contains("hidden");
+  for (var i = 0; i < menus.length; i++) {
+    menus.item(i).classList.add("hidden")
+  }
+  if (toggle && !hidden) {
+    context_menu.classList.add("hidden")
+    return
```
sdwoodbury, line 14: this should probably have a boolean return type.

ui/src/components/media/calling.rs
```diff
@@ -681,15 +677,82 @@ pub struct CallUserImageProps {
...
+                    }
+                    Err(e) => {
+                        log::error!("eval receiver failed: {:?}", e);
+                    }
+                }
+            }
+        }
+    });
+    let visible_amount = *amount.get() as usize;
+    let (visible, context) = if visible_amount > cx.props.participants.len() {
```
sdwoodbury: this probably needs to be `visible_amount >= cx.props.participants.len()`

kit/src/components/context_menu/context.js
Flemmli97: its just a function creating a function.
so calling click(true) returns the function with toggle being true

kit/src/components/context_menu/context.js
Flemmli97: actually the end return is not needed instead so removed that

sdwoodbury (approved):

phillsatellite (approved):

stavares843 (approved):

PR #1528, opened 2023-11-17
fix(QuickProfile): Only show quick profile if a profile is selected

InfamousVague:
### What this PR does 📖
The quick profile was being rendered even when there was no profile selected, which was causing a space below the chatbar. This only renders the quick profile if the data is some. This previously caused an empty space below the chatbar.

Before, you can see the space below the chatbar.:
image


phillsatellite:
Got permission from Matt to close this PR since Flem has opened up one earlier to fix 💯

PR #1533, opened 2023-11-21

dariusc93:
### What this PR does 📖

- Update dependency
- Enable quic protocol
- Increase the max storage size to 10GB 
- Disable discovery by default
- Add `discovery` flag

- N/A

- This PR includes an update to the lock file to fix the debug assert found upstream
- ~~Calling may not work since https://github.com/Satellite-im/Warp/pull/358 was not merged so a segment of code was commented out related to blink~~
- Online indication may have a delay when a users come on and offline. This may be updated outside of the PR since it is related to the relay. 

### Additional comments 🎤
- Right now, discovery flag is marked to be disabled so it would not rely on public services (eg dht) but instead direct connections, which at this point would require the use of the public key to discover the identity of another connected peer.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -441,24 +441,24 @@ dependencies = [
 
 [[package]]
 name = "async-channel"
-version = "2.1.0"
+version = "2.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d37875bd9915b7d67c2f117ea2c30a0989874d0b2cb694fe25403c85763c0c9e"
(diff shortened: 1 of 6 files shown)


luisecm:
The tests for macos/windows are failing on this PR since there is one test validating the Max Size value on Files Screen expecting the value 1 GB, which is being increased to 10GB. I can make the update to tests once that this PR is merged @stavares843 @phillsatellite

sdwoodbury (approved):

phillsatellite:
Just wanted to add some notes, Ive been testing with Darius over the weekend and have been noticing some signaling issues with users not appearing online/offline. Sent him some logs and he has been making some adjustments 👍

phillsatellite (approved):

PR #1543, opened 2023-11-22
feat(chat): User mentioning

### LINKED ISSUE #1460: chat(feat): Add ability to tag friend in chat
**Task:**
We want to be able to tag people in chat and them get notified for it so if someone was to tag me it would look like "hey @phill" and I would get a notification that someone directly tagged me in the chat 

**Screenshot:**
Screenshot 2023-11-06 at 4 31 57 PM

Flemmli97:
### What this PR does 📖

- Allows users to `@user` another user
- Rewrites chat suggestion to allow more easily adding additional suggestion types
- When user gets tagged notification sound plays regardless of focus etc.
- Missing:  
 ~~- clicking on the tagged field~~
  - a component containing all mentions? (maybe separate PR for this? how/do we want to handle this?)

- Resolve #1460


common/src/state/mod.rs
```diff
@@ -436,8 +438,17 @@ impl State {
...
+                mut message,
             } => {
+                if message.inner.lines().iter().any(|s| s.contains('@')) {
+                    if let Some(ids) = self
+                        .get_chat_by_id(conversation_id)
+                        .map(|c| self.chat_participants(&c))
+                    {
+                        message.resolve_message(&ids, &self.get_own_identity().did_key());
+                    }
+                }
```
sdwoodbury, line 450: I think it better if the check for the `@` character was moved into the `resolve_message()` function. I think the function could be named better too. recommend something like `insert_did()`

common/src/state/mod.rs
```diff
@@ -436,8 +438,17 @@ impl State {
...
+                let ping = message.is_mention;
```
sdwoodbury, line 449: not a blocker but i recommend using `message.is_mention` in-line and not adding this extra variable because that's one more thing we have to name.

common/src/state/mod.rs
sdwoodbury: I think it better if the check for the `@` character was moved into the `resolve_message()` function. I think the function could be named better too. recommend something like `insert_did()`

common/src/state/chats.rs
```diff
@@ -53,6 +53,8 @@ pub struct Chat {
     // Unread count for this chat, should be cleared when we view the chat.
     #[serde(default)]
     unreads: HashSet<Uuid>,
+    #[serde(default, skip)]
+    pub message_pings: VecDeque<ui_adapter::Message>,
```
sdwoodbury: i think this should be called `mentions`. 

also I have a question - what will this be used for? I'm asking because I see it updated when messages are added/removed from a chat but I don't see it actually used yet. Also the [ticket](https://github.com/Satellite-im/Uplink/issues/1460) doesn't say much about this besides the user will get a notification, which they would receive anyway for any chat when a message is received.

common/src/warp_runner/ui_adapter/mod.rs
```diff
@@ -34,16 +34,39 @@ use super::{
...
+    pub is_mention: bool,
     /// this field exists so that the UI can tell Dioxus when a message has been edited and thus
     /// needs to be re-rendered. Before the addition of this field, the compose view was
     /// using the message Uuid, but this doesn't change when a message is edited.
     pub key: String,
 }
 
+impl Message {
+    // resolve the message lines to e.g. format user mentions correctly
+    pub fn resolve_message(&mut self, c: &[state::Identity], own: &DID) {
```
sdwoodbury: suggested a different name  for this function in previous comment

kit/src/components/message/mod.rs
```diff
@@ -68,6 +68,8 @@ pub struct Props<'a> {
     // An optional field that, if set, will be used as the text content of a nested p element with a class of "text".
     with_text: Option<String>,
 
+    with_text_to_render: Option<String>,
```
sdwoodbury: I dislike this name. it's basically the same as `with_text`. I feel that the name doesn't adequately describe what the code does. 

after looking at how this is used in `render_message()`, I believe this variable is unnecessary.

kit/src/layout/chatbar/mod.rs
```diff
@@ -14,6 +15,34 @@ use common::{icons, language::get_local_text, warp_runner::thumbnail_to_base64};
 
 pub type To = &'static str;
 
+pub enum SuggestionType {
```
sdwoodbury, line 18: I'm surprised that clippy doesn't complain about enumerations that are all caps

sdwoodbury (changes requested):

kit/src/components/message/mod.rs
Flemmli97: we need this one for e.g. editing as the text rendered differs from the text that should be edited.
the rendered text contains html formatting elements

common/src/state/chats.rs
Flemmli97: initially it was supposed to be used to display a list of mentioned messages. but its atm not used except for simply tracking the mentioned messages

common/src/state/mod.rs
Flemmli97: reason is simply for future usage/changes. if we want to do some other preprocessing before rendering in the future  we can just put it in this function.
but ye for that the `@` check should be moved into this

common/src/state/mod.rs
Flemmli97: its cause the message is moved in https://github.com/Satellite-im/Uplink/pull/1543/files/5f9838315aedf0e3f5c3d43a957aa9b6baa846a8#diff-aa2878cb5d6011c19cff5595a21131f6920da8d255b0ec874ab6237ff9349d9bR455 before using `is_mention`

kit/src/components/message/mod.rs
sdwoodbury: ok I understand now. I can't think of a better way to do this at the moment though it would be nice if we could avoid storing messages in an intermediate form. For now i'll request it be renamed to `tagged_text` because that's what it is.

sdwoodbury (approved):

phillsatellite:
@Flemmli97 super cool PR! I noticed it doesnt show the @username in the sidebar instead it will show the entire did. Would this be easier to handle in a separate ticket? I could write up if so

Screenshot 2023-11-30 at 10 31 31 AM

Flemmli97:
> @Flemmli97 super cool PR! I noticed it doesnt show the @username in the sidebar instead it will show the entire did. Would this be easier to handle in a separate ticket? I could write up if so
> 
> Screenshot 2023-11-30 at 10 31 31 AM

do we want the sidebar to display the message in chat 1:1 though? as in with that blue background and all?
if not we might need another component...

phillsatellite:
> > @Flemmli97 super cool PR! I noticed it doesnt show the @username in the sidebar instead it will show the entire did. Would this be easier to handle in a separate ticket? I could write up if so
> > Screenshot 2023-11-30 at 10 31 31 AM
> 
> do we want the sidebar to display the message in chat 1:1 though? as in with that blue background and all? if not we might need another component...

Checked with Matt and we can do a 1:1 in the sidebar 👍  which would work better for you handling that in this pr or a new ticket?

Flemmli97:
this pr would be better

PR #1554, opened 2023-11-28
feat(sidebar): move chat to top of sidebar when call is offered/answered

### LINKED ISSUE #1553: feat(sidebar): joining a call should send that chat to the top of the sidebar
what the title says. I noticed this when testing calling. I called someone but didn't send any messages. I tried to call them again and had to scroll through the sidebar to find them again. It would have been nicer if the chat had been sent to the top of the sidebar

sdwoodbury:
### What this PR does 📖

- when a call is incoming, move the chat to the top of the sidebar, without selecting it. 
- when a call is answered, also move it to the top of the sidebar. 
- when a call is offered (note that the caller already has to click the chat to make this happen), make that chat the active call, which moves it to the top of the sidebar. 

- Resolve #1553


common/src/state/mod.rs
```diff
@@ -649,6 +651,7 @@ impl State {
                         return;
                     }
                 };
+                self.send_chat_to_top_of_sidebar(conversation_id);
```
lgmarchi, line 676: @sdwoodbury If I reduse the call, chat will back for its original position?

sdwoodbury: no and i think that such behavior would be most unpleasant.

phillsatellite (approved):

PR #1558, opened 2023-11-29
update(PasteFiles): Fix paste multiple times

### LINKED ISSUE #1286: bug(general): Pasting from clipboard will paste twice or 4 times
**Issue:**
When you copy an imagine from clipboard and paste it into chat, it will paste twice. On Windows it will paste it 4 times

**OS:**
Windows 
MacOS 

**Screen recording:**

https://github.com/Satellite-im/Uplink/assets/93608357/897fed8d-c1ce-4139-97cb-e15b355ee10b

lgmarchi:

- Resolve #1286


phillsatellite (approved):

sdwoodbury (commented):
could the problem be that there were 2 `paste_files_shortcut` and that the `use_future` added on line 72 of `paste_files_with_shortcut/mod.rs` isn't needed?

lgmarchi:
> could the problem be that there were 2 `paste_files_shortcut` and that the `use_future` added on line 72 of `paste_files_with_shortcut/mod.rs` isn't needed?

This is the first problem, so when I pasted, it was pasting 4 times, we had duplicated code. But after remove one code, it was pasting 2 times, so yet pasting more than one time. 

This use_future, is just a hack to guarantee user will paste just one time, as temporary solution, I am figuring out a better solutin without it.

sdwoodbury (approved):

PR #1568, opened 2023-12-01
chore(Logger): Added primer for log level and state tabs in logger

InfamousVague:
### What this PR does 📖

Added additional comments in the PR to describe the additional functionality added 
image


InfamousVague:
image

TODO: Need to watch state here and update when changes happen

InfamousVague:
image
image
Also included a quick switcher between dark & light modes in the dev console, this should make it easier to test new components against both

ui/src/components/debug_logger/mod.rs
```diff
@@ -24,6 +47,13 @@ pub fn DebugLogger(cx: Scope) -> Element {
 
     let eval = use_eval(cx);
 
+    let active_tab: &UseState<String> = use_state(cx, || "Logs".into());
```
sdwoodbury: this should be an enumeration, not a string.

ui/src/components/debug_logger/mod.rs
```diff
@@ -34,49 +64,165 @@ pub fn DebugLogger(cx: Scope) -> Element {
...
+                        onpress: |_| {
+                            window.webview.open_devtools();
+                        }
+                    },
+                },
+                div {
+                    class: "logger-nav-right",
+                    aria_label: "debug-logger-nav-right",
+                    IconElement {
+                        icon: if state.read().ui.theme.clone().unwrap_or_default().name == "Light" {
```
sdwoodbury, line 159: fyi this would be more elegant: `ui.theme.map(|theme| theme.name == "Light").unwrap_or_default()`

common/src/state/mod.rs
```diff
@@ -742,6 +742,12 @@ impl State {
         fs::write(path, serialized)?;
         Ok(())
     }
+
+    pub fn get_json(&self) -> Result<String, Box<dyn std::error::Error>> {
```
sdwoodbury: unless the error is going to get logged or handled, this function may as well return a String and use `unwrap_or_default()`.

sdwoodbury (commented):
it might be helpful if the `Log` struct has a `from_str` implementation. This would make it easier to filter `logs_to_show` because it could then be a `Log` instead of a `String`.

InfamousVague:
image
Fixed color mapping to messages

common/src/state/mod.rs
```diff
@@ -743,9 +743,9 @@ impl State {
         Ok(())
     }
 
-    pub fn get_json(&self) -> Result<String, Box<dyn std::error::Error>> {
-        let serialized = serde_json::to_string_pretty(self)?;
-        Ok(serialized)
+    pub fn get_json(&self) -> String {
+        let serialized = serde_json::to_string_pretty(self).unwrap_or("".into());
```
dariusc93:
```suggestion
        let serialized = serde_json::to_string_pretty(self).unwrap_or_default();
```

InfamousVague:
This includes various small fixes and UI tweaks across the application @phillsatellite @stavares843 I'm not sure if there is anything specific to test outside of making sure it all works well and trying out different variations of things to ensure I haven't broken anything :)

tooshel (approved):
It works! It was a little confusing that it defaults to errors but I like the logic behind it (that you want to always see errors and don't want them hidden).

ui/src/components/debug_logger/mod.rs
```diff
@@ -34,49 +64,179 @@ pub fn DebugLogger(cx: Scope) -> Element {
...
+                },
+            },
+            match active_tab.get() {
+                Tab::Logs => rsx!(div {
+                    aria_label: "debug-logger-body",
+                    class: "body",
+                    div {
+                        class: "body-scroll",
+                        logs_to_show.iter().filter(
+                            |&x| &x.level == filter_level.get() || filter_level.get() == &Level::Debug
```
dariusc93: This is more of a nik pik, but if `Level` derives `Copy` you can dereference it instead. 

```suggestion
                            |&x| x.level == *filter_level.get() || *filter_level.get() == Level::Debug
```

ui/src/components/debug_logger/mod.rs
```diff
@@ -34,49 +64,179 @@ pub fn DebugLogger(cx: Scope) -> Element {
...
+                                        text: "Trace".into()
+                                    }
+                                )),
+                            },
+                        }
+                    })),
+                    Button {
+                        text: "State".into(),
+                        icon: Icon::Square3Stack3d,
+                        appearance: if active_tab.get() == &Tab::State { Appearance::Primary } else { Appearance::Secondary },
```
dariusc93:
```suggestion
                        appearance: if *active_tab.get() == Tab::State { Appearance::Primary } else { Appearance::Secondary },
```

ui/src/components/debug_logger/mod.rs
```diff
@@ -34,49 +64,179 @@ pub fn DebugLogger(cx: Scope) -> Element {
...
+                                tooltip: cx.render(rsx!(
+                                    Tooltip {
+                                        arrow_position: ArrowPosition::Top,
+                                        text: "Info".into()
+                                    }
+                                )),
+                            },
+                            Button {
+                                icon: Icon::ExclamationTriangle,
+                                appearance: if filter_level.get() == &Level::Error { Appearance::Danger } else { Appearance::Secondary },
```
dariusc93:
```suggestion
                                appearance: if *filter_level.get() == Level::Error { Appearance::Danger } else { Appearance::Secondary },
```

ui/src/components/debug_logger/mod.rs
```diff
@@ -34,49 +64,179 @@ pub fn DebugLogger(cx: Scope) -> Element {
...
+                                tooltip: cx.render(rsx!(
+                                    Tooltip {
+                                        arrow_position: ArrowPosition::Top,
+                                        text: "Debug".into()
+                                    }
+                                )),
+                            },
+                            Button {
+                                icon: Icon::InformationCircle,
+                                appearance: if filter_level.get() == &Level::Info { Appearance::Info } else { Appearance::Secondary },
```
dariusc93:
```suggestion
                                appearance: if *filter_level.get() == Level::Info { Appearance::Info } else { Appearance::Secondary },
```

ui/src/components/debug_logger/mod.rs
```diff
@@ -34,49 +64,179 @@ pub fn DebugLogger(cx: Scope) -> Element {
...
-                                span {
-                                    aria_label: "debug-logger-item-timestamp",
-                                    class: "log-text muted",
-                                    "〇 {log_datetime}"
+                    class: "logger-nav",
+                    aria_label: "debug-logger-nav",
+                    Button {
+                        text: "Logs".into(),
+                        icon: Icon::CommandLine,
+                        appearance: if active_tab.get() == &Tab::Logs { Appearance::Primary } else { Appearance::Secondary },
```
dariusc93:
```suggestion
                        appearance: if *active_tab.get() == Tab::Logs { Appearance::Primary } else { Appearance::Secondary },
```

ui/src/components/debug_logger/mod.rs
```diff
@@ -1,29 +1,59 @@
...
+
+use dioxus_desktop::use_window;
+use log::Level;
 
 use crate::logger;
 
 const STYLE: &str = include_str!("./style.scss");
 const SCRIPT: &str = include_str!("./script.js");
 
+#[derive(PartialEq, Eq)]
```
dariusc93:
```suggestion
#[derive(PartialEq, Eq, Copy, Clone)]
```

ui/src/components/friends/add.rs
```diff
@@ -189,44 +189,88 @@ pub fn AddFriend(cx: Scope) -> Element {
...
+                            }
+                        },
+                        ContextItem {
+                            icon: Icon::ClipboardDocument,
+                            aria_label: "friend-add-input-paste".into(),
+                            text: get_local_text("uplink.paste"),
+                            onpress: move |_| {
+                                match Clipboard::new() {
+                                    Ok(mut c) => {
+                                        let text = c.get_text().unwrap_or_default();
```
dariusc93: In the event there is an error using this function, we should log so we would know in case it comes back to uses the default string due to an error.

```suggestion
                                        let text = c.get_text().map_err(|e| {
                                             log::error!("Unable to get clipboard content: {e}");
                                             e
                                        }).unwrap_or_default();
```

dariusc93 (commented):
Left some suggestions to avoid creating references where its not needed.

ui/src/components/debug_logger/mod.rs
```diff
@@ -34,49 +64,179 @@ pub fn DebugLogger(cx: Scope) -> Element {
...
+                            Icon::Moon
+                        }
+                    },
+                    Switch {
+                        active: state.read().ui.theme.clone().unwrap_or_default().name == "Light",
+                        onflipped: move |_| {
+                            let current_theme = state.read().ui.theme.clone().unwrap_or_default();
+
+                            if current_theme.name != "Light" {
+                                let light_theme = get_available_themes().iter().find(|t| t.name == "Light").unwrap().clone();
```
dariusc93: Is there a change of the themes not containing `Light`? If so, it might be better to avoid unwrapping to prevent panicking. Instead, we could probably either use `cloned()` on the `Option` it returns and supply it below or use a if-let condition. If it is going to be available, then it could be left as is but could do something like this 

```suggestion
                                let light_theme = get_available_themes().iter().find(|t| t.name == "Light").cloned().expect("theme is available");;
```

stavares843:
the tests are failing due to the change on the copy

image


it's expected and we will address when this gets merged

phillsatellite (approved):

PR #1571, opened 2023-12-01
feat(settings): Impl online status change setting

### LINKED ISSUE #1215: feat(settings): add ability to change online status
**Task:** Just like in Core we want to let the User have the ability to change they're online status. This was located at the bottom of the sidebar, if the User clicked they're avatar a menu would appear with the option to log out of the app and also to change the online status to _Online_ , _Away_ , _Busy_ , _Offline_

**Screenshot**
Change Status

Flemmli97:
### What this PR does 📖

- Adds ability to change your online settings. This is done by going to the profile settings

- Resolve #1215


phillsatellite:
Noticed a couple minor things 

- It looks like Do Not Disturb overlaps Offline
Screenshot 2023-12-01 at 2 16 25 PM

- If user A is set to idle (or DND, Offline) and they log out and log back in, they will appear as online even though status is still set to Idle

Picture shows after logging back in again status is still set to idle 
Screenshot 2023-12-01 at 2 22 32 PM

But on User B side its appearing as online again 
Screenshot_2023-12-01_142205

common/src/warp_runner/manager/commands/multipass_commands.rs
```diff
@@ -425,6 +430,28 @@ pub async fn handle_multipass_cmd(cmd: MultiPassCmd, warp: &mut super::super::Wa
                 Err(err) => Err(err),
             };
             let _ = rsp.send(r);
+        }
+        MultiPassCmd::SetOnlineStatus { status, rsp } => {
```
dariusc93:
```suggestion
        MultiPassCmd::SetStatus { status, rsp } => {
```

common/src/warp_runner/manager/commands/multipass_commands.rs
```diff
@@ -123,6 +123,11 @@ pub enum MultiPassCmd {
         did: DID,
         rsp: oneshot::Sender<Result<Identity, warp::error::Error>>,
     },
+    #[display(fmt = "SetOnlineStatus")]
+    SetOnlineStatus {
```
dariusc93:
```suggestion
    #[display(fmt = "SetStatus")]
    SetStatus {
```

ui/src/components/settings/sub_pages/profile/mod.rs
```diff
@@ -101,14 +113,15 @@ pub fn ProfileSettings(cx: Scope) -> Element {
...
-                        status: Some(status),
+                    ChanCmd::StatusMessage(status) => MultiPassCmd::UpdateStatus {
+                        status: if status.is_empty() {
+                            None
+                        } else {
+                            Some(status)
+                        },
                         rsp: tx,
                     },
+                    ChanCmd::Status(status) => MultiPassCmd::SetOnlineStatus { status, rsp: tx },
```
dariusc93:
```suggestion
                    ChanCmd::Status(status) => MultiPassCmd::SetStatus { status, rsp: tx },
```

dariusc93 (commented):
I suggest changing the command name to `SetStatus`, mostly to prevent confusion since `SetOnlineStatus` sounds like the status is being set to online.

dariusc93:
>     * If user A is set to idle (or DND, Offline) and they log out and log back in, they will appear as online even though status is still set to Idle

This is likely due to the events being emitted as when the user comes online, warp emits `MultiPassEventKind::IdentityOnline`, which is likely overriding the status being set on uplink rather than go based on the set status. A workaround would be to check `MultiPass::identity_status` after they come online and forward that status to the state along with the `Identity` wrapper in uplink. I think on warp side, i could refactor the event to include the status, if its not set to offline.

Flemmli97:
> Noticed a couple minor things
> 
> * It looks like Do Not Disturb overlaps Offline
> 
> Screenshot 2023-12-01 at 2 16 25 PM
> * If user A is set to idle (or DND, Offline) and they log out and log back in, they will appear as online even though status is still set to Idle
> 
> Picture shows after logging back in again status is still set to idle Screenshot 2023-12-01 at 2 22 32 PM
> 
> But on User B side its appearing as online again Screenshot_2023-12-01_142205

resolved

Flemmli97:
> > ```
> > * If user A is set to idle (or DND, Offline) and they log out and log back in, they will appear as online even though status is still set to Idle
> > ```
> 
> 
> This is likely due to the events being emitted as when the user comes online, warp emits `MultiPassEventKind::IdentityOnline`, which is likely overriding the status being set on uplink rather than go based on the set status. A workaround would be to check `MultiPass::identity_status` after they come online and forward that status to the state along with the `Identity` wrapper in uplink. I think on warp side, i could refactor the event to include the status, if its not set to offline.

`MultiPassEventKind::IdentityOnline` actually tries to get the status from warp and forwards it to state. but state just ignored the status and set it to online always. 
i now made it so state uses the resolved status which should hopefully fix this

ui/src/components/settings/sub_pages/profile/mod.rs
```diff
@@ -538,3 +564,22 @@ fn get_input_options(validation_options: Validation) -> Options {
         ..Options::default()
     }
 }
+
+fn get_status_option<'a>(cx: Scope<'a>, status: &IdentityStatus) -> (String, Element<'a>) {
```
dariusc93, line 568: `IdentityStatus` derives `Copy` so you probably shouldnt need to have it here as a reference. Doesnt hurt anything though.

phillsatellite (approved):

ui/src/components/settings/sub_pages/profile/mod.rs
```diff
@@ -538,3 +564,22 @@ fn get_input_options(validation_options: Validation) -> Options {
...
+    let indicator = Status::from(*status);
+    (
+        serde_json::to_string::<IdentityStatus>(status).unwrap_or_default(),
```
dariusc93, line 571: Since `IdentityStatus` implements `Display` (indirectly), you could use `status.to_string()` instead.

ui/src/components/settings/sub_pages/profile/mod.rs
```diff
@@ -538,3 +564,22 @@ fn get_input_options(validation_options: Validation) -> Options {
...
+        cx.render(rsx!(div {
+                class: "settings-online-status",
+                Indicator {
+                    status: indicator,
+                    platform: Platform::Unknown
```
dariusc93, line 576: Though only implemented in mock data (to my knowledge anyway - not sure if this changed since the last time it was ever used) you could probably passthrough the profile platform itself rather than leaving it to unknown.

ui/src/components/settings/sub_pages/profile/mod.rs
Flemmli97: i usually try to always use refs when possible. its prob micro optimization but prevents unnecessary clones

ui/src/components/settings/sub_pages/profile/mod.rs
Flemmli97: that way though you cant get it back from the string. as to_string and deserialization provide different strings

dariusc93: Yea they both would be different (eg from `serde_json` the string would contain quotes and may be formatted depending on the attr macro). Might be overlooking it though but i dont see anyplace where you would be converting it back into a `IdentityStatus`.

Flemmli97: when selecting it parses it back to `IdentityStatus` 
at https://github.com/Satellite-im/Uplink/pull/1571/files/68489ff7fb34364820dc0db7d3ffbee6564be160#diff-993f168fccb56d7662f6c89e41a0ff85392c38375ac2e8e9d1d230bcc580ddb9R449

ui/src/components/settings/sub_pages/profile/mod.rs
Flemmli97: Screenshot 2023-12-05 at 19 31 13

there are only these platforms and i simply choose the circle icon one since that one is the most generic one.
and should better reflect it being in settings than the others

dariusc93 (approved):

stavares843 (approved):

stavares843:
very cool! 🎨

PR #1574, opened 2023-12-04
feat(blink): improve call quality

### LINKED ISSUE #1409: bug(calling): sometimes users can't hear each other

### LINKED ISSUE #1474: bug(calling): Latency when in a call
**Issue:**
When you are on a call with someone, sometimes their voice will cutout for 10 seconds or so and then they will cut back in at the same part they cut out at. Its almost like there is a network latency but there isnt anything wrong with the network 

**OS:**
All

sdwoodbury:
### What this PR does 📖

- use the latest version of Warp, which includes a breaking update to Blink implementation. It listens on different gossipsub channels, meaning that anyone using the old software won't be able to participate in calls with peers who are using newer versions of the software. 
- update logging so that blink logs are included in debug logs, and added an option to `trace-blink`. 
- updated the way call state (muted/unmuted/etc) is tracked for participants. 

- Resolve #1409 
- Resolve #1474 

### Additional comments 🎤
wait for https://github.com/Satellite-im/Uplink/pull/1569 to be merged first


common/src/state/call.rs
```diff
@@ -113,36 +112,36 @@ impl CallInfo {
...
-        if active_call.call.id != call_id {
-            bail!("wrong call id");
-        }
-        active_call.call.participant_joined(id);
+        active_call.call.participant_speaking(id);
         Ok(())
     }
 
-    pub fn participant_left(&mut self, call_id: Uuid, id: &DID) -> anyhow::Result<()> {
+    pub fn participant_joined(&mut self, call_id: Uuid, id: DID) -> anyhow::Result<()> {
```
lgmarchi, line 124: Stu, following clean code, could we change to handle_participant_speaking? Functions are supposed to start with a verb. Same thing for other functions here.

ui/src/bootstrap.rs
```diff
@@ -89,6 +89,7 @@ pub fn configure_logger(profile: Option<LogProfile>) {
     let max_log_level = if let Some(profile) = profile {
         match profile {
             LogProfile::Debug => {
+                logger::allow_other_crates(Level::Debug, Some(&["warp-blink-wrtc"]));
```
lgmarchi: Could we let "warp-blink-wrtc", as const variable in the page and repeat it in the other places?

ui/src/logger/mod.rs
```diff
@@ -56,10 +56,8 @@ pub struct Logger {
     max_logs: usize,
     write_to_stdout: bool,
 
-    // display trace logs from uplink
-    uplink_trace: bool,
     // minimum log level for 3rd party crates (warp, dioxus, etc)
```
lgmarchi, line 59: Need to change the comment, to be equal variable context.

lgmarchi (approved):
Left some comments to improve code quality.

common/src/lib.rs
```diff
@@ -48,6 +48,8 @@ pub enum LogProfile {
     DebugAll,
     /// Logs trace level from all crates to a file
     TraceAll,
+    /// maximum logging for blink
+    TraceBlink,
 }
```
zeenix, line 53: You're just adding another log option here so this comment is more generic: typically, people just make use of the `RUST_LOG` env, where you can set different log levels for different crates (and even mods). Was there a reason for going with custom/specific method?

common/src/state/call.rs
```diff
@@ -37,9 +37,8 @@ pub struct Call {
     pub id: Uuid,
     pub conversation_id: Uuid,
     pub participants: Vec<DID>,
-    pub participants_joined: Vec<DID>,
+    pub participants_joined: HashMap<DID, ParticipantState>,
     pub participants_speaking: HashMap<DID, Instant>,
-    pub participants_state: HashMap<DID, ParticipantState>,
```
zeenix, line 42: You did a great job with PR description (unlike me :laughing:) so I hate to ask this but it would be good to summarize the specific changes like this one so people new to the code (like me) can more easily understand what's going on. It will also help future archeologist (which may be you :smile:).

common/src/state/mod.rs
```diff
@@ -335,9 +336,18 @@ impl State {
...
+                | WarpEvent::Message(MessageEvent::TypingIndicator { .. })
         ) {
             log::debug!("process_warp_event: {event}");
+        } else if matches!(
+            event,
+            WarpEvent::Message(MessageEvent::TypingIndicator { .. })
+        ) {
+            {
+                log::trace!("process_warp_event: {event}");
+            }
```
zeenix: why the extra `{}`? :thinking:

common/src/state/mod.rs
```diff
@@ -720,14 +730,14 @@ impl State {
...
-            }
             BlinkEventKind::ParticipantStateChanged { peer_id, state } => {
-                if let Err(e) = self.ui.call_info.update_partcipant_state(peer_id, state) {
-                    log::error!("{e}");
+                if let Err(e) = self.ui.call_info.update_participant_state(peer_id, state) {
+                    log::error!("failed to process ParticipantStateChanged event : {e}");
                 }
             }
+            BlinkEventKind::AudioStreamError => {
+                // todo
```
zeenix, line 741: might be good to log a debug or trace here.

ui/src/components/media/calling.rs
```diff
@@ -101,7 +101,7 @@ pub struct ActiveCallProps {
 fn ActiveCallControl(cx: Scope<ActiveCallProps>) -> Element {
```
zeenix, line 101: unrelated to your PR but this should be `activate_call_control`.

dariusc93: This was done to match the same style as dioxus macro components

common/src/lib.rs
dariusc93: To my knowledge, the custom method was done so it would allow routing the logs to the component in uplink vs just outputting it to a file or to terminal

zeenix: hmm.. I'm not an expert on logging but aren't the loglevel control and output routing decoupled in Rust logging/tracing crates?

ui/src/components/media/calling.rs
zeenix: interesting, do you have a link to some example?

dariusc93: If you mean within uplink, then the function [CallUserImageGroup](https://github.com/Satellite-im/Uplink/blob/a7c216be2bf0948fd14ca7688d7c6798caf0431b/ui/src/components/media/calling.rs#L679) is used [here](https://github.com/Satellite-im/Uplink/blob/a7c216be2bf0948fd14ca7688d7c6798caf0431b/ui/src/components/media/calling.rs#L383-L385). In dioxus, you can probably see it more in some of their examples (eg https://github.com/DioxusLabs/dioxus/blob/b182a6ee75c3f69d5498de3b6d8122b94c8c8fbc/examples/borrowed.rs)

common/src/lib.rs
dariusc93: To my knowledge they are, but these are more specific to the [logging component](https://github.com/Satellite-im/Uplink/blob/dev/ui/src/components/debug_logger/mod.rs#L13) in uplink.

common/src/state/call.rs
dariusc93: Generally, it is usually more common to have the function contain a verb than to start with one when they are performing a action that modifies data or context (in this case, the function is mutable with it showing "joined"). However,  that usually comes down to the overall coding style.

sdwoodbury: @lgmarchi  I think the verb "handle" is too generic to be useful. Also, this function updates state when someone joins a call, not when they speak, so "speaking" shouldn't be in the function name either. I respectfully reject this proposal.

ui/src/logger/mod.rs
sdwoodbury: @lgmarchi I don't understand what comment needs to be changed and why. perhaps bad grammar is causing some confusion.

common/src/lib.rs
zeenix: > To my knowledge they are, but these are more specific to the [logging component](https://github.com/Satellite-im/Uplink/blob/dev/ui/src/components/debug_logger/mod.rs#L13) in uplink.

That's the question: why are they specific to Uplink? Sorry, I wasn't clear.

sdwoodbury: @zeenix  is there a way to make the logger emit the trace logs for one crate and ignore the trace logs for all other crates using environment variables? So far I only know how to use `RUST_LOG` to set the same log level for all crates.

common/src/state/call.rs
sdwoodbury: @zeenix I combined lines 40 and 42. I thought it unnecessary to maintain a vec of participants who joined and also a `HashMap<DID, ParticipantState>` because that map only contains state for participants who have joined the call.

common/src/state/mod.rs
sdwoodbury: i'll remove it

ui/src/components/media/calling.rs
zeenix: Thanks, that's super strange. I've never seen any Rust project make an exception to this convention.

common/src/lib.rs
zeenix: > @zeenix is there a way to make the logger emit the trace logs for one crate and ignore the trace logs for all other crates using environment variables?

Yes, sir. As I mentioned, you can even set different levels for mods. E.g `RUST_LOG=tokio=info,uplink=trace,zbus=debug,zbus::connection=trace` sets different levels for different crates and then sets a different level for `connection` mod of `zbus` than rest of `zbus`.

common/src/state/call.rs
zeenix: @sdwoodbury Thanks! Now please also help the future archeologists. :)

common/src/lib.rs
sdwoodbury: @zeenix  this does look a lot better. we could probably get rid of most of the logging code and just use RUST_LOG. maybe it's better to do that in a separate PR.

zeenix: > maybe it's better to do that in a separate PR.

Definitely a different PR. :)

phillsatellite:
Wanted to leave a comment so everyone knew where we were at. @tooshel @stavares843 and I tested today while doing a group call. It sounded very clear and there was no gargling noise but we did notice that there was some major delay during the conversation. I would say something to Sheldon and about 5 seconds later I would hear what I said through Sheldons speakers. There were other times when Sheldons sentence would get cut-off and I would just miss half of what he was saying. Nonetheless this is still working better than dev in my opinion and would probably be worth merging in and making further improvements in new tickets. Tomorrow morning @tooshel @stavares843 and I were going to do another call while recording it and share it here so we can better explain the issues that we faced. 

Added changes requested till Sheldon/Sara and I talk tomorrow morning to see if we want to merge it in

ui/src/logger/mod.rs
lgmarchi: You changed the name of variable for max_level, and comment is as minimum log level for 3rd party

phillsatellite (approved):

ui/src/logger/mod.rs
stavares843: marking this as resolved to get this merged but in case you wanna address @lgmarchi comment above cc @sdwoodbury

PR #1587, opened 2023-12-05
refactor(PasteShortcut): Fix paste for Linux OS System with shortcut

lgmarchi:
### 1. Fix paste for Linux OS System with shortcut

https://github.com/Satellite-im/Uplink/assets/63157656/9eed39c7-22ff-4091-863f-f941ed2dfed3

- Resolve #


phillsatellite (approved):

ui/src/layouts/chats/mod.rs
```diff
@@ -95,6 +98,46 @@ pub fn ChatLayout(cx: Scope) -> Element {
...
+                                    .map(|path| Location::Disk { path: path.clone() })
+                                    .collect();
+                                let mut current_files: Vec<_> = state
+                                    .read()
+                                    .get_active_chat()
+                                    .map(|f| f.files_attached_to_send)
+                                    .unwrap_or_default()
+                                    .drain(..)
+                                    .filter(|x| !new_files.contains(x))
+                                    .collect();
```
sdwoodbury, line 129: although this seems fine to me, it would reorder the attachments if you copy/paste files which are already attached, by moving the files to the end of the list. It may be even better to filter `files_local_path` (not the best name btw - `pasted_files` or `to_attach` would be better) by `files_attached_to_send` instead.

sdwoodbury:
https://github.com/Satellite-im/Uplink/assets/8636602/ac53611c-f29e-45ea-874b-aa6b7a6cea91

sometimes the filename gets pasted to the chatbar and the image is simultaneously attached.

ui/src/layouts/chats/presentation/chatbar/mod.rs
```diff
@@ -302,6 +305,35 @@ pub fn get_chatbar<'a>(cx: &'a Scoped<'a, ChatProps>) -> Element<'a> {
...
+                        .map(|f| f.files_attached_to_send)
+                        .unwrap_or_default()
+                        .drain(..)
+                        .filter(|x| !new_files.contains(x))
+                        .collect();
+                        current_files.extend(new_files);
+                    state
+                        .write()
+                        .mutate(Action::SetChatAttachments(active_chat_id, current_files));
+                    }
```
sdwoodbury, line 333: it would result in less code duplication if `Action::SetChatAttachments` (maybe make a new action called `Action::AppendChatAttachments`) performed the filtering of the pasted files.

sdwoodbury (approved):
seems better than before, though there's still room for improvement. 

also, it seems like the paste event is being caught by multiple elements at once - the chatbar and the chats view I think. in another comment I attached a video.

lgmarchi:
> seems better than before, though there's still room for improvement.
> 
> also, it seems like the paste event is being caught by multiple elements at once - the chatbar and the chats view I think. in another comment I attached a video.

It is not been caught multiple times, when chatbar is focused, onkeydown from Chats Layout is ignored. Because of this to paste if chatbar is focused and when it is not focused, I need code in both places.

ui/src/layouts/chats/presentation/chatbar/mod.rs
lgmarchi: You have a good point, but maybe not a new action, but filter code inside SetChatAttachments, and we can improve this name as well. And I can define if files are coming from local or Constellation, filters are a little bit different.

stavares843:
marked conversations above as resolved as lucas will address them in another PR 

ticket https://github.com/Satellite-im/Uplink/issues/1607

PR #1609, opened 2023-12-11

dariusc93:
### What this PR does 📖

- Internally add a timestamp to incoming and outgoing requests
- Add a placeholder for announcing identities in the network
- Initialize CallData with your own public key 
- Added signaling to `File` and `Directory` to trigger exporting

- N/A

- ~~This should not require a new account as requests are migrated to the new format, however this is not backwards compatible to anything prior to this PR so if moving back in testing, the account may lose its requests it has.~~ Due to a new internal format, a new account is recommended as it would not be backward compatible with any structure prior to this branch.
- Some internal data will migrate to a new format, which may cause the client to delay loading until it completes. This may not be backwards compatible with older commits prior to this PR


sdwoodbury (approved):

phillsatellite:
Just wanted to add some notes, while testing I noticed my files were going missing a lot when I would restart the app so I sent Darius over logs and the profile I was using and he is taking a look to see what is happening

phillsatellite (approved):

PR #1618, opened 2023-12-12
feat(logger): use env_logger and RUST_LOG for better control of logging

sdwoodbury:
### What this PR does 📖

- what the title says
- adds command line option to enable writing to file because the LogProfile was removed. 
- assumes that unless production mode is set the logger will be sending output to the terminal

- Resolve #

### Additional comments 🎤
example: `export RUST_LOG=uplink=debug,common=debug,warp_blink_wrtc=trace; target/debug/uplink` 

careful about doing things like `RUST_LOG=error` which sets a global log level - this will cause tons of ipfs related logs from other warp modules. 

note that each module in the "uplink" workspace needs to be enabled separately. in the above example, `uplink` (the `ui` module) and `common` are enabled but extensions and kit are not.


zeenix (dismissed):
Great stuff but I think we want `tracing` and `tracing-subscriber` instead of `log` and `env_logger`. The reasons are documented [here](https://www.mongodb.com/docs/drivers/rust/current/fundamentals/tracing-logging/#:~:text=In%20the%20Rust%20driver%2C%20the,to%20consume%20event%20messages%20programmatically.) but also to keep in mind that most of the Rust projects out there have switched to tracing [a while ago](https://github.com/rust-lang/rustc-dev-guide/issues/892).

zeenix:
> Great stuff but I think we want `tracing` and `tracing-subscriber` instead of `log` and `env_logger`. The reasons are documented [here](https://www.mongodb.com/docs/drivers/rust/current/fundamentals/tracing-logging/#:~:text=In%20the%20Rust%20driver%2C%20the,to%20consume%20event%20messages%20programmatically.) but also to keep in mind that most of the Rust projects out there have switched to tracing [a while ago](https://github.com/rust-lang/rustc-dev-guide/issues/892).

I was going to say that we can do that in a followup PR but skimming through `tracing` docs, I got reminded:

> Users of the [log](https://docs.rs/log/0.4.6/log/) crate should note that tracing exposes a set of macros for creating Events (trace!, debug!, info!, warn!, and error!) which may be invoked with the same syntax as the similarly-named macros from the log crate. Often, the process of converting a project to use tracing can begin with a simple drop-in replacement.

dariusc93:
> > Great stuff but I think we want `tracing` and `tracing-subscriber` instead of `log` and `env_logger`. The reasons are documented [here](https://www.mongodb.com/docs/drivers/rust/current/fundamentals/tracing-logging/#:~:text=In%20the%20Rust%20driver%2C%20the,to%20consume%20event%20messages%20programmatically.) but also to keep in mind that most of the Rust projects out there have switched to tracing [a while ago](https://github.com/rust-lang/rustc-dev-guide/issues/892).
> 
> I was going to say that we can do that in a followup PR but skimming through `tracing` docs, I got reminded:
> 
> > Users of the [log](https://docs.rs/log/0.4.6/log/) crate should note that tracing exposes a set of macros for creating Events (trace!, debug!, info!, warn!, and error!) which may be invoked with the same syntax as the similarly-named macros from the log crate. Often, the process of converting a project to use tracing can begin with a simple drop-in replacement.

Yea. tracing can usually be a drop-in replacement by just importing `tracing`. It can be done in a separate pr though since it may have to be looked into hooking `tracing-subscriber` into the dev tool logger (though should work fine out of the box for terminal and to file)

sdwoodbury:
Sheldon wants command line options to set RUST_LOG, and a default setting.

dariusc93:
> Sheldon wants command line options to set RUST_LOG, and a default setting.

Havent tried it myself with `log` but you should be able to set the env from rust using `std::env::set_var`. Maybe use dotenv crate to make it configurable? I know in tracing it could be made easier

zeenix:
> Sheldon wants command line options to set RUST_LOG, and a default setting.

@tooshel what's the use case here? :thinking: The defaults from upstream are typically good. You typically set them for debug and you can easily do that with the env.

sdwoodbury:
@zeenix  requesting the `tracing` stuff be added later so that I can focus elsewhere. 
@tooshel  a sensible default has been added. I don't think we were using all the previously available levels and a single default is probably good enough.

sdwoodbury:
maybe we should use a .env file for development but not use it for released software. it does seem like there's a lot of modules to set.

tooshel:
> maybe we should use a .env file for development but not use it for released software. it does seem like there's a lot of modules to set.

Usually in released software you use real env vars, not .env files. .env files are a developer convenience (because setting and managing env vars outside a file is painful and error prone). I like the sensible default! Let's get it though QA and merge it and come back to tracing later.

zeenix:
> Usually in released software you use real env vars, not .env files. .env files are a developer convenience (because setting and managing env vars outside a file is painful and error prone).

There are so many ways to easily set the env (both locally and globally), .env is just one of them. You can also set it per app (on Linux that's through the .desktop or .service files but I'm sure there are Windows and Mac equivalents of that). You can also provide people a script to run, instead of a command.

> I like the sensible default! Let's get it though QA and merge it and come back to tracing later.

For non-developers, the default default is good. Emitting debug log comes at a performance cost and typically non-developers won't need it. when we need more logs, we can always provide an easy way to set it to what we need. We'll likely need tracing level in those cases anyway.

----

Having said that, I don't have strong opinion on the subject even if it may not seem so. :) So please go ahead and merge.

stavares843 (approved):

dariusc93 (approved):

PR #1626, opened 2023-12-13
chore(license): Re-license to GPLv3

zeenix:
### What this PR does 📖

The current CCC license is not appropriate for software and it would be best to have a proper license before any external parties contribute. Otherwise, we would have to ask their permission before re-licensing.

Choosing GPLv3 because I understood from my chat with @InfamousVague that for the app, we want a stricter license than libraries, that only allows the code to be used in Open Source software. However, please do chime in on the discussion and let me know if there is a better one.

### Additional comments 🎤

We also need to mention the license in the settings (we already have a License page for it) and I'll follow-up with another PR for that once I've figured out sufficient Dioxus-foo. :)


dariusc93:
Putting aside my biased opinion on copyleft licenses, I think the one problem would really come to deriving projects in uplink. For example how would this impact future extensions/plugins (which may or may not contain code that is close source - once the security side of them are properly resolved), anything that is hooking into the code base for direct or indirect use of uplink code, etc. It also makes it difficult to re-license in the future if we feel this license isnt best to use. I think the idea he wants of having a "stricter license" would be to deter one from using our work as their own and likely prevent one from having some form of unapproved commercial and/or monetary gain, which GPL itself, may not serve that purpose despite its restrictive nature since one could still publish the source, its changes, etc., while still having such gains.​

If we are to relicense (which I do agree that we should move away from CC), I think it would be more appropriate to use Apache 2.0 and/or MIT (or a dual license thereof) and worry less about the use of the code from others until a later date to get things properly sorted out and to determine what we want to have that is "strict" to avoid a mess right now.

zeenix:
> For example how would this impact future extensions/plugins (which may or may not contain code that is close source - once the security side of them are properly resolved), anything that is hooking into the code base for direct or indirect use of uplink code,

That is a good point. I recall that that was problem for use of proprietary gstreamer plugins in Totem (GPL app) and my friend, Christian Schaller had long conversations with RMS about it and they solved it in the end using an [exception clause](https://spdx.org/licenses/GStreamer-exception-2008.html). We can do the same?

> It also makes it difficult to re-license in the future if we feel this license isnt best to use.

True but if we're welcoming contributions and someone does contribute, we'll have problems relicensing too. The only solution to that is a CLA (it could be implicit with a PR) allowing us to relicense in the future.

> I think the idea he wants of having a "stricter license" would be to deter one from using our work as their own and likely prevent one from having some form of unapproved commercial and/or monetary gain, which GPL itself, may not serve that purpose despite its restrictive nature since one could still publish the source, its changes, etc., while still having such gains.​

What GPL **will** protect from is use of our code in non-compatible licenses (including proprietary). If you ask me, we should choose to either go fully Open Source (and then MIT is a good license too) or proprietary. Something in between, with hopes of being able to change the license in the future can get us into trouble.

zeenix:
To summarize, I see a few options here:

1. Relicense to GPLv3 (as in this PR) and forget about relicensing later. GPLv3 will ensure that at least nobody uses our code in proprietary or incompatible OS licensed software.
1. Relicense to MIT and don't think about ever changing the license. Best option if we want to keep it simple and make our peace with any kind of use of our code.
1. Relicense to MIT and add a CLA. While this is a safe option, CLAs are generally frowned upon for very good reasons (it stops a lot of people from contributing).
1. Turn it proprietary (at least for now) and worry about open sourcing later.

I'd recommend we either go with 1. or 4 but wouldn't mind 2. either.

dariusc93:
> To summarize, I see a few options here:
> 
>     1. Relicense to GPLv3 (as in this PR) and forget about relicensing later. GPLv3 will ensure that at least nobody uses our code in proprietary or incompatible OS licensed software.
> 
>     2. Relicense to MIT and don't think about ever changing the license. Best option if we want to keep it simple and make our peace with any kind of use of our code.
> 
>     3. Relicense to MIT and add a CLA. While this is a safe option, CLAs are generally frowned upon for very good reasons (it stops a lot of people from contributing).
> 
>     4. Turn it proprietary (at least for now) and worry about open sourcing later.
> 
> 
> I'd recommend we either go with 1. or 4 but wouldn't mind 2. either.

I would prefer 2

zeenix:
closing in favor of #1637.

PR #1631, opened 2023-12-13
fix(format_text): escape special characters

sdwoodbury:
### What this PR does 📖

- prevents treating text as html unintentionally. 
- this isn't the best fix but it is the easiest and at worst seems like a good temporary solution.
- looks like `> ` doesn't result in blockquotes on dev so this PR doesn't break blockquotes.

- Resolve #

### Additional comments 🎤

before
Screenshot from 2023-12-13 15-23-56

after
Screenshot from 2023-12-13 15-24-49


kit/src/components/message/mod.rs
```diff
@@ -408,6 +408,14 @@ pub fn ChatText(cx: Scope<ChatMessageProps>) -> Element {
 }
 
 pub fn format_text(text: &str, should_markdown: bool, emojis: bool) -> String {
+    // warning: this will probably break markdown regarding block quotes. still seems like an improvement.
+    let mut safe_text = String::from(text);
+    safe_text = safe_text.replace("&", "&amp;");
+    safe_text = safe_text.replace("<", "&lt;");
+    safe_text = safe_text.replace(">", "&gt;");
+    safe_text = safe_text.replace("\"", "&quot;");
+    safe_text = safe_text.replace("'", "&#x27;");
```
dariusc93: You could chain it doing the following

```suggestion
  let safe_text = text
        .replace("&", "&amp;")
        .replace("<", "&lt;")
        .replace(">", "&gt;")
        .replace("\"", "&quot;")
        .replace("'", "&#x27;");
```

dariusc93 (approved):

phillsatellite (approved):

PR #1641, opened 2023-12-14
feat(markdown): Markdown redo

### LINKED ISSUE #1464: chore(markdown): add better support for list items to our markdown parser
Add list markdown in Uplink

example
- 1
	- subitem 1
	- subitem 2
- 2
	- subitem 1 from item 2
	- subitem 2 from item 2

Flemmli97:
### What this PR does 📖

- Adds markdown lists to chat messages
- Current markdown implementation runs the markdown parse for each line in the chat text. This PR changes it so the parse runs across the whole text instead. IDK why we ran it line by line but @lgmarchi said it messed up some code blocks etc. so we might need to check if this breaks any current markdowns.  
  From my test it at least didn't break codeblocks
- Fixes inline codes not working
- Fixes leading whitespaces not working (or message layout in general not kept)
- Blockquotes dont work atm because we escape some html elements to prevent allowing html in messages

- **One more thing to note but i think we can decide on this later how following things should be rendered:**

What should e.g. be bold?
```
**test**
** test **
**test **
** test**
```

- Resolve #1464


sdwoodbury:
- the debug logs need to be removed
- there's an unused function
- single backticks seem to have no effect
- and code blocks are now broken. in the image, SpaceKev is on dev. 

image

Flemmli97:
> * the debug logs need to be removed
> * there's an unused function
> * single backticks seem to have no effect
> * and code blocks are now broken. in the image, SpaceKev is on dev.
> 
> image

unused func was kept for now for easier reverting. will be removed if all problems are fixed

kit/src/components/message/mod.rs
```diff
@@ -489,102 +524,66 @@ fn markdown(text: &str, emojis: bool) -> String {
...
-                        in_code_block = false;
-                        add_text_language = true;
-                        // HACK: To close block code is necessary to push tags 2 times
-                        html_output.push_str("</code></pre>");
-                        html_output.push_str("</code></pre>");
-                    }
+                } else if let pulldown_cmark::Event::End(pulldown_cmark::Tag::CodeBlock(_)) = event
+                {
+                    in_code_block = false;
                 }
```
dariusc93: might be more idiomatic to use a match condition. Something like

```rust
match event {
    pulldown_cmark::Event::Start(pulldown_cmark::Tag::CodeBlock(_)) => in_code_block = true,
    pulldown_cmark::Event::End(pulldown_cmark::Tag::CodeBlock(_)) => in_code_block = false,
    _ => {}
};
```

phillsatellite:
Changes requested for Darius's comment above 👍

kit/src/components/message/mod.rs
Flemmli97: even for only 2 cases? hmm ok

phillsatellite (approved):

PR #1648, opened 2023-12-18
fix(IdentityStatus): Fix identity status update when user update any data in profile page

### LINKED ISSUE #1632: bug(profile): Status Indicator will switch to Offline when updating pfp
**Issue:**
When you update your profile picture in Settings, the status indicator dropdown will change from Online to Offline. Only the dropdown will change though, your indicator will still appear green on friends/chat page to you and your friends.

**OS:**
All

**STR:**
- Go to Settings/Profile 
- Update profile picture 
- After picture is updated online status dropdown will change from Online to Offline 

**Screen recording:**

https://github.com/Satellite-im/Uplink/assets/93608357/f8aeff55-96aa-4484-b0ce-678c21fb5ea5

lgmarchi:
### 1. Fix identity statys update when user update any data in profile page

https://github.com/Satellite-im/Uplink/assets/63157656/a8b5cc83-feef-49f1-a0b3-e2343b66dccc

### 2. Transform duplicated code in one function, to avoid DRY: Don't repeat yourself, and control better data flow

image

- Resolve #1632


common/src/warp_runner/manager/commands/multipass_commands.rs
```diff
@@ -417,14 +370,12 @@ pub async fn handle_multipass_cmd(cmd: MultiPassCmd, warp: &mut super::super::Wa
...
-                        if let Ok(pic) = warp.multipass.identity_picture(&did).await {
-                            id.set_profile_picture(&identity_image_to_base64(&pic));
-                        }
-                        if let Ok(banner) = warp.multipass.identity_banner(&did).await {
-                            id.set_profile_banner(&identity_image_to_base64(&banner));
+                        let mut id = Ok(Identity::from(ids[0].clone()));
+                        update_identity_with_correct_values(&mut id, warp).await;
+                        match id {
+                            Ok(id) => Ok(id),
+                            Err(e) => Err(e),
```
dariusc93: This can instead just return `id` since it is a `Result`

common/src/warp_runner/manager/commands/multipass_commands.rs
```diff
@@ -459,6 +400,26 @@ pub async fn handle_multipass_cmd(cmd: MultiPassCmd, warp: &mut super::super::Wa
     }
 }
 
+async fn update_identity_with_correct_values(
```
dariusc93: To be more consistent with the actual task since this isnt about updating with "correct values" but more just mapping data out to the `Identity` wrapper.

```suggestion
async fn update_identity(
```

common/src/warp_runner/manager/commands/multipass_commands.rs
```diff
@@ -459,6 +400,26 @@ pub async fn handle_multipass_cmd(cmd: MultiPassCmd, warp: &mut super::super::Wa
...
+    id: &mut Result<Identity, Error>,
```
dariusc93: Maybe we can have this be `&mut Identity` instead and try the result early on and return any errors prior to setting the values into the wrapper

lgmarchi: How would be your suggestion? I think I didn't understand

dariusc93: Up where we use the oneshot to send the results we can do something like eg

```rust
let mut id = match warp.multipass.get_own_identity().await.map(Identity::from) {
     Ok(id) => id,
     Err(e) => {
          let _ = rsp.send(Err(e));
          continue;
     }
};
```

That way the results would be returned early, and kind of cleaning up the logic more.

phillsatellite (approved):

dariusc93 (approved):

sdwoodbury (approved):

PR #1656, opened 2023-12-20
feat(about): made in

stavares843:
### What this PR does 📖

- Adds made in section on about section

Captura de ecrã 2023-12-20, às 18 53 26


lgmarchi (approved):

phillsatellite:
Looks like the flags arent appearing when tested on Windows 

image

InfamousVague:
@phillsatellite try now

InfamousVague (approved):

phillsatellite:
Re-tested with the latest on windows 
image

stavares843:
so seems its a issue on windows side

https://answers.microsoft.com/en-us/windows/forum/all/flag-emoji/85b163bc-786a-4918-9042-763ccf4b6c05

after discussing with matt we will ignore this bug as it seems is from windows itself and not the app

PR #1657, opened 2023-12-20
feat(keybinds): keyboard shortcuts

InfamousVague:
### What this PR does 📖
Record custom keybinds:
image
Keybind settings:
image
Find shortcut by kebind: 
image


InfamousVague:
There are a few things left to fix before this is ready for review:

- [ ] When recording a keybind, if you attempt to record a keybind that already exists the focus will shift to the the highlight functionality and break recording the duplicate keybind.
- [ ] I've added the code to warn of duplicate keybinds but it is un-tested because of the aforementioned item.
- [ ] Create a `TooltipWrapper` component in UIKit to enable adding common tooltips to any given children on the component.
- [ ] We should include a button somewhere on the page to "reset to defaults".
- [ ] It can be confusing because META changes to COMMAND on the keybind after recording. We should normalize this somehow between the two.

InfamousVague:
Example of the highlight bug when trying to record duplicate keybinds: 
image

InfamousVague:
image

Added tooltips for conflicting keybinds

InfamousVague:
We are going to leave this feature on experimental until we can add some more keybindings and squash a few more bugs, but this is ready for general testing and we can open further bugs for any bugs that do not cause regressions on existing features.

phillsatellite:
Good by me when workflow is passing. I can create minor tickets for the things that will need addressed 👍

PR #1698, opened 2024-01-10
update(FilePreview): Open image preview in high quality

### LINKED ISSUE #1692: task(chat): Show full resolution when viewing image in chat
**Issue:**
When you open a image in chat the quality is not that great since we are using the thumbnail to preview. Then when you download it the quality is dramatically better. We should show that full resolution quality when trying to just preview an image in chat 

**OS:**
All

lgmarchi:

- Resolve #1692


phillsatellite:
While testing I ran into something weird on Windows

When I try to open the preview, it wont show the image. This doesnt seem to happen with all images though. Like in the screen recording from my PC you can see I am able to open some that were sent to me earlier in chat. Sometimes if I go up and preview those old images and come back to the new ones they will all the sudden appear fine. I replicated this on my PC and on my Windows VM as well. Although on the VM no matter how many times I tried, the image would never preview 

Windows PC: 

https://github.com/Satellite-im/Uplink/assets/93608357/982f892b-29e7-4360-94c1-20f9220f4065

Windows VM: 

https://github.com/Satellite-im/Uplink/assets/93608357/b090c702-09eb-4959-bd36-890bf71b8d59

phillsatellite (approved):

common/src/utils/lifecycle.rs
```diff
@@ -0,0 +1,32 @@
...
+/// and when the component is unmounted
+pub fn use_component_lifecycle<C: FnOnce() + 'static, D: FnOnce() + 'static>(
+    cx: &ScopeState,
+    create: C,
+    destroy: D,
+) -> &LifeCycle<D> {
+    cx.use_hook(|| {
+        cx.spawn(async move {
+            // This will be run once the component is mounted
+            std::future::ready::<()>(()).await;
```
dariusc93, line 18: Just curious, does the task need to yield first before calling `create`?

common/src/utils/local_file_path.rs
```diff
@@ -0,0 +1,18 @@
+use std::path::PathBuf;
+
+/// It will work to load local files in img or video tags, but will ignore drive
+const PREFIX_TO_WORK_ON_WINDOWS_OS: &str = "http://dioxus.";
```
dariusc93: Does this only work on windows or does it work across all platforms? If so, maybe we can add the prefix to the path for loading the local file in the webview (though i believe you said this was webview specific but dont recall if it was ever specific to a single platform)

common/src/utils/local_file_path.rs
```diff
@@ -0,0 +1,18 @@
...
+pub fn get_fixed_path_to_load_local_file(path: PathBuf) -> String {
+    if !cfg!(target_os = "windows") {
+        path.to_string_lossy().to_string()
+    } else {
+        format!(
+            "{}{}",
+            PREFIX_TO_WORK_ON_WINDOWS_OS,
+            path.to_string_lossy().to_string().replace('\\', "/")
+        )
+    }
```
dariusc93: Not required but maybe convert this into a match condition just to make it cleaner.

kit/src/components/embeds/file_embed/mod.rs
```diff
@@ -184,23 +201,30 @@ pub fn FileEmbed<'a>(cx: Scope<'a, Props<'a>>) -> Element<'a> {
...
+                                                onclose: move |_| fullscreen_preview.set(false),
+                                                transparent: false,
+                                                close_on_click_inside_modal: true,
+                                                dont_pad: true,
+                                                img {
+                                                    id: "image-preview-modal-file-embed",
+                                                    aria_label: "image-preview-modal-file-embed",
+                                                    src: format_args!("{}", if temp_dir.exists()
+                                                        { temp_file_path_as_string}
+                                                        else {large_thumbnail} ),
```
dariusc93, line 221: You may be able to avoid using `format_args` (or `format`) if both variables are strings.

ui/src/layouts/chats/presentation/messages/mod.rs
```diff
@@ -626,20 +626,33 @@ fn render_message<'a>(cx: Scope<'a, MessageProps<'a>>) -> Element<'a> {
...
-                    .set_directory(dirs::download_dir().unwrap_or_default()).set_file_name(&file_stem).add_filter("", &[&file_extension]).save_file() {
-                        let conv_id = message.inner.conversation_id();
-                        if !pending_downloads.read().contains_key(&conv_id) {
-                            pending_downloads.write().insert(conv_id, HashSet::new());
-                        }
-                        pending_downloads.write().get_mut(&conv_id).map(|conv| conv.insert(file.clone()));
+                    let conv_id = message.inner.conversation_id();
+
+                    if temp_dir.is_some() {
+                        let file_path_to_download = temp_dir.unwrap_or_default();
```
dariusc93: We should be able to do a `if let` here instead
 
```suggestion
                    if let Some(file_path_to_download) = temp_dir {
```

common/src/utils/local_file_path.rs
lgmarchi: Maybe it can work in all platforms, I will test here.

kit/src/components/embeds/file_embed/mod.rs
lgmarchi: Good point, I will fix it.

stavares843 (changes requested):
https://github.com/Satellite-im/Uplink/assets/29093946/118dd6c9-4379-4f4f-bb94-4c9f14a4b483

you can fast forward the video to the end 😂 

its always reproducible on this branch i couldn't reproduce on dev

macOS, m1


let me know if you need the image

common/src/utils/lifecycle.rs
```diff
@@ -0,0 +1,32 @@
...
+            create();
```
dariusc93, line 19: Something ive overlooked here is that youre calling this in a async task. While im not sure what the context behind it might be for every use case, if blocking functions are called, it might be best to spawn a blocking task here so it does not block the executor by using `tokio::task::spawn_blocking`

dariusc93 (changes requested):

lgmarchi:
Closing this PR. Preview Video Files is already doing what this PR does and even better.

PR #1727, opened 2024-01-17
fix(files): Move download/upload to new tasks

Flemmli97:
### What this PR does 📖

- This PR changes uploads and downloads for both chat and files to be multithreaded:
  Each time a file is uploaded or downloaded a new task is made that deals with it.
  This fixes the problem where uploading/downloading a file will stop the whole app till the file is fully done.


ui/src/utils/async_task_queue.rs
```diff
@@ -0,0 +1,128 @@
...
+{
+    let queue_ref: &UseRef<AsyncRef<T>> = use_ref(cx, || AsyncRef { inner_ref: None });
+    if let Some(queue) = queue_ref.write_silent().inner_ref.take() {
+        for entry in queue {
+            let future = fut(entry);
+            tokio::spawn(future);
+        }
+    }
+    queue_ref
+}
```
dariusc93, line 78: I wonder if this should be separated out so we can track the tasks being spawned and be able to cancel them if needed. Also is there any case where it doesnt rerenders? If so, that can pose a problem since that mean the future wouldnt be polled and no progress would be made (pretty much leaving it at a standstill).

ui/src/utils/async_task_queue.rs
```diff
@@ -0,0 +1,128 @@
...
+            Uuid,
+            Vec<String>,
+            Vec<Location>,
+            Option<Uuid>,
+            AttachmentEventStream,
+        )| {
+            async move {
+                loop {
+                    let msg_clone = msg.clone();
+                    if let Some(kind) = stream.next().await {
```
dariusc93:
```suggestion
                    let msg_clone = msg.clone();
                    while let Some(kind) = stream.next().await {
```

ui/src/utils/async_task_queue.rs
```diff
@@ -0,0 +1,128 @@
...
+                        match kind {
+                            AttachmentKind::Pending(_) => {
+                                return;
```
dariusc93: Keep in mind `AttachmentKind::Pending` provides a result one can use to determine if sending the attachment has been successful or if it failed for some reason.

ui/src/utils/async_task_queue.rs
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// Channel for actions that should be done on the main thread
+pub static ACTION_LISTENER: Lazy<ListenerChannel> = Lazy::new(|| {
+    let (tx, rx) = tokio::sync::mpsc::unbounded_channel();
+    ListenerChannel {
+        tx,
+        rx: Arc::new(Mutex::new(rx)),
+    }
+});
```
dariusc93, line 42: I do wonder if this should, in some way, be localized instead of it being a static. Not required to change in this PR but just something to think about for the future since ideally it shouldnt be sitting as a static.

ui/src/utils/async_task_queue.rs
dariusc93: As I think about it, wouldnt `cx.spawn` or `cx.push_future` work in this case since i believe they track the futures internally?

ui/src/utils/async_task_queue.rs
dariusc93: Note: This is just a suggestion. There is still a closing brace that would need to be remove

ui/src/utils/async_task_queue.rs
Flemmli97: i think it should. as this can be used as a base to expand in the future in case we need to do other offthread stuff that needs to send to the main thread

ui/src/utils/async_task_queue.rs
Flemmli97: no. both `cx.spawn` and `cx.push_future` spawn a future on the main thread. 
also the futures dont cause a rerender. rather the way it works is by adding things to the `AsyncRef` aka you want to do something off thread (which causes a rerender) you cause `tokio::spawn`

ui/src/utils/async_task_queue.rs
Flemmli97: is the loop not needed?

ui/src/utils/async_task_queue.rs
dariusc93: Hmm i would have to check but i thought they push the task into tokio on desktop (since the single threaded part of it is for web side of it iirc) and poll it there until its complete

ui/src/utils/async_task_queue.rs
dariusc93: a `while let` usually is cleaner and easy to read.

ui/src/utils/async_task_queue.rs
Flemmli97: ye no. dioxus is completly single threaded. i already tried spawn and future

luisecm:
Tests are detecting an issue on build process:
image

ui/src/utils/async_task_queue.rs
dariusc93: Hmm I guess ill look more into how dioxus is polling futures

stavares843 (changes requested):
this branch


https://github.com/Satellite-im/Uplink/assets/29093946/ed7a6b94-b8f0-4ddc-a457-d322d59a387d



vs dev

https://github.com/Satellite-im/Uplink/assets/29093946/b5bba16c-8968-488b-af80-3af433cebd8b


STR:
- go to files
- bulk upload 20 files for example
- while is uploading the files number in queue is always the same even is really uploading one by one
- going to dev, using the same files and the same flow, the counter works properly goes from 22, 21, 20 and so on

macOS, m1

Flemmli97:
fixed.
but due to parallel upload processing ui for files upload will flicker cause they all update at the same time.
personally would do a ui rework though on another PR or something

stavares843:
got it, thanks for fixing 🔨

stavares843 (approved):

phillsatellite (approved):

dariusc93 (approved):

PR #1730, opened 2024-01-17
chore(cargo): Update warp dependency

### LINKED ISSUE #1680: bug(profile): Profile and Banner picture will be missing when updated
**Issue:**
If you update either your profile or banner picture, it will show up as grey to all your friends 

**OS:**
All

**Screenshot:**
Screenshot 2024-01-04 at 2 27 42 PM

image

dariusc93:
### What this PR does 📖

- Change how the index signals the task to export
- Prevent request from being excluded when receiving document over mesh
- Change internal events to send conversation document directly, and how update events are handled internally
- Added `GroupSettings` to `RayGun::create_group_conversation`, with a single setting to allow for open or closed groups
- Update group conversation with a block list when group creator blocks/unblocks a user.
- Change internal conversation signature.
- Optimize internal task to prevent a possible race during cleanup
- Use `BTreeMap` for storing reactions instead of a vector
- Store thumbnail internally, adding a reference to `File`.

- Resolve #

- This PR breaks compatibility with the previous commit. Although you should not need a new identity (since there is no breaking change there), conversations may not load or may not be compatible. Furthermore, group conversations in this PR will not be compatible with the any previous commit.
- Should possibly resolve https://github.com/Satellite-im/Uplink/issues/1680 when updating identity using the shuttle discovery.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -534,6 +534,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-recursion"
+version = "1.0.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5fd55a5ba1179988837d24ab4c7cc8ed6efdeff578ede0416b4225a5fca35bd0"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.48",
+]
+
+[[package]]
 name = "async-stream"
 version = "0.3.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


zeenix (approved):
LGTM and despite many changes, it's only changing the code to reflect the new warp API AFAICT so let's merge this since we need it for the new group settings and I'm away next week.

phillsatellite:
@stavares843 and I have been going through the PR a lot this morning, first we had some issues with profile pictures updating but after working directly with Darius he's made some changes and its working good now. The only problem Sara and I are facing now are the online/offline indicators. It seems like if you dont touch the app for a few minutes all of your friends show up as offline. If you message them they will reappear as online again. Other than that, the rest of the pr seems to be working good

phillsatellite (approved):

PR #1748, opened 2024-01-19
chore(ci): fix warning on workflow

stavares843:
### What this PR does 📖
- I noticed this warning `unexpected end of the stream within a flow collection` on the workflow when the file is on edit mode

Captura de ecrã 2024-01-19, às 20 33 37
Captura de ecrã 2024-01-19, às 20 33 40

so i tried this

Captura de ecrã 2024-01-19, às 20 31 53

checking if is all green with CI

### DIFF
--- a/.github/workflows/ui-test-automation.yml
+++ b/.github/workflows/ui-test-automation.yml
@@ -531,14 +531,13 @@ jobs:
 
   remove-label:
     needs:
-      [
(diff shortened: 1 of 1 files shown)


luisecm (approved):
Looks good, didnt know that lol, but as long as it works, it looks even better to read than the array

PR #1765, opened 2024-01-25
fix(chat): Fix user tagging in messages

Flemmli97:
### What this PR does 📖

- Fixes tagging a user not working in chat anymore:
  HTML tags where escaped after the div indicating tags where added causing it to render the div as plain text.
  This PR moves it after they are escaped but it comes with a caveat:
  The process of tagging a user runs as follows:
  User tags someone in their message using `@Name#ShortID` -> converted to `@DID`.
  On receiving side did gets converted to `@Name` (and previously this was cached but its now not possible anymore).
  TBH though this caveat might just be a case of micro optimization.


common/src/state/mod.rs
```diff
@@ -452,13 +446,7 @@ impl State {
...
                 mut message,
             } => {
-                if let Some(ids) = self
-                    .get_chat_by_id(conversation_id)
-                    .map(|c| self.chat_participants(&c))
-                {
-                    message.insert_did(&ids, &self.get_own_identity().did_key());
-                }
-                let ping = message.is_mention;
+                let ping = message.is_mention_self(&self.get_own_identity().did_key());
```
dariusc93: Not related to this PR but might be best to move `self.get_own_identity()` outside of the match condition and borrow from that

common/src/warp_runner/ui_adapter/mod.rs
```diff
@@ -38,49 +38,38 @@ use super::{
...
+            match reg.find(&self.inner.lines().join("\n")) {
+                Some(c) => self.is_mention = if c.as_str().starts_with("`") { 1 } else { 2 },
+                None => self.is_mention = 1,
+            }
         }
-        result = replaced.to_string();
-    });
-    (result, is_mention)
+        self.is_mention == 2
+    }
```
dariusc93: Im a bit confused here. Wouldnt `self.is_mention` that is set in the match condition be overridden by the last one before the function ends?

common/src/warp_runner/ui_adapter/mod.rs
```diff
@@ -38,49 +38,38 @@ use super::{
 pub struct Message {
     pub inner: warp::raygun::Message,
     pub in_reply_to: Option<(String, Vec<File>, DID)>,
-    pub lines_to_render: Option<String>,
-    pub is_mention: bool,
+    is_mention: i8,
```
dariusc93: What does this represent overall?

phillsatellite:
Noticed a few weird things while testing with Sara 

So it seems that anyone Im not friends with, the tag will just show as a did is that intended? Sara tagged my old Phill account (capital P) and since Im not friends with that account Im unable to see it 

Captura_de_ecra_2024-01-25_as_15 56 35
Screenshot_2024-01-25_at_10 55 44_AM

Something I noticed too that was different from how we have it on dev branch if you type out something like header wrapped in h1 tags and wrap it in backticks is shows like this 

Screenshot 2024-01-25 at 11 14 00 AM

 compared to dev that looks like a spoiler message 
 
Screenshot 2024-01-25 at 11 17 33 AM

Flemmli97:
> Noticed a few weird things while testing with Sara
> 
> So it seems that anyone Im not friends with, the tag will just show as a did is that intended? Sara tagged my old Phill account (capital P) and since Im not friends with that account Im unable to see it
> 
> Captura_de_ecra_2024-01-25_as_15 56 35 Screenshot_2024-01-25_at_10 55 44_AM
> Something I noticed too that was different from how we have it on dev branch if you type out something like header wrapped in h1 tags and wrap it in backticks is shows like this
> 
> Screenshot 2024-01-25 at 11 14 00 AM
> compared to dev that looks like a spoiler message
> 
> Screenshot 2024-01-25 at 11 17 33 AM

it is taking the identity from state. so if state has not resolved the correct name it will take the incorrect one.

as for the 2. thing: prob just needed to update this branch to dev? as i dont see how this pr should affect it

common/src/warp_runner/ui_adapter/mod.rs
Flemmli97: if the message is mentioning the current user. its used to decide whether to play a notification sound and highlight the message

common/src/warp_runner/ui_adapter/mod.rs
Flemmli97: yes thats the goal. its lazily evaluated.
its 0 at first and if its 0 it will evaluate to 1 or 2

~~thought this can be prob done in the constructor in the first place hmm...~~

nvm. we need state which is not always present i think during initialization

phillsatellite (approved):

common/src/warp_runner/ui_adapter/mod.rs
dariusc93: Maybe this should be changed to an enum to provide a clear representation

Flemmli97: ~~maybe an option<bool> then?~~
changed

common/src/warp_runner/ui_adapter/mod.rs
```diff
@@ -62,14 +62,20 @@ impl Message {
...
             let reg = mention_regex_epattern(&own.to_string());
             match reg.find(&self.inner.lines().join("\n")) {
-                Some(c) => self.is_mention = if c.as_str().starts_with('`') { 1 } else { 2 },
-                None => self.is_mention = 1,
+                Some(c) => {
+                    self.is_mention = if c.as_str().starts_with('`') {
+                        Some(false)
+                    } else {
+                        Some(true)
+                    }
```
dariusc93:
```suggestion
                    self.is_mention = Some(c.as_str().starts_with('`'));
```

dariusc93: Actually, this could be changed so `reg.find` is used

Flemmli97: actually no. the regex used is a bit special:
regex in rust does not allow lookaheads etc. so to check if a mention is inside codeblocks or not (which should be excluded) i do check
1. since its not possible to check if it is OUTSIDE a codeblock i first detect if its inside
2. if its inside the capture starts with a backtick so its ignored

Flemmli97: the regex used here is a bit special as we need to check and exclude mentions that are inside codeblocks.
but because its not possible (or without lookaheads which the rust regex does not support) to detect if something is NOT true (e.g. not inside a codeblock) i first check if it IS in a codeblock and filter that match out

any other matches then is then outside of a codeblock

dariusc93: I was referring to changing it so it would be 

```rust
self.is_mention = reg.find(&self.inner.lines().join("\n")).map(|c| !c.as_str().starts_with('`')).unwrap_or_default();
```

Flemmli97: oh ok. done. needs to be some(<>.unwrap()) though

common/src/warp_runner/ui_adapter/mod.rs
```diff
@@ -39,49 +39,39 @@ use super::{
...
-            is_mention = true;
+    // Lazily evaluate if the user is mentioned
+    pub fn is_mention_self(&mut self, own: &DID) -> bool {
+        if self.is_mention.is_none() {
+            let reg = mention_regex_epattern(&own.to_string());
+            self.is_mention = Some(
+                reg.find(&self.inner.lines().join("\n"))
+                    .map(|c| !c.as_str().starts_with('`'))
+                    .unwrap_or_default(),
+            );
```
dariusc93, line 71: Realized this may be a option so unwrapping or defaulting may not be needed here.

```suggestion
            self.is_mention = reg.find(&self.inner.lines().join("\n"))
                    .map(|c| !c.as_str().starts_with('`'));
```

Flemmli97: well we want to have it so it is always `Some` after running it once. we dont want it to stay `None`

dariusc93 (approved):

stavares843 (approved):

PR #1775, opened 2024-01-29
feat(misc): Profile picture/banner split

### LINKED ISSUE #1137: chore(misc): Load identity images separately
In https://github.com/Satellite-im/Uplink/pull/976, warp made changes that separated pictures from `Identity`, using separate functions to fetch them. This helps improves performance in different ways (please see PR for some info). Apart of that PR, and to maintain compatibility at the time, images are fetched when the identity is fetched and stored inside the `Identity` wrapper, but long term doing this may not realistically make use of the benefits of the changes made.

Uplink should instead separate this into two task. First would be to fetch the identities (which can be left as is in its current state in warp runner) and separately (for any related identities that were loaded) to fetch their profile and banner images. This would improve loading performance in uplink since the images can be loaded separately and updated in the state accordingly.

Notes:
- The PR above to introduce 2 commands `MultiPassCmd::GetProfilePicture` and `MultiPassCmd::GetProfileBanner` that can be used to obtain pictures and banners of a related identity via `MultiPass::identity_picture` and `MultiPass::identity_banner`
- Would suggest having a separate task that would send events to update the state during the initial load. This should keep things serial while having each task polled separately unless one wish to process each request serially in a separate task.

Flemmli97:
### What this PR does 📖

- Makes it so profile pictures and banner queries are fetched and updated off thread. For each query a thread will be spawned that fetches the data from warp (also batchable). Afterwards a message will be send across a channel from which uplink will handle the data.

- Resolve #1137


stavares843 (approved):

phillsatellite (approved):

common/src/profile_update_channel.rs
```diff
@@ -0,0 +1,62 @@
...
+    pub rx: Arc<Mutex<tokio::sync::mpsc::UnboundedReceiver<ProfileUpdateAction>>>,
+}
+
+pub static PROFILE_CHANNEL_LISTENER: Lazy<ProfileUpdateChannel> = Lazy::new(|| {
+    let (tx, rx) = tokio::sync::mpsc::unbounded_channel();
+    ProfileUpdateChannel {
+        tx,
+        rx: Arc::new(Mutex::new(rx)),
+    }
+});
```
dariusc93: Although previously noted, but this should be localized in some manner than being stored in a lazy imo. Doesnt have to happen in this PR though.

common/src/profile_update_channel.rs
```diff
@@ -0,0 +1,62 @@
...
+                let cmd = if banner {
+                    ProfileUpdateAction::ProfileBannerUpdate(identity, pic)
+                } else {
+                    ProfileUpdateAction::ProfilePictureUpdate(identity, pic)
+                };
+                let _ = PROFILE_CHANNEL_LISTENER.tx.send(cmd);
+            }
+        }
+    });
+}
```
dariusc93: Why not fetch both banner and picture in a single update? Would reduce the round trip.

Flemmli97: initially it was so unneeded data is not fetched but looking at it closer it seems that they are always fetched together so i guess yes that would be good

dariusc93 (approved):

stavares843 (changes requested):
on this branch says this right after entering the app

image


and no profile picture

image


vs dev, same account

image

phillsatellite:
I got the same error as Sara but I also noticed if I try to update my profile picture to my own photo, still doesnt update and remains blank 

https://github.com/Satellite-im/Uplink/assets/93608357/112e2343-21b6-43d4-bc5b-103dc7169bdf

Flemmli97:
> I got the same error as Sara but I also noticed if I try to update my profile picture to my own photo, still doesnt update and remains blank
> 
>  Screen.Recording.2024-02-02.at.12.04.37.PM.mov

fixed but also wondering:
it will wait for the data from warp before updating the profile pic.
do we maybe want to have it so it updates immediately? (though in this case if e.g. an error comes from warp the pic will be reset again which can cause confusion)

dariusc93:
> fixed but also wondering: it will wait for the data from warp before updating the profile pic. do we maybe want to have it so it updates immediately? (though in this case if e.g. an error comes from warp the pic will be reset again which can cause confusion)

If im understanding this statement, then for your own profile, you shouldnt need to concurrently fetch your own identity after updating your profile since the purpose of this is to improve the initial load time by offloading the loading of images into a separate task and update the state with that data if its available. If you are to stick with using this while updating your profile then you might have to wait a bit because it might be that the state is being updated out of order

stavares843:
the avatar is showing now but still appears the error in the log


Captura de ecrã 2024-02-05, às 19 50 34

stavares843 (approved):

phillsatellite (approved):

common/src/profile_update_channel.rs
```diff
@@ -0,0 +1,80 @@
...
+            let _ = warp_cmd_tx.send(WarpCmd::MultiPass(MultiPassCmd::GetProfilePicture {
+                did: identity.clone(),
+                rsp: tx,
+            }));
+
+            let profile_picture = match rx.await {
+                Ok(res) => match res {
+                    Ok(pic) => Some(pic),
+                    Err(_) => None,
+                },
```
dariusc93:
```suggestion
                Ok(res) => res.ok(),
```

common/src/profile_update_channel.rs
```diff
@@ -0,0 +1,80 @@
...
+            let (tx, rx) = oneshot::channel();
+            let _ = warp_cmd_tx.send(WarpCmd::MultiPass(MultiPassCmd::GetProfileBanner {
+                did: identity.clone(),
+                rsp: tx,
+            }));
+            let profile_banner = match rx.await {
+                Ok(res) => match res {
+                    Ok(pic) => Some(pic),
+                    Err(_) => None,
+                },
```
dariusc93:
```suggestion
                Ok(res) => res.ok(),
```

Issue #1718, opened 2024-01-16
task(files): display % while downloading a file


stavares843:
STR:
- go to chat
- send a file
- receive a file

specially when a file is big example 100mb, theres no indication UI wise how many % regarding the download

dariusc93:
This can be resolved by performing the checks while polling the stream

https://github.com/Satellite-im/Uplink/blob/03cd0782de4612f3df3e69a43549c0a032ce3224/ui/src/utils/async_task_queue.rs#L150-L164

PR #1794, opened 2024-01-31
update(FilePreview): Add option to listen to audio files 

lgmarchi:

- Resolve #

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -219,12 +219,6 @@ pub fn return_correct_icon(file_name: &str) -> Icon {
     if is_audio(file_name) {
         return Icon::DocumentAudio;
     }
-    if DOC_EXTENSIONS
-        .iter()
-        .any(|x| file_name.to_lowercase().ends_with(x))
-    {
-        return Icon::Document;
-    }
     Icon::Document
 }
 
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -217,7 +217,7 @@ pub fn return_correct_icon(file_name: &str) -> Icon {
         return Icon::DocumentMedia;
     }
     if is_audio(file_name) {
-        return Icon::DocumentMedia;
+        return Icon::DocumentAudio;
     }
     if DOC_EXTENSIONS
         .iter()
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -11,6 +11,7 @@ use anyhow::bail;
 use clap::Parser;
 // export icons crate
 pub use icons;
+use icons::outline::Shape as Icon;
 use once_cell::sync::Lazy;
 use std::{
     path::{Path, PathBuf},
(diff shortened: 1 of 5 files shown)


phillsatellite:
Noticed a couple things that Lucas has requested be made into new tickets, on MacOS/Linux the icons for mp3 appear very small, Windows the size is normal but it isnt the correct icon 

MacOS/Linux
Screenshot 2024-01-31 at 12 25 38 PM
Windows
image

Another ticket that will be made is the audio doesnt play on Linux (neither does videos) I think its just missing a lib due to the error I receive when trying to play file 

WebKit wasn't able to find a WebVTT encoder. Subtitles handling will be degraded unless gst-plugins-bad is installed.

https://github.com/Satellite-im/Uplink/assets/93608357/c485328c-416a-432b-a9f7-7eb4f1d97a0a

phillsatellite (approved):

common/src/lib.rs
```diff
@@ -203,6 +206,28 @@ pub fn is_video(file_name: &str) -> bool {
...
+    }
+    if DOC_EXTENSIONS
+        .iter()
+        .any(|x| file_name.to_lowercase().ends_with(x))
+    {
+        return Icon::Document;
+    }
+    Icon::Document
+}
+
```
dariusc93: I wonder if we can check the mime from the `File` object itself vs the extension since long term we can have the backend (warp) perform checks long term when uploading a file so we can confirm the file. Nothing is wrong with using the file extension though

stavares843 (commented):
-

stavares843:
phil sent me a mp3 but on my side appears like this

https://github.com/Satellite-im/Uplink/assets/29093946/84f9ac41-e3ca-45d0-b89a-60f92b77ed7b


error on log

Captura de ecrã 2024-01-31, às 22 15 49


macOS, m1

stavares843:
aded changes requested label only due to the mp3 comment above cc @lgmarchi because the mp3 does not play here

dariusc93:
> phil sent me a mp3 but on my side appears like this
> Gravacao.do.ecra.2024-01-31.as.22.14.52.mov
> 
> error on log
> Captura de ecrã 2024-01-31, às 22 15 49
> 
> macOS, m1

thats because it failed to download the file. The UI isnt aware that it failed and only assumed that it succeeded and would attempt to finish its operation as such

https://github.com/Satellite-im/Uplink/blob/03cd0782de4612f3df3e69a43549c0a032ce3224/ui/src/utils/async_task_queue.rs#L150-L164

Related to https://github.com/Satellite-im/Uplink/issues/1718 

EDIT: This issue is not related to this PR specifically.

stavares843:
thanks @dariusc93

common/src/lib.rs
```diff
@@ -203,6 +206,28 @@ pub fn is_video(file_name: &str) -> bool {
...
+    if is_audio(file_name) {
+        return Icon::DocumentAudio;
+    }
+    if DOC_EXTENSIONS
+        .iter()
+        .any(|x| file_name.to_lowercase().ends_with(x))
+    {
+        return Icon::Document;
+    }
+    Icon::Document
```
dariusc93: Since this still return `Icon::Document`, how about we just remove the if condition unless there is another icon that can be used?

lgmarchi: I agree with you

common/src/lib.rs
lgmarchi: I think we can let it for now. Because there are some places we don't have File from warp yet to make this verification.

stavares843 (approved):

dariusc93 (approved):

PR #1846, opened 2024-02-13
feat(files): File transfer UI

### LINKED ISSUE #1718: task(files): display % while downloading a file

### LINKED ISSUE #1749: task(files): Rework files upload ui
With #1727 upload and downloads are now done in parallel for each file.
The ui in file storage for upload though does only support displaying one file per time.
Since the files are uploaded at the same time they will all send their upload progress which causes the upload bar to flicker between the uploads.

We prob want to rework the ui for that to reflect this change

Flemmli97:
### What this PR does 📖

- This PR reworks the download/upload ui:
  - First it creates a tracker globally for all download/uploads (chat, files)
  - If a download/upload is in progress the file button on the navigation bar will display a progress
  - Right clicking will open a progress window
  - If on the files page this progress window will be in the sidebar
  - Do note that chat uploads are NOT tracked as they are visualised via a pending message component. So tracked stuff are:
    - Downloads in from a chat attachment
    - Downloads in the storage page
    - Uploads to the storage

- Resolve #1749
- Resolve #1718

~~Currently pause button does not work (i dont think we have anything to pause it yet atm).~~
~~Also for downloads cancel button will not work as thats not implemented yet either i think.~~


phillsatellite:
@Flemmli97 Noticed a few things while testing 

1. If the user is on the file storage page and they try to open the download tracker it will not open. Download tracker only seems to open if user is on any other page

https://github.com/Satellite-im/Uplink/assets/93608357/0c665dfe-43cc-4a12-a9de-8140508e4743

2. With all the grey the text in the download tracker UI is very hard to see 

image

stavares843:
btw on my side is vertical instead of horizontally which is prettier


image

macOS, m1

Flemmli97:
> @Flemmli97 Noticed a few things while testing
> 
> 1. If the user is on the file storage page and they try to open the download tracker it will not open. Download tracker only seems to open if user is on any other page
> 
>  Screen.Recording.2024-02-13.at.11.55.39.AM.mov 
> 
> 2. With all the grey the text in the download tracker UI is very hard to see
> 
> image

the tracker is below the chat in the sidebar.
and that gray thing is cause (for some reason) the context menu gets selected, or ctrl a. idk why it does it. if you click on the side it should show how it really looks

luisecm:
I have not been able to reproduce the following bug from tests:

However, this bug was not presented before so I am still doing research on it. I pulled your branch and ran the tests locally against it and worked okay, so not sure if the GH CI runner is executing the tests slowly and somehow the app is crashing. I will keep looking on it

image

Flemmli97:
> > @Flemmli97 Noticed a few things while testing
> > 
> > 1. If the user is on the file storage page and they try to open the download tracker it will not open. Download tracker only seems to open if user is on any other page
> > 
> > Screen.Recording.2024-02-13.at.11.55.39.AM.mov
> > 
> > 2. With all the grey the text in the download tracker UI is very hard to see
> > 
> > image
> 
> the tracker is below the chat in the sidebar. and that gray thing is cause (for some reason) the context menu gets selected, or ctrl a. idk why it does it. if you click on the side it should show how it really looks

after some testing it turns out the cause is the markdown live editor in chatbar??
it does some stuff with selection when the app gets refocused (which happens if you download a file) and it selects the wrong thing

common/src/state/data_transfer.rs
```diff
@@ -0,0 +1,258 @@
...
+        }
+    }
+
+    pub fn get_tracker(&self, upload: bool) -> Vec<FileProgress> {
+        if upload {
+            self.file_progress_upload.clone()
+        } else {
+            self.file_progress_download.clone()
+        }
+    }
```
dariusc93: We should borrow from it instead of cloning.

```suggestion
    pub fn get_tracker(&self, upload: bool) -> &[FileProgress] {
        if upload {
            &self.file_progress_upload
        } else {
            &self.file_progress_download
        }
    }
```

common/src/state/data_transfer.rs
```diff
@@ -0,0 +1,258 @@
...
+                progress: TransferProgress::Starting,
+                description: None,
+                state,
+            }),
+        }
+    }
+
+    pub fn update_file_upload(
+        &mut self,
+        file_id: &Uuid,
```
dariusc93:
```suggestion
        file_id: Uuid,
```

common/src/state/data_transfer.rs
```diff
@@ -0,0 +1,258 @@
...
+            .iter_mut()
+            .find(|p| file_id.eq(&p.id))
+        {
+            f.progress = progress;
+        }
+    }
+
+    pub fn update_file_description(
+        &mut self,
+        file_id: &Uuid,
```
dariusc93:
```suggestion
        file_id: Uuid,
```

common/src/state/data_transfer.rs
```diff
@@ -0,0 +1,258 @@
...
+        if let Some(f) = self
+            .get_tracker_from(tracker)
+            .iter_mut()
+            .find(|p| file_id.eq(&p.id))
+        {
+            f.description = Some(description);
+        }
+    }
+
+    pub fn pause_file_upload(&mut self, file_id: &Uuid, tracker: TrackerType) {
```
dariusc93:
```suggestion
    pub fn pause_file_upload(&mut self, file_id: Uuid, tracker: TrackerType) {
```

common/src/state/data_transfer.rs
```diff
@@ -0,0 +1,258 @@
...
+            let current = if let TransferProgress::Progress(p) = f.progress {
+                p
+            } else {
+                0
+            };
+            f.progress = TransferProgress::Paused(current);
+        }
+    }
+
+    pub fn cancel_file_upload(&mut self, file_id: &Uuid, tracker: TrackerType) {
```
dariusc93:
```suggestion
    pub fn cancel_file_upload(&mut self, file_id: Uuid, tracker: TrackerType) {
```

common/src/state/data_transfer.rs
```diff
@@ -0,0 +1,258 @@
...
+        if let Some(f) = self
+            .get_tracker_from(tracker)
+            .iter_mut()
+            .find(|p| file_id.eq(&p.id))
+        {
+            f.progress = TransferProgress::Cancelling;
+        }
+    }
+
+    pub fn error_file_upload(&mut self, file_id: &Uuid, tracker: TrackerType) {
```
dariusc93:
```suggestion
    pub fn error_file_upload(&mut self, file_id: Uuid, tracker: TrackerType) {
```

common/src/state/data_transfer.rs
```diff
@@ -0,0 +1,258 @@
...
+            .get_tracker_from(tracker)
+            .iter_mut()
+            .find(|p| file_id.eq(&p.id))
+        {
+            f.progress = TransferProgress::Error;
+            f.description = Some(get_local_text("files.error-to-upload"));
+        }
+    }
+
+    pub fn remove_file_upload(&mut self, file_id: &Uuid, tracker: TrackerType) {
```
dariusc93:
```suggestion
    pub fn remove_file_upload(&mut self, file_id: Uuid, tracker: TrackerType) {
```

common/src/state/data_transfer.rs
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    fn get_tracker_from(&mut self, tracker: TrackerType) -> &mut Vec<FileProgress> {
+        match tracker {
+            TrackerType::FileUpload => &mut self.file_progress_upload,
+            TrackerType::FileDownload => &mut self.file_progress_download,
+        }
+    }
+
+    pub fn get_tracker(&self, upload: bool) -> Vec<FileProgress> {
```
dariusc93: We could use `TrackerType` here couldnt we?

common/src/warp_runner/manager/commands/constellation_commands.rs
```diff
@@ -464,31 +474,37 @@ async fn handle_upload_progress(
...
-                    .clone()
-                    .lock()
-                    .await
-                    .try_recv()
-                {
-                    if received_tx {
-                        let _ = tx_upload_file.send(UploadFileAction::Cancelling);
-                        break;
+                let mut first = true;
+                while file_state.matches(TransferStates::Pause).await {
```
dariusc93: What we could do instead is use `select` macro from either futures or tokio and send a signal to pause the polling of the stream.

common/src/state/data_transfer.rs
Flemmli97: isnt just using refs better if it can be? reduces clones.

dariusc93: It is better to borrow when possible, but since `Uuid` derives `Copy`, it is inexpensive to do

common/src/warp_runner/manager/commands/constellation_commands.rs
Flemmli97: hmm not too sure what you mean by that.
you mean replacing it with something like this?
```rust
tokio::select! {
  v = file_state.matches(TransferStates::Pause) => if v { sleep(Duration::from_secs(1)).await },
}
```

or getting rid of file_state completely

dariusc93: Something along those lines, although we wouldnt need to sleep or anything. Unless we wish to impl our own stream surrounding the state of the transfer (which might more desirable to do but not really needed), we can add the stream to the select branch and be biased about the order of which things are polled. Eg it would be something like the following 

```rust
    loop {
        tokio::select! {
            biased;
            _ = file_state.matches(TransferStates::Cancel) => return,
            _ = file_state.matches(TransferStates::Pause) => {
                tokio::time::sleep(Duration::from_secs(1)).await;
                continue;
            },
            upload_progress = upload_progress.next() => {
                let Some(progress) = upload_progress else {
                    break;
                };
                match progress {
                    Progression::CurrentProgress { name, current, total } => {
                        if !upload_process_started {
                            upload_process_started = true;
                            log::info!("Starting upload for {name}");
                            log::info!("0% completed -> written {current} bytes")
                        };

                        if let Some(total_size) = total {
                            let current_percentage = (((current as f64) / (total_size as f64)) * 100.) as usize;
                            if previous_percentage != current_percentage {
                                previous_percentage = current_percentage;
                                let readable_current = format_size(current, DECIMAL);
                                let percentage_number = ((current as f64) / (total_size as f64)) * 100.;
                                let _ = tx_upload_file.send(UploadFileAction::Uploading((
                                    Some(Progression::CurrentProgress { name, current, total }), //format!("{}%", percentage_number as usize),
                                    get_local_text("files.uploading-file"),
                                    file_id,
                                )));
                                log::info!(
                                    "{}% completed -> written {readable_current}",
                                    percentage_number as usize
                                )
                            }
                            // ConstellationProgressStream only ends (atm) when all files in the queue are done uploading
                            // This causes pending file count to not be updated which is way we send a message here too
                            if current_percentage == 100 {
                                let _ = tx_upload_file.send(UploadFileAction::Finishing(
                                    file_path.clone(),
                                    file_id,
                                    false,
                                ));
                            }
                        }
                    },
                    Progression::ProgressComplete { name, total } => {
                        let total_size = total.unwrap_or_default();
                        let readable_total = format_size(total_size, DECIMAL);
                        let _ = tx_upload_file.send(UploadFileAction::Uploading((
                            Some(Progression::ProgressComplete { name: name.clone(), total }), //"100%".into(),
                            get_local_text("files.uploading-file"),
                            file_id,
                        )));
                        log::info!("{name} has been uploaded with {}", readable_total);
                    },
                    Progression::ProgressFailed { name, last_size, error } => {
                        log::info!(
                            "{name} failed to upload at {} MB due to: {}",
                            last_size.unwrap_or_default(),
                            error.unwrap_or_default()
                        );
                        let _ = tx_upload_file.send(UploadFileAction::Error(
                            Some(file_path.clone()),
                            Some(file_id),
                        ));
                        break;
                    },
                }
            }
        }
    }

```

Do note this is just from me eyeing it but it would be something like this.

Flemmli97: hmm im not too convinced about that.
we need to do something like this:
```rust
    let mut paused = false;

    loop {
        tokio::select! {
            biased;
            _ = async {}, if file_state.matches(TransferStates::Cancel).await => break,
            _ = async {}, if file_state.matches(TransferStates::Pause).await => {
                if !paused {
                    let _ = tx_upload_file.send(UploadFileAction::Pausing(file_id));
                    paused = true;
                }
            },
            upload_progress = upload_progress.next() => {
                paused = false;
                let Some(upload_progress) = upload_progress else {
                    break;
                };
                last_progress = Some(upload_progress.clone());
                let current_progress = upload_progress.clone();
                match upload_progress {
                    Progression::CurrentProgress {
[...]
```

as `file_state.matches` returns an async bool that should decide whether that section is run. (your version e.g. would always run the first task)

so this whole thing just feels kinda bloated

dariusc93: The purpose of `biased` is to run in order instead of polling a branch at random (which is done by [default for fairness ](https://docs.rs/tokio/latest/tokio/macro.select.html#fairness) of which future is polled) so it would keep it in order. 

> as file_state.matches returns an async bool that should decide whether that section is run. (your version e.g. would always run the first task)

The code I provided was just an example, although since its being polled there, you could technically use a channel to communicate the state so it would always be local, in which a check could be done in the select branch.

The changes you made to the example, however, would still continue to poll the stream `upload_progress` since it does not continue to the beginning when its paused (as attempted in my example) while previously (before my example), we are polling `upload_progress` (with `while let Some(upload_progress) = upload_progress.next().await`), which would cause it to progress the stream regardless of the state for pausing, or resuming.

EDIT: Just as a note, you can check the condition when attempting to poll the stream too

Flemmli97: actually no. as my version
`_ = async {}, if file_state.matches(TransferStates::Cancel).await => break,`
 will run that task if the state is cancel
but your version
`v = file_state.matches(TransferStates::Cancel) => break`
will run it regardless of the state as there is no condition to stop that.
and having something like 
`v = file_state.matches(TransferStates::Cancel), if v => break` will not work as its not supported

dariusc93: Keep in mind i stated the code I provided was an example (since I was eyeing it from github). I am aware that my example doesnt check the boolean there (which was intentional). I was more referring to the stream itself. On top of that, i did mention in my previous statement that we could have the state local so it could be checked or changed there when attempting to poll the stream by doing `upload_progress = async {}, matches!(state, TransferStates::Normal) => {}` (or if you wish to continue to use `files_state` you can use that in place of the match macro) or something along those lines.

Flemmli97: if we want to have the state local though we would need some sort of channel that tells it to set the value.
and when i initially designed this i opted against channels cause:
1. the channel needs to be either global and when sending e.g. an id that matches the current stream over that channel we update the value but this would mean every stream would need to listen to that channel (and thus wouldn't it be better if they only need to listen to things actually affecting them?) and i also am not sure what happens if they all try to poll some value there
2. if the channel is also local we would need to send the channel itself back to our tracker which in that case would be just better to keep the current impl

3. maybe something like this could work?
```rust
    let mut paused = false;

    loop {
        let state = file_state.get().await;
        tokio::select! {
            biased;
            _ = async {}, if matches!(state, TransferStates::Cancel).await => break,
            _ = async {}, if matches!(state, TransferStates::Pause).await => {
                if !paused {
                    let _ = tx_upload_file.send(UploadFileAction::Pausing(file_id));
                    paused = true;
                }
            },
            upload_progress = upload_progress.next() => {
                paused = false;
                let Some(upload_progress) = upload_progress else {
                    break;
                };
                last_progress = Some(upload_progress.clone());
                let current_progress = upload_progress.clone();
                match upload_progress {
                    Progression::CurrentProgress {
[...]
```
but like i said i feel like this whole thing has not much of a point

common/src/warp_runner/manager/commands/constellation_commands.rs
```diff
@@ -464,92 +472,112 @@ async fn handle_upload_progress(
...
+                    let _ = tx_upload_file.send(UploadFileAction::Pausing(file_id));
+                    paused = true;
                 }
-                if !upload_process_started {
-                    upload_process_started = true;
-                    log::info!("Starting upload for {name}");
-                    log::info!("0% completed -> written 0 bytes")
+            },
+            upload_progress = upload_progress.next() => {
+                paused = false;
```
dariusc93, line 504: We can check `paused` to determine if we can poll `upload_progress`

```suggestion
            upload_progress = upload_progress.next(), if !paused => {
```
Though note that we would need to have a previous condition to check to see if `TransferStates` is `TransferStates::Normal` so we can set that flag to false.

Flemmli97: hmm idk though.
not too much of a fan with this:
Screenshot 2024-02-20 at 14 35 03

InfamousVague:
Let's create a high priority ticket to fix the crashing on mobile but since this has been a lot of work let's get what @Flemmli97 has built in! 😀

phillsatellite:
Noticed one other thing we might want to address before we merge this in, if the user has enough chats that they are able to scroll they will not see the upload/download UI. It should be appear overtop of the chats similar to how the call modal we have is 

If we want to merge and just make a ticket for this, thats fine with me 

**Uploading file**

https://github.com/Satellite-im/Uplink/assets/93608357/1d632dab-8c04-4c84-8b2c-666e297b8b1f

**Call UI**

https://github.com/Satellite-im/Uplink/assets/93608357/83b0508f-7700-48cd-9d47-35ff959d7695

phillsatellite:
@Flemmli97 Looks a lot better! Thank you for the changes! There is one other thing I wanted to mention (I am not able to replicate this each time. @tooshel was also able to replicate but @stavares843 was not) 

When you upload multiple files at once, sometimes the UI will only show the first file it starts to upload and then not update when it starts to upload the rest of the files. I could make a new ticket for this to be handled in since it doesnt happen each time just let me know what works best for you! 

https://github.com/Satellite-im/Uplink/assets/93608357/67f55646-394c-4412-8743-671341b493ac

Flemmli97:
> @Flemmli97 Looks a lot better! Thank you for the changes! There is one other thing I wanted to mention (I am not able to replicate this each time. @tooshel was also able to replicate but @stavares843 was not)
> 
> When you upload multiple files at once, sometimes the UI will only show the first file it starts to upload and then not update when it starts to upload the rest of the files. I could make a new ticket for this to be handled in since it doesnt happen each time just let me know what works best for you!
> 
>  Screen.Recording.2024-02-26.at.12.56.47.PM.mov

are those files uploading? despite not being shown?

Flemmli97:
added tracking for duplicate files which hopefully should solve that problem with uploading same file in quick session.
previously duplicates where only tracked if storage already contained the file and not if duplicates are already uploading

phillsatellite:
> > @Flemmli97 Looks a lot better! Thank you for the changes! There is one other thing I wanted to mention (I am not able to replicate this each time. @tooshel was also able to replicate but @stavares843 was not)
> > When you upload multiple files at once, sometimes the UI will only show the first file it starts to upload and then not update when it starts to upload the rest of the files. I could make a new ticket for this to be handled in since it doesnt happen each time just let me know what works best for you!
> > Screen.Recording.2024-02-26.at.12.56.47.PM.mov
> 
> are those files uploading? despite not being shown?

Yes, they are

Flemmli97:
> > > @Flemmli97 Looks a lot better! Thank you for the changes! There is one other thing I wanted to mention (I am not able to replicate this each time. @tooshel was also able to replicate but @stavares843 was not)
> > > When you upload multiple files at once, sometimes the UI will only show the first file it starts to upload and then not update when it starts to upload the rest of the files. I could make a new ticket for this to be handled in since it doesnt happen each time just let me know what works best for you!
> > > Screen.Recording.2024-02-26.at.12.56.47.PM.mov
> > 
> > 
> > are those files uploading? despite not being shown?
> 
> Yes, they are

can you maybe send the logs that occur during that time?

Flemmli97:
i tried the most recent warp branch to see and found out that it still takes a bit for the complete upload.
as such i opted to not remove the tracking if the file reaches 100% but when the process is done instead.
i also added a description component indicating that the process is finishing up which should be more clear now

stavares843:
as discussed in discord, the UI atm isn't ideal can we revert to be able to merge this? thanks 🔨

phillsatellite (approved):

stavares843 (approved):

dariusc93 (approved):

PR #1876, opened 2024-02-27
task(files): Add failed msg to upload failing

### LINKED ISSUE #1848: task(files): Show toast notification when user is sending file but limit is reached
**Issue:** When you are trying to send a file but you've already reached your max limit your message will fail immediately when trying to send. We should show a toast notification saying something like "Max file limit has been reached!" 

**OS:**
All

**Screen recording:**

https://github.com/Satellite-im/Uplink/assets/93608357/9b670b01-c34f-4e29-920d-4ecd02c22ca5

Flemmli97:
### What this PR does 📖

- Adds the error message to pending message if an error occurred during file upload
- Also show the error as a toast notification 

- Resolve #1848

### DIFF
--- a/common/locales/en-US/main.ftl
+++ b/common/locales/en-US/main.ftl
@@ -107,6 +107,8 @@ messages = Messages
     .username-suggestion = Suggested Users
     .control-group = More
     .no-chats = No chats available
+    .attachments-fail = Upload Failed!
+    .attachments-fail-msg = Upload failed: { $reason }
     
 favorites = Favorites
     .favorites = Favorites
--- a/common/src/state/mod.rs
+++ b/common/src/state/mod.rs
@@ -1262,7 +1262,8 @@ impl State {
         conv_id: Uuid,
         msg: PendingMessage,
         progress: Progression,
-    ) {
+    ) -> bool {
+        let mut update = false;
         if let Progression::ProgressFailed {
             name,
             last_size: _,
@@ -1271,9 +1272,9 @@ impl State {
         {
             let err = match error.as_ref() {
                 Some(err) => {
-                    get_local_text_with_args("attachments-fail-msg", vec![("reason", err)])
+                    get_local_text_with_args("messages.attachments-fail-msg", vec![("reason", err)])
                 }
-                None => get_local_text("attachments-fail"),
+                None => get_local_text("messages.attachments-fail"),
             };
             self.mutate(Action::AddToastNotification(ToastNotification::init(
                 name.clone(),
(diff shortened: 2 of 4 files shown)


phillsatellite:
Sara and I both tried and had the same results 

File limit was reached and we tried to then send a file in the chat but got no toast notification and message would not show as failed 

PR branch

We tried to send the same file on dev branch and on dev on the message bubble it will at least say failed

Dev

stavares843:
same here 🔨

Flemmli97:
fixed. but the toast msg will also have the same problem as #1877

stavares843:
thanks 🔨

phillsatellite (approved):

stavares843 (changes requested):
Captura de ecrã 2024-03-04, às 19 23 13

is there a way to make the copy more user friendly? 

can we remove where it says

> Current length: 11461366070. Minimum Length: None, Maximum: Some(10737418240)

and add the MB's instead?

Flemmli97:
> Captura de ecrã 2024-03-04, às 19 23 13
> is there a way to make the copy more user friendly?
> 
> can we remove where it says
> 
> > Current length: 11461366070. Minimum Length: None, Maximum: Some(10737418240)
> 
> and add the MB's instead?

hmm idk about that. it currently simply prints the error provided by a failed download.
and that error msg is provided by warp.

we can of course simply print a simple error message with something along the line of:
"Error during download - <file size>"
but this also looses any other potential context

stavares843:
we can just add a message saying "Error on upload, no space left on storage" and maybe we could add the same storage size we use on the files section

Captura de ecrã 2024-03-05, às 16 21 05

Flemmli97:
> we can just add a message saying "Error on upload, no space left on storage" and maybe we could add the same storage size we use on the files section
> 
> Captura de ecrã 2024-03-05, às 16 21 05

the thing is though that error might not be related to a full storage as its just a generic one. so always displaying it as "full storage" is misleading.
so we could only have "Error on upload"

stavares843:
"Error on upload - lack of storage" works fine

Flemmli97:
like i said what if that error is not caused by lack of storage?
e.g. the wifi went out? 
there is no way to tell. we would need to have warp send the actual error struct instead of just passing in a string

dariusc93:
> like i said what if that error is not caused by lack of storage? e.g. the wifi went out? there is no way to tell. we would need to have warp send the actual error struct instead of just passing in a string

My PR already changes it to an `Error`, so once that is merged in, we can use that instead

stavares843:
sounds good, thank you 🔨

stavares843 (approved):

dariusc93 (approved):

PR #1917, opened 2024-03-11
chore(cargo): Update dependencies

Flemmli97:
### What this PR does 📖

- This PR updates all our dependencies to their latest version and fixes compiling problems caused by that.
- The one exception is the `muda` crate as the newest version requires also a newer gtk version which conflicts with (i think) our current dioxus version or something.
- While there shouldn't be any bugs with this there is no guarantee as i've only fixed compile problems


luisecm:
@Flemmli97 Building the app on Windows is erroring on CI - https://github.com/Satellite-im/Uplink/actions/runs/8233090959/job/22512102062

dariusc93 (approved):

zeenix (approved):

stavares843 (approved):

PR #1918, opened 2024-03-11
fix(Keybinds): Fix keybinds order when recording

### LINKED ISSUE #1845: bug(keyBindings): Alt always appears first when mapping keybindings
**Issue:**
Not 100% sure if this is how we had it intended or if its a bug but Luis had pointed out that when you set a custom keybinding like Control+Shift+Alt+something it will always appear as Alt+Control+Shift like its in alphabetic order instead of the intended order set by the user. 

**OS:**
All

**Screen recording:**

https://github.com/Satellite-im/Uplink/assets/93608357/27301d5c-42b8-459d-8588-05dec3785ce1

lgmarchi:
### 1.  Fix keybinds order when recording

Order of keybinds will be always like image above. Being "E" any other key is not modifier.

image

- Resolve #1845

### DIFF
--- a/common/src/state/settings.rs
+++ b/common/src/state/settings.rs
@@ -60,6 +60,25 @@ impl Shortcut {
             .unwrap_or(false)
     }
 
+    pub fn reorder_keybind_string(mut keybinds: Vec<String>) -> Vec<String> {
+        keybinds.sort_by(|a, b| match (a.as_str(), b.as_str()) {
+            ("Command", "Shift")
+            | ("Command", "Alt")
+            | ("Ctrl", "Shift")
+            | ("Ctrl", "Alt")
+            | ("Ctrl", "Command") => std::cmp::Ordering::Less,
+            ("Shift", "Alt") => std::cmp::Ordering::Less,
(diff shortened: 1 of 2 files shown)


luisecm:
@lgmarchi The automated test currently failing will need to be updated once that this PR is merged, since its expecting that when sending Alt + Ctrl + Shift + P, it will record Alt + Ctrl + Shift + P. Given the change from this PR, the new expected result should be Ctrl + Shift + Alt + P, regardless of the order of push, right?

lgmarchi:
> @lgmarchi The automated test currently failing will need to be updated once that this PR is merged, since its expecting that when sending Alt + Ctrl + Shift + P, it will record Alt + Ctrl + Shift + P. Given the change from this PR, the new expected result should be Ctrl + Shift + Alt + P, regardless of the order of push, right?

Perfect @luisecm, we will keep this pattern, I saw other apps, and we will follow always same order, independent of order user typed

stavares843 (approved):

dariusc93 (approved):

PR #1922, opened 2024-03-12
chore(deps): Drop unused tokio-util dep

zeenix:
### What this PR does 📖

Drop unused tokio-util dependency.


dariusc93 (approved):
Would you like to also add `linkify` (in kit), `notify`  (in common) to this PR too?

stavares843 (approved):

zeenix:
> Would you like to also add `linkify` (in kit), `notify` (in common) to this PR too?

I'll just do them all in a separate PR now that this is in and all approved already. :)

PR #1942, opened 2024-03-25
feat(chat): Chat upload rework

### LINKED ISSUE #1911: task(chats): if message fails, message should clear from chat
**Issue:**
If you try to send a message and it fails, the message just remains in chat as if its still trying to send. The only way to get the failed message to clear is to close and re open the app. Expected behavior is if message fails, the failed message should clear out of the chat 

**Steps to reproduce:**
- Fill up file storage limit 
- Try to send a file in chat that is more than your allowed storage limit 
- When you send the message and it fails, the message will remain in chat until you close/reopen 

**OS:**
- MacOS
- Windows
- Linux

**Screen recording:**

https://github.com/Satellite-im/Uplink/assets/93608357/27c8fe3e-c4fa-4e73-8a7b-4a57c2de8ed5

Flemmli97:
### What this PR does 📖

- Redo how uploads in chats are done:
  - When sending a message with attachment in chat 2 buttons will be added if that upload fails for some reason:
  - A button to delete the attachment. If that pending message has only that attachment left deleting it will delete the pending message
  - A button to resend the attachment. If that message has only that attachment resending it will resend the whole message (attachment + text). Else it will simply resend that attachment as a new message without text

- Resolve #1911

### For testing:
One way to have a file fail upload is by reaching the storage limit


common/src/state/pending_message.rs
```diff
@@ -1,17 +1,44 @@
...
 // Else it is just a guesswork
 #[derive(Clone, Debug)]
 pub struct PendingMessage {
-    pub attachments_progress: HashMap<String, FileProgression>,
+    pub attachments_progress: HashMap<FileLocation, FileProgression>,
     pub message: Message,
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub enum FileLocation {
```
dariusc93, line 16: Assuming this is done because `Hash` is not implemented in `Location`, a PR on warp side does add this so this might be redundant in the neat future

dariusc93 (approved):

common/src/state/pending_message.rs
Flemmli97: ye that was the reason for it

phillsatellite (approved):

stavares843 (approved):

PR #1951, opened 2024-03-27
fix(themes): Text selection color

### LINKED ISSUE #1948: task(themes): in windows select text is not very visible
STR:
- go to recover account
- add text
- select
- appears like this

image

windows only

Flemmli97:
### What this PR does 📖

- Changes the text selection color. While fixing the issue i noticed some stuff though:
  - Text selection color gets applied with some blending algorithm with the background if the provided color either has no alpha value provided (e.g. `#ff00ff`) or an alpha value of 1 (e.g. `#ff00ffff`). This seems to only apply to macos though as just judging from the linked issue on windows the color gets applied as it is causing the problem.
    This makes a consistent ui kinda hard.
  - For now i made the contrast between selection and text color a bit more to see how it goes

- Resolve #1948

### DIFF
--- a/kit/src/elements/textarea/markdown.scss
+++ b/kit/src/elements/textarea/markdown.scss
@@ -47,7 +47,7 @@
     }
 
     &.cm-focused > .cm-scroller > .cm-selectionLayer .cm-selectionBackground, .cm-selectionBackground, .cm-content ::selection {
-        background: var(--text-selection);
+        background: color-mix(in srgb, var(--text-selection) 40%, transparent);
     }
 }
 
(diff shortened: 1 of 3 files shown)


phillsatellite:
Hey @Flemmli97 👋  When Sara and I tested it seemed to be almost the same as dev 

This branch
image

Dev branch 
image

Flemmli97:
i now added an alpha value (of 0.4 cause that seems apparently the default on mac). can you try and see how it looks? on mac this makes it sometimes look different though than before

original
Screenshot 2024-03-28 at 13 13 32


now:
Screenshot 2024-03-28 at 13 18 27

phillsatellite (approved):

stavares843 (approved):

dariusc93 (approved):

PR #1954, opened 2024-03-28
chore(warp): Drop unneeeded spawn_blocking

zeenix:
### What this PR does 📖

Make use of tokio's async API.

### DIFF
--- a/common/src/warp_runner/mod.rs
+++ b/common/src/warp_runner/mod.rs
@@ -331,7 +331,6 @@ async fn init_tesseract(overwrite_old_account: bool) -> Result<Tesseract, Error>
 
     // open existing file or create new one
     Tesseract::open_or_create(&STATIC_ARGS.warp_path, &STATIC_ARGS.tesseract_file)
-        .map_err(Into::into)
 }
 
 impl From<&DiscoveryMode> for Discovery {
--- a/common/src/warp_runner/mod.rs
+++ b/common/src/warp_runner/mod.rs
@@ -315,26 +315,23 @@ async fn wait_for_multipass(
 // assumes that all anyone needs from tesseract is "keypair"
 // otherwise, Tesseract::to_file probably needs to call file.sync_all()
 async fn init_tesseract(overwrite_old_account: bool) -> Result<Tesseract, Error> {
-    tokio::task::spawn_blocking(move || {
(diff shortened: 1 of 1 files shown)


common/src/warp_runner/mod.rs
```diff
@@ -315,26 +315,23 @@ async fn wait_for_multipass(
...
+        if let Err(e) = tokio::fs::remove_dir_all(&STATIC_ARGS.uplink_path).await {
+            log::warn!("failed to delete uplink directory: {}", e);
+        }
 
-            // create directories
-            if let Err(e) = std::fs::create_dir_all(&STATIC_ARGS.warp_path) {
-                log::warn!("failed to create warp directory: {}", e);
-            }
+        // create directories
+        if let Err(e) = tokio::fs::create_dir_all(&STATIC_ARGS.warp_path).await {
```
zeenix, line 327: @dariusc93 btw, we remove the `uplink_path` (just above) but create `warp_path`. Is this on purpose or a typo? 🤔

dariusc93: It is done on purpose iirc since the `uplink_path` would cover the `.user` directory (which includes the warp instance, state and other metadata) and we create the `warp_path` with `create_dir_all`, which would include the `uplink_path`. If we only remove the `warp_path`, any metadata and other bits of data in `uplink_path` would remain and could cause problems. I will admit the naming of the variables should be better so there is clarification, but could be done in a separate PR.

common/src/warp_runner/mod.rs
```diff
@@ -315,26 +315,23 @@ async fn wait_for_multipass(
...
+    }
 
-        // open existing file or create new one
-        Tesseract::open_or_create(&STATIC_ARGS.warp_path, &STATIC_ARGS.tesseract_file)
-    })
-    .await
-    .map_err(anyhow::Error::from)?
+    // open existing file or create new one
+    Tesseract::open_or_create(&STATIC_ARGS.warp_path, &STATIC_ARGS.tesseract_file)
+        .map_err(Into::into)
```
dariusc93: The conversion here shouldnt be needed

dariusc93 (approved):
Left one small comment but besides that LGTM

phillsatellite (approved):

PR #1976, opened 2024-04-15
fix(login): change ordering of account creation pages

### LINKED ISSUE #1968: bug: seed generation & display should come after account creation
Currently, seed is generated upfront and shown to the user, before warp-ipfs is asked to create the account with the given seed. This seems a bit atypical to how this UX is usually done: recover seed is show **after** the account is already created.

Moreover, this has the side-effect of Uplink having to generate the code, while warp-ipfs has the code to do so, if `None` is passed. i-e unnecessary duplication of code.

Flemmli97:
### What this PR does 📖

- Reorders the pages in account creation. Previously it went Password -> Create/Import -> Phrases -> Username.
  Now: Password -> Create/Import -> Username -> Phrases.

- Resolve #1968

### DIFF
--- a/ui/src/layouts/log_in/copy_seed_words.rs
+++ b/ui/src/layouts/log_in/copy_seed_words.rs
@@ -7,13 +7,21 @@ use dioxus_desktop::{use_window, LogicalSize};
 use kit::elements::{button::Button, label::Label, Appearance};
 use tokio::time::sleep;
 
-use crate::get_app_style;
-
 use super::AuthPages;
+use crate::get_app_style;
+use common::state::configuration::Configuration;
+use common::{
+    sounds,
+    warp_runner::{MultiPassCmd, WarpCmd},
+    WARP_CMD_CH,
+};
+use futures::channel::oneshot;
+use futures::StreamExt;
+use warp::multipass;
 
 // styles for this layout are in layouts/style.scss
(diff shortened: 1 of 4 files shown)


phillsatellite (approved):

ui/src/layouts/log_in/copy_seed_words.rs
```diff
@@ -24,20 +32,18 @@ pub fn Layout(cx: Scope, page: UseState<AuthPages>, seed_words: UseRef<String>)
...
-    let words = use_future(cx, (), |_| {
-        to_owned![seed_words];
-        async move {
-            let mnemonic = warp::crypto::keypair::generate_mnemonic_phrase(
-                warp::crypto::keypair::PhraseType::Standard,
-            )
-            .into_phrase();
-
-            seed_words.set(mnemonic.clone());
+    let words = use_future(cx, (), |_| async move {
```
dariusc93, line 35: Just as a note, warp does generate a mnemonic phrase if the `passphrase` field is `None`, in which case you can access it after the account creation in `IdentityProfile::passphrase`

stavares843: marking as resolved to merge 

cc @Flemmli97 if you wanna add that in another PR

